PR #109, opened 2014-12-22
use 127.0.0.1 instead of localhost for db host parameter

Flyingmana:
localhost means connect via socket, which is in some cases not available
or wrong configured or differs from the place php assumes it.

known Downsides: there may be performance impacts, but I never had something noticeable

Reason: It happens regularly that people have connection problems with localhost, which nearly always got resolved by changing to 127.0.0.1

PS: re submitted PR, as master was set previous


buskamuza:
@Flyingmana , according to MySQL documentation (http://dev.mysql.com/doc/refman/5.5/en/connecting.html ), when someone uses `mysql` command as a client, it's also connected to `localhost` via socket. So in case of any issues with sockets, such user won't be able to use default MySQL CLI as well.
IMO: I'd prefer default Magento behavior to be consistent with default MySQL behavior, when it works with MySQL.

Does it make any sense?

Flyingmana:
makes complete sense for me :)

PR #161, opened 2015-01-13
Adding Apache 2.2 VirtualDocumentRoot / RewriteRule bug workaround

jonathanvaughn:
We use VirtualDocumentRoot for part of our development environment so that we don't have to manually create virtual hosts for all developers and their projects, but this occasionally leads to a surprise snag. 

There's a problem with relative rewrites in Apache 2.2 with VirtualDocumentRoot, the fix is to not use relative paths. This affects the `pub/static` and `pub/media` .htaccess rewrites.

While this is an uncommon scenario for production, there may be others with similar development setups that might be scratching their heads over this issue so we're submitting this PR so it will be documented.

### DIFF
--- a/guides/v1.0/install-gde/prereq/apache.md
+++ b/guides/v1.0/install-gde/prereq/apache.md
@@ -69,6 +69,11 @@ Use this section to enable Apache rewrites and specify <code>.htaccess</code> if
 
 	<pre>service apache2 restart</pre>
 
+    <div class="bs-callout bs-callout-info" id="info">
+    <span class="glyphicon-class">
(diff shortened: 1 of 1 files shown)


xcomSteveJohnson:
Can you provide an external reference that mentions this issue? I have not found one yet. Thanks

jonathanvaughn:
Sorry I haven't gotten back to you on this until now.

Here is the Apache httpd bug, and a comment specifically stating the bugfix will likely never be backported to 2.2.x from 2.4.x due to needing infrastructure from 2.4.x for the fix : https://bz.apache.org/bugzilla/show_bug.cgi?id=26052#c40

Many distributions including RHEL/CentOS (and derivatives) 6 still only have 2.2 available (though RHEL/CentOS 7 has 2.4, you can't simply upgrade, have to reinstall the OS...). Building 2.4 from source is possible but then can have problems of it's own with maintaining updates etc.

Again, as I said originally, VirtualDocumentRoot will probably not matter for production environments, but can be a trip hazard for those with similar development environments to ours.

xcomSteveJohnson:
Check it out: http://devdocs.magento.com/guides/v1.0/install-gde/prereq/apache.html#install-ubuntu-apache-rewrites

I shortened it a little bit, eliminating the last sentence. I assume a developer who can use VirtualDocumentRoot knows where their document root is and this is a bit of an edge case as you noted. I haven't seen CentOS 7 yet but Ubuntu 14 has Apache 2.4 by default. Thanks for the tip!

PR #162, opened 2015-01-15
Update Proxy Link in Dependency Injection

bluk:
Missing opening quote for attribute. The generated site displays the HTML tag instead of the link.


xcomSteveJohnson:
Tana, please fix.

PR #163, opened 2015-01-15
Remove method injection

bluk:
Method injection sections may be confused with "setter" method injection (as opposed to constructor injection). There isn't really a difference between the example of method injection and constructor injection as shown.


xcomSteveJohnson:
Tana, please check.

xcomSteveJohnson:
@bluk Please check the topic now: http://devdocs.magento.com/guides/v1.0/extension-dev-guide/depend-inj.html

I believe we made all of your suggestions but I'd like your feedback to be sure.

PR #164, opened 2015-01-15
Move Plugin and Dependency injection links

bluk:
- Move the links from Configuration to Extension. The topics as written describe
  the details of plugins and dependency injection more than the actual config.
  The information is more useful (and should maybe) only be relevant to
  extension developers (e.g. deployers should not normally need to change
  these configuration files).

This was requested by an extension developer and makes sense. I only moved the links because technically the config files described are part of the configuration while the rest is conceptual and is needed for extension developers.

### DIFF
--- a/guides/v1.0/config-guide/bk-config-guide.md
+++ b/guides/v1.0/config-guide/bk-config-guide.md
@@ -15,13 +15,6 @@ github_link: config-guide/bk-config-guide.md
    <li>
       <p><a href="{{ site.gdeurl }}config-guide/config/config-files.html">Configuration files</a></p>
    </li>
-   <li>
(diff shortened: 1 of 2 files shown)


xcomSteveJohnson:
I believe this has been fixed. Please verify and reopen if not.

bluk:
I do not believe this has been fixed but I can't re-open the issue (permissions?).

xcomSteveJohnson:
Apologies. I'll fix.

xcomSteveJohnson:
Sorry for the confusion but this should be fixed now.

http://localhost:4000/guides/v1.0/extension-dev-guide/bk-extension-dev-guide.html

http://devdocs.magento.com/guides/v1.0/config-guide/bk-config-guide.html

PR #167, opened 2015-01-20
Lifecycle instead of Lifestyle

bluk:
I think instead of the word "lifestyle", most documentation refers to the object "lifecycle" instead.

http://en.wikipedia.org/wiki/Object_lifetime


antonkril:
Lifetime is how long does object live. And lifecycle is how object lives. So I would not replace it.

xcomSteveJohnson:
@antonkril The question is about "lifestyle". Do you think it's OK to use "lifecycle" instead of "lifestyle"? Thanks

PR #169, opened 2015-01-20
Update depend-inj.md

bluk:
Proxies can also be used to break cyclical dependencies (where class A depends on class B and class B depends on class A). There's a specific exception that is thrown when the ObjectManager detects a cycle ( https://github.com/magento/magento2/blob/master/lib/internal/Magento/Framework/ObjectManager/Factory/Dynamic/Developer.php#L80 ) in non-production mode. One way to resolve this is to use a Proxy as a constructor argument in the di.xml.

### DIFF
--- a/guides/v1.0/config-guide/config/depend-inj.md
+++ b/guides/v1.0/config-guide/config/depend-inj.md
@@ -54,7 +54,7 @@ Factory
 
 Proxy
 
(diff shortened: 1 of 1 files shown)


xcomSteveJohnson:
Would you mind providing a code example of how to use a proxy as an argument?

bluk:
The section for `Preview of constructor injection` already provides an example of how to inject an auto-generated Proxy. Just giving another reason why someone may choose to use a Proxy.

xcomSteveJohnson:
OK, thanks for clarifying.

PR #195, opened 2015-02-17
Fix sample data install tool param

matkowalski:
It's just a typo fix for PHP install tool param, previously it was *en dash* as a param prefix, corrected to *two hyphens* as the console usage says:
```
Usage: php -f install.php -- --admin_username= [--bootstrap=]
    --admin_username - store's admin username. Required for installation.
    [--bootstrap] - add or override parameters of the bootstrap
```

It's possible to simply copy-paste the example now.


xcomSteveJohnson:
Closing because we coincidentally fixed this today: 

http://devdocs.magento.com/guides/v1.0/install-gde/install/sample-data.html

PR #201, opened 2015-02-27
Update sample-data.md

TexanHogman:
sample data has been simplified to only be one package.  This package is now a composer meta package that automatically pulls in other dependencies.  I also recommend revisit sections on this page that reference beta1 and beta2 as we are now at beta10

### DIFF
--- a/guides/v1.0/install-gde/install/sample-data.md
+++ b/guides/v1.0/install-gde/install/sample-data.md
@@ -29,7 +29,7 @@ To enable sample data, you must edit `composer.json` in the Magento root install
 
 To enable sample data, you must:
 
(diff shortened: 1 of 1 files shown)


xcomSteveJohnson:
Thanks! Coincidentally I just pushed this change. Please check it:

http://devdocs.magento.com/guides/v1.0/install-gde/install/sample-data.html

TexanHogman:
looks great Steve -- tweeted a link to it!!  Great timing -- thanks






Sent from Surface Pro





From: Steve Johnson
Sent: ‎Friday‎, ‎February‎ ‎27‎, ‎2015 ‎4‎:‎24‎ ‎PM
To: magento/devdocs
Cc: Rick Hogge





Thanks! Coincidentally I just pushed this change. Please check it:

http://devdocs.magento.com/guides/v1.0/install-gde/install/sample-data.html

PR #202, opened 2015-02-27
Update to remove reference to sample-data-media

otoolec:
Should  no longer be deciding on which version for both, just sample-data


otoolec:
Related to #201

xcomSteveJohnson:
Good catch, thanks!

PR #215, opened 2015-03-03
Fixed section about static view files deployment tool

buskamuza:
- fixed name of the tool: 
  - it should include all of the words "static", "view", "files". W/o static or view it may have too general meaning
  - it's not "creation" tool because it doesn't create files from scratch, it just copies existing files to proper location and may do some modifications in the files. Currently we use "deployment" word for this tool internally. It may be changed for something else, but the script name and all other documentation should then correspond it
- replaced old errors in troubleshooting section with new one (MAGETWO-31415)
- fixed some inaccuracies:
  - Extended "To deploy static view files" section with one exception
  - fixed command for displaying list of available languages
  - fixed introduction to "Static view files deployment tool" section. Old explanation "because static file URLs are created on the fly" is incorrect - this "because static file URLs are created on the fly" happens in any mode, so this is not the reason for the tool.

### DIFF
--- a/guides/v1.0/config-guide/config/magento-mode.md
+++ b/guides/v1.0/config-guide/config/magento-mode.md
@@ -14,7 +14,7 @@ github_link: config-guide/config/magento-mode.md
 *	<a href="#mode-default">Default mode</a>
 *	<a href="#mode-production">Production mode</a>
 *	<a href="#mode-specify">Specify a mode</a>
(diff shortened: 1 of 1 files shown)


xcomSteveJohnson:
Please verify it: http://devdocs.magento.com/guides/v1.0/config-guide/config/magento-mode.html#mode-production-view

PR #216, opened 2015-03-03
I think clearer dropping "areas" from the list in arch_whatis.md

alankent:
I think clearer dropping "areas" from the list - "Themes and Language packs" then lines up better with "visual design and language capabilities".

(My laptop being repaired - trying a submission purely via GitHub online.)


tanberry:
Thanks Alan! Looks like you don't need no stinking laptop to do PRs. ;-)
Let me take a look at this topic and get back to you again, but yes, sounds like that might make the topic more clear.

PR #220, opened 2015-03-07
apache2 -v instead of apache -v for Ubuntu

mariuszgomse:


xcomSteveJohnson:
Nice catch! Big whoops!

PR #221, opened 2015-03-09
PR#220, correct Apache version mistake

xcomSteveJohnson:


xcomSteveJohnson:
Please check: http://devdocs.magento.com/guides/v1.0/install-gde/bk-install-guide.html

PR #231, opened 2015-03-20
Adds documentation for Forbidden errors

bluk:
Due to a possible change in magento/magento2#533 , we should add a statement in case users can't access their Magento site due to Apache config.


xcomSteveJohnson:
@bluk I changed your PR a little bit because the directive is different for Apache 2.4. Please check: http://devdocs.magento.com/guides/v1.0/install-gde/prereq/apache.html#403-apache

bluk:
@xcomSteveJohnson Looks good, thanks Steve!

PR #238, opened 2015-03-29
Default filename of logo in Magento2 is logo.svg and not logo.png

hostep:
While reading through the docs and experimenting with Magento2, I noticed my custom logo wouldn't show up in a new theme.
Apparantly it was because Magento2 defaults to an svg logo instead of a png one.
See: https://github.com/magento/magento2/blob/1289e4e08fbf8c5f84a069e2a183595fcd14276f/app/code/Magento/Theme/Block/Html/Header/Logo.php#L102


zanilee:
The default logo for a theme is in .svg format. To use a logo with other format your theme you need to  place the logo to `\app\design\<area>\<vendor_name>\<theme_name>\web\images` folder, then create a default.xml in your theme `\app\design\<area>\<vendor_name>\<theme_name>\Magento_Theme\layout\default.xml` and add the logo declaration there:
`<referenceBlock name="logo"> `
`<arguments>`
`<argument name="logo_file" xsi:type="string">images/logo.png</argument>`
`</arguments> `
`</referenceBlock>`

[Documentation](http://devdocs.magento.com/guides/v1.0/frontend-dev-guide/layouts/xml-manage.html#xml-manage-ref-block)

almarchenko:
@hostep Thank you for your input! Updated:
http://devdocs.magento.com/guides/v1.0/frontend-dev-guide/themes/theme-create.html

PR #241, opened 2015-03-31
Fix documentation for Oauth authentication process

anupdugar:
- Added details about Integration registration
- Fix Oauth workflow process steps
- Added missing details for Oauth handshake
- Moved signature and Error codes section
- Removed use of add ons and replaced with Integration


xcomSteveJohnson:
Thanks! Please check it: http://devdocs.magento.com/guides/v1.0/get-started/authentication/gs-authentication-oauth.html

PR #249, opened 2015-04-15
Fix php command to install sample data

bluk:
Add a "-f" to specify the file to run


bluk:
This technically isn't necessary on PHP 5.6 even though it is in the suggested usage.

PR #250, opened 2015-04-15
Add argument delimiter for php command

bluk:
Instead of the "--admin_username" belonging as a PHP command argument, it should be an argument to the script itself.


xcomSteveJohnson:
Closing because the syntax is wrong: should be --admin_user; this is new. I made your change.

xcomSteveJohnson:
Please verify: http://devdocs.magento.com/guides/v1.0/install-gde/install/sample-data.html#instgde-install-sample-enable-after

PR #264, opened 2015-05-02
Removed extra <p> tags from "Theme Components" table

erikhansen:
Some rows in the "Theme Components" table on the http://devdocs.magento.com/guides/v1.0/frontend-dev-guide/themes/theme-structure.html page have extraneous &lt;p&gt; tags, which is causing those cells to display text at a large size:

magento_theme_structure

This pull request removes those tags.


xcomSteveJohnson:
@almarchenko please check this out.

almarchenko:
Thanks for your input!

PR #265, opened 2015-05-02
Fix incorrect module definition XML and add note about how to manually enable module.

ericthehacker:
Page currently has `schema_version` as module attribute for module version, when `setup_version` is correct.

Also added precise instructions for enabling a module. The note above about it no longer being controlled by the module was helpful, but it took me a while to figure out how to actually enable it.

### DIFF
--- a/guides/v1.0/architecture/modules/mod_intro.md
+++ b/guides/v1.0/architecture/modules/mod_intro.md
@@ -53,14 +53,14 @@ Minimal declaration sample:
     To enable a module manually, the following commands can be used.
 </p>
 
-<ol>
-    <li><code>php -f setup/index.php module-enable --modules=Vendor_ModuleName # enable module</code></li>
-    <li><code>php -f setup/index.php update # flush appropriate caches</code></li>
-</ol>  
+<pre><code>
+# Enable Module
+php -f bin/magento module:enable Vendor_ModuleName 
+# Run Enabled Module Setup Scripts
+php -f bin/magento setup:upgrade
+</code></pre>
 
 </div>
-
-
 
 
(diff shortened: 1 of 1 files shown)


xcomSteveJohnson:
@keharper, please check this out.

guides/v1.0/architecture/modules/mod_intro.md
```diff
@@ -41,16 +41,28 @@ Minimal declaration sample:
...
 <div class="bs-callout bs-callout-info" id="info">
   <p>The enabled/disabled flag for a module is no longer set within a module; it is controlled by the deployment configuration file, with enabled=1, disabled=0. This is controlled by administrators and integrators, not by module developers.</p>
+  
+<p>
+    To enable a module manually, the following commands can be used.
+</p>
+
+<ol>
+    <li><code>php -f setup/index.php module-enable --modules=Vendor_ModuleName # enable module</code></li>
+    <li><code>php -f setup/index.php update # flush appropriate caches</code></li>
```
otoolec: I would consider using the new `bin/magento` command as an example instead. 

    # Enable Module
    php -f bin/magento module:enable Vendor_ModuleName 
    # Run Enabled Module Setup Scripts
    php -f bin/magento setup:upgrade

xcomSteveJohnson:
@keharper, Chris is correct. The command has changed. Suggest you add a cross-ref to http://bit.ly/1IJJUS4. It's also safe to use the command Chris suggests.

ericthehacker:
@keharper @xcomSteveJohnson 
Your suggestions have been implemented and the pull request updated.

PR #266, opened 2015-05-05
just a minor grammar error

7php:
use either 'a' or 'the' module in the phrase "the architecture of the a module itself"


tanberry:
Yikes, thanks, we will fix that! We'll be right back...

PR #275, opened 2015-05-23
Update sample-data.md

7php:
ToDate, --admin-user is not working, instead --admin_user works.
Commandline hints clearly shows:
Usage: php -f install.php -- --admin_user= [--modules=] [--bootstrap=]

### DIFF
--- a/guides/v1.0/install-gde/install/sample-data.md
+++ b/guides/v1.0/install-gde/install/sample-data.md
@@ -49,7 +49,7 @@ This section discusses how to install optional sample data *after* you install t
 
 To install the sample data, enter the following command:
 
-	php dev/tools/Magento/Tools/SampleData/install.php -- --admin-user=<your Magento admin user name>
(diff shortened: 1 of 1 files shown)


xcomSteveJohnson:
Note that I changed this after I closed the PR. The command syntax is now:

php dev/tools/Magento/Tools/SampleData/install.php --admin_user=

7php:
ok, so the extra -- in "install.php -- --admin_user" has been removed?

xcomSteveJohnson:
Yes, my understanding is that you use the extra -- to designate the end of PHP command arguments and the start of a CLI/script argument. You either use both or neither (that might be oversimplifying). IOW, 'php -f' is a PHP argument but --admin_user is a script argument.

PHP documentation isn't clear about why you use -f and my understanding is it's safe to omit it.

7php:
thanks for this piece of info Steve! Cheers!
Yes I agree that the usage of -f is really not clear.

PR #276, opened 2015-05-26
admin-password was wrong in the guide

jamescowie:
In the install snippet there is an error with `admin_password` it should be `admin-password` so that the installer can complete.


xcomSteveJohnson:
This should be fixed. I noticed it about the same time you did so closing this PR.

PR #285, opened 2015-06-12
SSH Clone Instructions Give HTTPS URL

maxbucknell:


xcomSteveJohnson:
Thanks! Fixed on http://devdocs.magento.com/guides/v1.0/install-gde/install/composer-clone.html

PR #293, opened 2015-06-21
Improve readability.

denchev:


tanberry:
Thank you, I agree, that does help readability to separate the three terms on their own lines, and to make each term bold. 

Sasha, can you review these changes, and if you agree, go ahead and merge Marush's PR? 

Thanks so much, Marush, we really appreciate the community's help! Keep it coming!

almarchenko:
Wow, it was some kind of technical issue I guess.
Thank you for noticing and correcting!

PR #294, opened 2015-06-23
Update prepare-install.md

mazhalai:
Fix for server paths


xcomSteveJohnson:
Closing because I don't think this is correct strictly speaking and besides, they're just examples.

*  On CentOS 6.x, I know that if you install Apache 2.2 (and probably 2.4), the docroot is /var/www/html by default

*  On Ubuntu, I believe that if you install Apache 2.2 the docroot is /var/www by default. If you install Apache 2.4, the docroot is /var/www/html by default.

This is true to the best of my memory.

PR #299, opened 2015-07-06
Fix HTML rendering

kumabotz:
This fix the HTML rendered on http://devdocs.magento.com/guides/v1.0/coding-standards/code-standard-javascript.html


xcomSteveJohnson:
Thanks for pointing this out. We disagree on the approach, however. We prefer more Markdown and less HTML.

The page should now display properly with more Markdown. If there are still issues, please don't hesitate to make another PR.

xcomSteveJohnson:
http://devdocs.magento.com/guides/v1.0/coding-standards/code-standard-javascript.html

kumabotz:
Agree that and thanks for making the changes. I have submitted #300 to fix some formatting issues.

PR #303, opened 2015-07-10
Update installer-privileges_after.html

Tenacious-G:
amended a small typing error on line49, should be "dedicated Magento server", not "decided Magento server"


tanberry:
Thank you, Tenacious-G, we appreaciate your tenacity in helping us with the Magneto Dev Docs!

PR #305, opened 2015-07-13
Fix indentation

kumabotz:
For http://devdocs.magento.com/guides/v1.0/extension-dev-guide/attributes.html#ext-auth


xcomSteveJohnson:
I decided to get rid of the <pre> tags and thereby might have caused another issue. Closing without merging. Please check it again and let us know if there are issues.

PR #311, opened 2015-07-20
Update config-cli-subcommands-i18n.md

almarchenko:
I suggest we put the Overview section first. Otherwise it is not clear what is the topic about, and what are the First steps for.


xcomSteveJohnson:
Closing

Issue #315, opened 2015-07-23
Old Dev Docs URLs in Search Engine


astorm:
Today I was searching for some Magento 2 information -- and Google returned a result with the following URL

http://devdocs.magento.com/guides/v1.0/config-guide/config/plugins.html

However, this URL returned a 404.  If the old versions of docs aren't going to stick around, there old URLs should redirect (`301`, permanent) to the current equivalent document, or an appropriate page if the old URLs are going away.

xcomSteveJohnson:
Suggestion for how to do it?

steverobbins:
You need to create the page at that location but specify the redirect: https://help.github.com/articles/redirects-on-github-pages/

xcomSteveJohnson:
Is there a simpler way, that would redirect everything from guides/v1.0 to guides/v2.0?

And how long before search engines change those URLs on their own?

steverobbins:
GitHub pages doesn't allow `.htaccess` files so you can't perform blanket redirects using that method.  I'm not sure if there's another option.

Aren't there still [valid pages](https://github.com/magento/devdocs/tree/develop/guides/v1.0) in `v1.0`?

xcomSteveJohnson:
Ironically the only pages left in 1.0 are some temporary Meta tag redirects to cover URLs referenced by code. Those will go away in the next few weeks.

We're investigating. Thanks for bringing it up.

astorm:
It sounds like you're willing let Google realize the pages are 404ing, and to automatically slurp up the new pages.  This might just be frozen caveman web developer talking -- but with that as standard policy aren't you sacrificing any SEO reputation the older URLs had built up?  Every time you do a release of the docs you'll end up needing to build the search engine rep of that page back up.

xcomSteveJohnson:
Sometimes a question is just a question <g>. We're not SEO experts or web designers. We'll find a solution.

alankent:
I think this is partly a symptom of trying to get docs published quickly. We know the infrastructure needs more work. Release early, release often - and get bitten occasionally as well sorry.

xcomSteveJohnson:
@astorm http://devdocs.magento.com/guides/v1.0/config-guide/config/plugins.html

That has not been a valid URL in a very long time, that might go back to Dev Beta in December 2014. However, everything we renamed on 7/21 is now being redirected.

xcomSteveJohnson:
[...]

PR #317, opened 2015-07-23
Issue #315: Redirects broken URL to home page

steverobbins:
I'm not sure if this is the correct destination, but this is how you might specify redirects in pages, rather then using meta refresh tags.


xcomSteveJohnson:
Merged with modifications:

*  Added an explanation of git-redirect-from in comments to _config.devdocs.yml
*  Undo change to index.html; that's not the correct way to do it and I had already fixed the redirect for that particular page. See https://github.com/magento/devdocs/blob/develop/guides/v2.0/extension-dev-guide/plugins.md

We really appreciate your help in identifying this solution. It turned out to be simple after we got the hang of it. Our automation team scripted the vast majority of topic modifications.

Thanks again and keep em coming! We love feedback!!

PR #318, opened 2015-07-27
Sample data

t-richards:
This PR contains:
 - CE 1.9.1.0 content for the [sample data guide](http://devdocs.magento.com/guides/m1x/ce18-ee113/ht_magento-ce-sample.data.html)
 - `m1x` layout for re-use in other Magento 1.x pages

### DIFF
--- a/_layouts/m1x.html
+++ b/_layouts/m1x.html
@@ -1,0 +1,22 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+	<meta charset="utf-8">
+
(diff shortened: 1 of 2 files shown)


xcomSteveJohnson:
Thanks for the tip! We'll consider the layout suggestion but implementation depends on available time to do it. 

We'll get to your sample data links soon.

t-richards:
It's understandable that all m1x pages won't be magically converted overnight. Thanks for the update! :smile:

xcomSteveJohnson:
Thanks!

PR #319, opened 2015-07-27
Jekyll chores

t-richards:
Highlights:
 - Sets `github-pages` version at 38
 - Adds bundler-generated stub for jekyll (this allows running `bin/jekyll` instead of `bundle exec jekyll`)
 - Readme clarification
 - `exclude` key for site configuration


xcomSteveJohnson:
We'll need some help with the programming aspects of this PR. The README changes look good.

What is the advantage of setting the gh-pages version?

I also wonder about exclude: because the README and .yml aren't being converted to HTML anyway. What are your thoughts about that?

t-richards:
>I also wonder about exclude: because the README and .yml aren't being converted to HTML anyway.

This is a somewhat opinionated change on my part. Given the fact that those files aren't being converted to HTML (as you pointed out), and they aren't linked to by any of the pages / guides, then why should making an HTTP request for http://devdocs.magento.com/README.md give you the file? It is my opinion that requests for this URL should be given a 404 response.

Here's another reason to exclude these files: Let's say that I've got a `$ bin/jekyll serve` running in the background and I make a change to the `README.md` file. Welp, now I get to wait 23 seconds while the site regenerates. I don't believe changing the readme file should cause a rebuild.

>What is the advantage of setting the gh-pages version?

Another semi-opinionated change; I believe this article does a good job at explaining the rationale behind this decision: http://everydayrails.com/2012/09/11/bundler-rails-specify-versions.html

The idea behind this article isn't Rails-specific. It's a good idea to explicitly specify dependency versions for any project; static Jekyll sites included. I understand that we already have a `Gemfile.lock` to hold the exact versions of installed gems. However, if a version constraint is set in the Gemfile, it allows for a `$ bundle update` to pull in new patch/minor versions of dependencies, without say, upgrading jekyll to the new 3.0.0 version and breaking everything (which could happen soon!).

xcomSteveJohnson:
Thank you for providing that detail. It will be a while before we merge. We'll update you if there are issues.

PR #324, opened 2015-08-09
Making deployment configuration more readable/relevant

wsakaren:
This page really needs a re-write. I've done a few bits but I feel like it's written from a technical perspective and needs to be altered to be written from a user perspective.

For instance headings like 'config.php contents' mean nothing, a heading that says how to configure modules installed means a lot more.  I'm not sure that I care about knowing that '`config.php` and `env.php` are PHP files that return a multi-dimensional associative array' - is this relevant?  Feels like there is a lot of noise here and the true message about the importance of these files is being lost.

### DIFF
--- a/guides/v2.0/config-guide/config/config-php.md
+++ b/guides/v2.0/config-guide/config/config-php.md
@@ -14,7 +14,6 @@ redirect_from: /guides/v1.0/config-guide/config/config-php.html
 
 *  <a href="#config-php-overview">What is the Magento deployment configuration?</a>
 *  <a href="#config-php-contents">config.php and env.php contents</a>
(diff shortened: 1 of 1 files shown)


xcomSteveJohnson:
Hi @wsakaren Thanks very much! This kind of feedback is very important to us. Please read the updates carefully and let us know if more changes are needed.

wsakaren:
Looks much improved. There is a recursive process required here, but think its more important to hit some of the other pages and get it overall working better than to focus on 1 being perfect. Def looking better.

xcomSteveJohnson:
Please let us know how we can improve to help you more. I'm currently focused on another area so if you have additional suggestions, let us know.

PR #327, opened 2015-08-11
removed outdated reference to static file deployer

riconeitzel:
magento setup:static-content:deploy is used instead of dev/tools/Magento/Tools/View/deploy.php


xcomSteveJohnson:
Derp! Thanks for fixing that!

PR #333, opened 2015-08-24
asp_tags should be always disabled on php.ini

biotech:
Discover that issue when asp_tags was enabled by default , installation goes well and even home and some pages works, but product page and admin no. So any templates which use <% %> abstraction not can been loaded.


biotech:
Actually asp_tags should be complete removed only from PHP7 so this can be really common issue for lot of end users.

xcomSteveJohnson:
@biotech Made some changes. 

Put the info about the config option here (it's in both CentOS and Ubuntu sections): http://devdocs.magento.com/guides/v2.0/install-gde/prereq/php-ubuntu.html#instgde-prereq-timezone

Modified troubleshooting a bit: http://devdocs.magento.com/guides/v2.0/install-gde/trouble/php/tshoot_php-set.html

Let us know what you think.

biotech:
@xcomSteveJohnson looks good! thanks!

PR #334, opened 2015-08-24
Error on installation in system with no files session storage

biotech:
All details on files, also issue created for discuss that https://github.com/magento/magento2/issues/1735 , but this is huge change which will take some time to resolve and issue could be common problem for users who configure session storage at redis or memcached

### DIFF
--- a/guides/v2.0/install-gde/trouble/php/tshoot_session.md
+++ b/guides/v2.0/install-gde/trouble/php/tshoot_session.md
@@ -1,0 +1,33 @@
+---
+layout: default
+group: install_trouble
+subgroup: Errors during installation
+title: During installation, exception SessionHandler::read()
+menu_title: During installation, exception SessionHandler::read()
+menu_node: 
+menu_order: 26
+github_link: install-gde/trouble/tshoot_session.md
+---
+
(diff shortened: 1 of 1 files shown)


xcomSteveJohnson:
@biotech I'll wait until the Magento 2 GitHub issue is resolved before doing this.

biotech:
@xcomSteveJohnson ok, but there is could be long discussion and way to patch, in same time this is common issue which can be blocking for lot of users - really hard to debug and understand the issue, not reported yet on SE or any other place. let's wait for some feedback from core team first - agree :)

biotech:
@xcomSteveJohnson issue confirmed and ticket internal created, but until resolved solution can be helpful for users faced with that.

xcomSteveJohnson:
Thanks, I will investigate. It will take some time, please be patient.

xcomSteveJohnson:
MAGEDOC-2450 replaced by MAGETWO-42059

biotech:
Great , nice to this is accepted. Thanks @xcomSteveJohnson

xcomSteveJohnson:
Sure, thanks!

PR #348, opened 2015-09-12
Use Correct PHP Operators, Not Unicode Arrows

kojiromike:
I'm not sure why these symbols are appearing – perhaps someone's editor is autoreplacing -> for them. Perhaps they're there to avoid problems with HTML output. In this case, though, they occur in examples of PHP, and as such they wouldn't work as intended.


xcomSteveJohnson:
@keharper Please take a look. Thanks.

PR #358, opened 2015-09-24
Changed typo.

euperia:
Changed "You can either add a `require` section do `composer.json`" to "You can either add a `require` section _to_ `composer.json`"


tanberry:
Ooops! Thanks so much @euperia! We really appreciate all contributions to our docs, even when it is catching our typos.

PR #361, opened 2015-09-30
Update example code with the correct variables

debuysserk:
The example variables didn't correspond with the variables used in 1.0.0 beta. These are the correct ones (new naming syntax)


almarchenko:
Thanks! We'll take a look and update!

almarchenko:
Thanks, merged!

PR #363, opened 2015-10-02
Update module-file-structure.md

maxrasguidogoja:
Changed il8n to i18n as modules use i18n not il8n, the image of the folder structure should also be updated to note this change.


tdobbs:
Thanks, @maxrasguidogoja

PR #369, opened 2015-10-08
CSS Design Pattern of W3C

rafaelstz:


xcomSteveJohnson:
Please elaborate on the reason for this change.

rafaelstz:
According to the [W3C CSS 2.1 Specification](http://www.w3.org/TR/CSS2/syndata.html#length-units) for Syntax and basic data types:

*The format of a length value (denoted by <length> in this specification) is a <number> (with or without a decimal point) immediately followed by a unit identifier (e.g., px, em, etc.).* **After a zero length, the unit identifier is optional**.

xcomSteveJohnson:
Sorry, we don't accept PRs into master. Please submit the PR to develop.

rafaelstz:
Ready https://github.com/magento/devdocs/pull/376 :+1: :smile:

PR #371, opened 2015-10-08
Link to Magento Tabs Widget doesn't work

Tym17:
Probably using an older url model


xcomSteveJohnson:
@almarchenko or @rsavchuk Please look at and fix.

almarchenko:
Thanks!!

PR #373, opened 2015-10-09
"Continue to installing the Magento software" doesn't make grammatical sense

benmarks:


xcomSteveJohnson:
Closing for two reasons: first, this is probably mentioned in more than one place and second, we're going to make a big change to the install guide in the near future. This link will be either reworded or will go away.

PR #387, opened 2015-10-19
Contributor's list update

mcspronko:
- added "How do I develop and maintain my custom Magento 2 extension in a separate repository" topic into Contributor's list


tanberry:
Thanks so much Max!! We look forward to adding this article to devdocs.magento.com, and adding to the article your name and a linkback as an extra "Thanks."

PR #389, opened 2015-10-19
Remove extra row

rafaelstz:
**Before**

image

**After**

image


xcomSteveJohnson:
How about adding your name to that row? <g> 

We left the blank row on purpose to encourage others to sign up.

PR #390, opened 2015-10-21
"Click here" links are bad practice

MidnightDesign:


xcomSteveJohnson:
Please explain.

MidnightDesign:
http://www.w3.org/QA/Tips/noClickHere
http://www.smashingmagazine.com/2012/06/links-should-never-say-click-here/
http://uxmovement.com/content/why-your-links-should-never-say-click-here/

(Sorry for just dropping some links, but there are hundreds of articles on this topic and I think they all do a fantastic job explaining it.)

PR #393, opened 2015-10-22
Same padding min-width: 1130px

rafaelstz:
Same padding and semantic code.

Before:
image

After:
image


tahiryasin:
(y)

xcomSteveJohnson:
We're going to fix this in an upcoming code push. Look for our announcement about RWD.

PR #401, opened 2015-10-28
Changed arugment type names to valid options

andrewnk:


xcomSteveJohnson:
It looks like this PR includes PR#402. If so, can I close 402 and merge this one?

PR #404, opened 2015-10-28
Update plugins.md

ldusan84:
It wasn't obvious to me that plugins cannot be created for inherited methods, so I thought it would be useful to put that information.


xcomSteveJohnson:
@tdobbs Please take a look.

tdobbs:
@ldusan84 Thanks for the update!

PR #417, opened 2015-11-18
Updated admin:user:create to update an existing admin user.

miguelbalparda:
Added explanation on how to update an existing admin user via CLI.


xcomSteveJohnson:
We thank you for your contribution; however, after internal discussions, we've decide not to merge it.

We believe that a command named admin:user:create should be used only to create a new admin user. There is a risk that an update by this command could overwrite another user's information. Thus, we created an internal issue MAGETWO-44629 to track creating a new command to update an existing user.

Thank you again and we hope to hear from you in the future.

miguelbalparda:
Good call, keep up the good work!

miguelbalparda:
Can we check on this again? I can include a message stating the risk of overwriting an existing user while we wait on the update command.

PR #422, opened 2015-11-21
Fixed Example of Magento_AdminNotification

tobi-pb:
Code example of Magento_AdminNotification looks outdated comparing it to current magento 2 code


klein0r:
+1

PR #423, opened 2015-11-23
Use correct class constant name

bh-ref:
in class \Magento\Store\Model\StoreManager the constants are named PARAM_RUN_CODE and PARAM_RUN_TYPE

### DIFF
--- a/guides/v2.0/config-guide/bootstrap/mage-run.md
+++ b/guides/v2.0/config-guide/bootstrap/mage-run.md
@@ -70,8 +70,8 @@ If necessary, copy the existing `index.php` entry point script for your website 
 {% highlight php %}
 <?php
  $params = $_SERVER;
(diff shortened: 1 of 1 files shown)


buskamuza:
https://github.com/magento/magento2/blob/develop/app/code/Magento/Store/Model/StoreManager.php - the constants are named as in the old version.

bh-ref:
https://github.com/magento/magento2/blob/develop/app/code/Magento/Store/Model/StoreManager.php#L18 -> `const PARAM_RUN_CODE = 'MAGE_RUN_CODE';`

the constant *value* (**MAGE**_RUN_CODE) is maybe like the old version, but the constant *name* is not (**PARAM**_RUN_CODE), I think?

buskamuza:
@bh-ref , sorry for the confusion. You're absolutely right. The PR is correct.

PR #456, opened 2015-12-18
Property call to local variable change

augsteyer:
In the current context Client is not assigned to a class property


tanberry:
Oops. Client is defined immediately before this section, yes.
Thanks, @augsteyer, we appreciate your help there!

PR #457, opened 2015-12-20
Removing Order directive

butcher211:
Hello,

I have an Ubuntu 14.04 server installed with Apache 2.4.7 and I was trying the following scenarios:

### 1. Scenario - the magento2-doc way

```
<Directory /var/www/>
	Options Indexes FollowSymLinks MultiViews
	AllowOverride All
	Order allow,deny
	Require all granted
</Directory>
```
### 2. Scenario - the Apache 2.2 way

```
<Directory /var/www/>
	Options Indexes FollowSymLinks MultiViews
	AllowOverride All
	Order allow,deny
	Allow from all
</Directory>
```
### 3. Scenario - the Apache 2.4 way

```
<Directory /var/www/>
	Options Indexes FollowSymLinks MultiViews
	AllowOverride All
	Require all granted
</Directory>
```

Scenario 2 and 3 are working. Scenario 1 isn't. But since [this link](Require all granted) is telling me to use the third scenario, I thought it would be better to have this also in the official magento2 docs.

Greets! :)

### DIFF
--- a/guides/v2.0/install-gde/prereq/apache.md
+++ b/guides/v2.0/install-gde/prereq/apache.md
@@ -187,7 +187,6 @@ For example:
 	<Directory /var/www/>
 		Options Indexes FollowSymLinks MultiViews
 		AllowOverride <value from Apache site>
-		Order allow,deny
 		Require all granted
 	</Directory>
 


xcomSteveJohnson:
@butcher211  Thank you for your contribution. We look forward to hearing from you again. I'm closing this without merging for these reasons:

*  Your discussion talks about an Apache 2.4 configuration that doesn't work but your "magento2-doc way" example doesn't match the current [Apache 2.4 documentation](http://localhost:4000/guides/v2.0/install-gde/prereq/apache.html#apache-rewrites2.4)
* Things like this probably depend on what you're trying to accomplish. The examples for Apache 2.2 and 2.4 in the documentation work for us; maybe we have a simpler Apache configuration than you have.

What I ended up doing is adding a note for both Apache [2.2](http://devdocs.magento.com/guides/v2.0/install-gde/prereq/apache.html#apache-rewrites2.2) and [2.4](http://devdocs.magento.com/guides/v2.0/install-gde/prereq/apache.html#apache-rewrites2.4) with links to Apache documentation for more information about .htaccess and the Order directive.

If you're not happy with this resolution, please reopen or resubmit the PR with more details.

Thanks again!

butcher211:
Hello @xcomSteveJohnson,

I just wanted to ask if you are sure, that you are talking about [this](https://github.com/magento/devdocs/blob/develop/guides/v2.0/install-gde/prereq/apache.md#solving-403-forbidden-errors) or [this](http://devdocs.magento.com/guides/v2.0/install-gde/prereq/apache.html#403-apache)?

There is a section "Solving 403 Forbidden errors for Apache 2.4" and another section called "Solving 403 Forbidden errors for Apache 2.2". I was referring to these sections.

I have the default ubuntu configuration files with the 3rd scenario directive of my first comment and everything just works fine. I just wanted to say that.

Greets.

xcomSteveJohnson:
Thanks for pointing that out. I added the same note to the 403 section.

I have never seen a 403, probably because my Magento 2 setup isn't very sophisticated. I believe that the Order directives you use probably depend on what you're trying to accomplish. Rather than remove it entirely, I think it's better to provide information so the administrator can decide what to do themselves.

butcher211:
Hello,
yes, you're right.

Issue can be closed if you want.

Thank you

PR #467, opened 2015-12-29
correct static files permissions

magenx:
working as root, but run command as web user.


xcomSteveJohnson:
@magenx Thanks! Your suggestion applies to all CLI commands. It concerns me a little bit that it's for bash shell only and assumes bash is in /bin/ but I included it. 

[Example1](http://devdocs.magento.com/guides/v2.0/config-guide/cli/config-cli-subcommands-cache.html#config-cli-before), [Example2](http://devdocs.magento.com/guides/v2.0/config-guide/cli/config-cli-subcommands-compiler.html#config-cli-before)

I am closing this PR without merging because I added the modification to all CLI commands. Please check it and let us know what you think.

PR #484, opened 2016-01-19
Update create_component.md

jeffminor:
Updating to reflect need to xsi:noNamespaceSchemaLocation

### DIFF
--- a/guides/v2.0/extension-dev-guide/create_component.md
+++ b/guides/v2.0/extension-dev-guide/create_component.md
@@ -44,7 +44,7 @@ The smallest working module.xml file would look something like this:
 To be able to properly install the extension the module.xml must include a reference to xsi:noNamespaceSchemaLocation:
 
 	<?xml version="1.0"?>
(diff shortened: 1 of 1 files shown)


tanberry:
Thanks, @jeffminor. @imenikov, can you please make a quick tech review of this contribution? Thanks!

guides/v2.0/extension-dev-guide/create_component.md
```diff
@@ -41,6 +41,18 @@ The smallest working module.xml file would look something like this:
 
 ...where `name`  is the name of your component, and `setup_version` is your module's database schema version. Both of these attributes are required.
 
+To be able to properly install the extension the module.xml must include a reference to xsi:noNamespaceSchemaLocation:
```
melnikovi: @jeffminor thank you for your contribution. I was not able to reproduce error you mentioned in the update to documentation. Also we use URNs to the XSD files instead of full pathes to do not make application depend on the directory structure.

Please update your pull request to only change config node and the formatting. Correct config node is below
<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Module/etc/module.xsd">

Thank you.

PR #485, opened 2016-01-20
Remove leftovers in examples

anatolykazantsev:
It looks the article was copied from another resource and pieces of an old markup remained in some examples.


keharper:
Thanks for finding these errors and fixing them.

PR #487, opened 2016-01-20
File permissions update

krasnuydyx:
Permissions update
- By 770 and 660 you are saying to system to make full rights to the owner and group.
But this is incorrect, as you are writing that group apache or www-data should not have full access.
So 750 and 650 will be fine (full access for owner and read + execute for group).


xcomSteveJohnson:
Thank you for your contribution. Closing without merging because, although this might make sense in some scenarios:

*  We're providing advice before installing Magento so the permissions have to work in all scenarios 
*  When Magento generates files, they have the permissions shown in [this topic](http://devdocs.magento.com/guides/v2.0/config-guide/cli/config-cli-subcommands-mode.html#config-mode-over) (it depends on the mode)
*  The group must have full access in developer mode

PR #488, opened 2016-01-20
Incorrect description of an example

anatolykazantsev:
Description of an example of adding an extension attributes uses names not presented in the example


tanberry:
Thanks again, @anatolykazantsev!

PR #495, opened 2016-01-25
Mention which Composer binary Magento 2 uses

joshuaswarren:
Quick PR based on this tweet - https://twitter.com/benmarks/status/691730733891743744


tanberry:
Thanks so much, @joshuaswarren. There are a couple of StackExchange conversations and tweets (what a day) that I was just lining up for some PRs. :-/

PR #515, opened 2016-02-07
Correct grunt configuration path

AurelienLavorel:


dshevtsov:
@AurelienLavorel thanks!

PR #521, opened 2016-02-10
Array vs [] fix

biotech:
After fresh install of CE& EE 2.0.2 env.php have all config with arrays, this why [] not works there and should be array() - then works correct

### DIFF
--- a/guides/v2.0/config-guide/redis/config-redis.md
+++ b/guides/v2.0/config-guide/redis/config-redis.md
@@ -60,19 +60,22 @@ Installing and configuring the Redis software is beyond the scope of this guide.
 <h2 id="config-redis-config">Configure Magento to use Redis</h2>
 Following is a sample configuration to add to `<your Magento install dir>app/etc/env.php`:
 
-	'cache' => [
-		'frontend' => [
-			'page_cache' => [
-				'backend' => 'Cm_Cache_Backend_Redis',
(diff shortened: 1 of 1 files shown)


davidalger:
@biotech I'm curious what problems you ran into with the config as currently documented, because mixing and matching the `array(…)` and `[…]` syntaxes doesn't affect the end result and is valid PHP code. IMO the documentation here is still correct. Maybe you'd missed a comma on the initial copy/paste?

AlexPernot:
I haven't tried the current documentation code, but isn't there 4 ']' missing anyway? I know it was confusing for me.

davidalger:
@AlexPernot Well, that's a good point! Now I what felt wrong when I was looking at that diff ;) So maybe it'd be good to update the PR to add the closing brackets… @biotech do you want to do that? :)

xcomSteveJohnson:
@davidalger @AlexPernot @biotech Would you mind checking [the topic](http://devdocs.magento.com/guides/v2.0/config-guide/redis/config-redis.html#config-redis-config) again? I've made some recent updates.

davidalger:
With the updates, I think this is all irrelevant now. Looks good to me :)

--
David Alger
Director of Technology, Classy Llama

xcomSteveJohnson:
Closing, thanks!

PR #522, opened 2016-02-11
Its "themes.js" not "theme.js"

fahadonline:
There does not appear to be any "theme.js" file in "dev/tools/grunt/configs/",
it seems the correct one is "themes.js".


almarchenko:
Thanks!

PR #523, opened 2016-02-11
Php5670 php55 eol

magenx:
PHP 5.5 EOL . 
proper repository configured
PHP 7.0 installation

### DIFF
--- a/guides/v2.0/install-gde/prereq/php-centos.md
+++ b/guides/v2.0/install-gde/prereq/php-centos.md
@@ -74,7 +74,7 @@ To upgrade PHP version:
 **CentOS 6 php 5.6**:
 		
 		rpm -Uvh http://rpms.famillecollet.com/enterprise/remi-release-6.rpm
-		yum --enablerepo=remi,remi-php56 -y install php php-pecl-redis php-pecl-lzf php-pecl-geoip php-pecl-zip 				php-pecl-memcache \
+		yum --enablerepo=remi,remi-php56 -y install php php-pecl-redis php-pecl-lzf php-pecl-geoip php-pecl-zip                               php-pecl-memcache \
 		php-cli php-common php-fpm php-opcache php-gd php-curl php-mbstring php-bcmath php-soap php-mcrypt php-mysqlnd php-pdo 		php-xml php-xmlrpc php-intl
 
 <h2 id="instgde-prereq-php70-install-centos">PHP 7.0 on CentOS</h2>
--- a/guides/v2.0/install-gde/prereq/php-centos.md
+++ b/guides/v2.0/install-gde/prereq/php-centos.md
@@ -12,7 +12,6 @@ redirect_from: /guides/v1.0/install-gde/prereq/php-centos.html
 <!-- This topic is referred to from Magento 2 code! Don't change the URL without informing engineering! -->
 <!-- Referring file: README.md owned by core -->
 
-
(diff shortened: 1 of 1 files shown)


xcomSteveJohnson:
Thanks for the suggestion; please be patient. It might take a while for me to verify everything, esp. with CentOS 7. We'll keep you updated.

xcomSteveJohnson:
Thanks for your patience. We're finally ready to make this change. Be aware that we're going to change your PR quite a bit; namely:

*  Add instructions for Ubuntu
*  Change the example repository to IUS. I've read it's better maintained and might be preferable to webtatic and remi. Also clarified that if you're a developer, the choice of repository might not matter that much.

xcomSteveJohnson:
Oops, very sorry. I did not notice you were merging to develop, which is a branch we no longer use. Anyway, check out the changes [here](http://devdocs.magento.com/guides/v2.0/install-gde/prereq/php-centos.html)

magenx:
as always magento team is running ahead of a train and making stupid things... is it something that you are specially trained in the company? :)

IUS repository is not better than webtatic, outdated builds, missing packages, not compatible with latest libs. and mostly selectively maintained for their own corporate needs, not for the community and web developer either.

this is why we have stopped using it many years ago. 

I think it's time for you to have to get out of the tank and start looking around properly. seriously.

xcomSteveJohnson:
We take the responsibility of helping our community seriously. When writing any particular instruction, our goal is to provide a *reasonable* path to achieve a goal, not all possible paths. Particularly regarding third-party software like PHP, there are often no clear-cut answers.

The reason I decided to document IUS rather than webtatic, remi, or others follows:

*  The [CentOS recommended repository list](https://wiki.centos.org/AdditionalResources/Repositories) lists Epel and IUS under Community Approved Repositories, saying:

  bq. These repositories are frequently recommended by the community, are usually well maintained, and provide a substantial number of additional packages to CentOS. They are still not associated with CentOS, but are independent.

*  Webtatic is listed under Problem Repositories as follows:

  bq.  Provides updated packages with later versions of PHP, MySQL, and other packages. IUS or SCL are better alternatives.

*  [reddit](https://www.reddit.com/r/linuxadmin/comments/3o4p0q/ius_vs_webtatic/) supports IUS, albeit from one of the maintainers of IUS.

*  A commenter on [serverfault](http://serverfault.com/questions/456968/how-do-i-upgrade-to-the-latest-php-version-in-centos-with-yum) says IUS introduces "dependency hell" but there are favorable comments there also.

*  I installed PHP 5.5, 5.6, and 7.0.3 from IUS on CentOS 6 with few issues. Magento installs without issues.

*  We state ion our [CentOS PHP topic](http://devdocs.magento.com/install-gde/prereq/php-centos.html) that you have choices: upgrade CentOS, rely on a hosting provider, or choose a repo. We state explicitly that Magento does *not recommend* a repo, including IUS, but we provide the instructions because it's a reasonable way to do it.

I see you disagree with this choice but then again, you rely on your own experience and wouldn't have used our instructions anyway. Neither myself nor Magento as a whole has anything bad to say about webtatic; our sysadmins say they use it frequently.

In merging your PR, we acknowledge your contribution to our documentation. Thank you and we hope to hear from you again.

magenx:
magento 2 have redis and memcached support. how exactly your "sysadmins" configured servers with php 7.0 from IUS and Webtatic to test magento 2???   hmmm

PR #529, opened 2016-02-16
add link to index config guide

bh-ref:
you may also remove the link entirely, since the same link is already mentioned in step 4


xcomSteveJohnson:
Thanks!

PR #531, opened 2016-02-18
Fix @magento-import -> @magento_import

MidnightDesign:


almarchenko:
Thank you for noticing and correcting!

PR #541, opened 2016-02-24
Update footer.html

rafaelstz:
Class of buttons are diferents.

Before:
image

After:
image


xcomSteveJohnson:
Our footer is maintained by the web engineering team and they are about to send us a PR. We appreciate the update!

We have to stay in sync with the magento.com footer, which is about to change apparently.

PR #546, opened 2016-02-26
Flawed Description

fahadonline:
Existing description seemed to be flawed, suggested more accurate one.


almarchenko:
You are right, thank you!

PR #547, opened 2016-02-26
Example for third level of nesting.

fahadonline:
It will be nice to include example for third level of nesting as well.


almarchenko:
Thanks for your input!

PR #552, opened 2016-03-01
Tweaks to example code blocks in factories.md

davidalger:


xcomSteveJohnson:
@davidalger Question. If I do not end a PHP code example with ?>, Sublime Text highlights all the text after the code sample in a strange color like purple. Also, do you know if there is an advantage in using `{% highlight php startinline=true %}` with PHP? I think it enables you to omit the `<?php` or something.

Can you confirm?

xcomSteveJohnson:
@davidalger Answered my own question. Using `{% highlight php startinline=true %}` means you can leave out both `<?php` and `?>` and Sublime Text is happy. I'll merge this and update it soon.

davidalger:
@xcomSteveJohnson Sorry I didn't get back sooner…was tied up all day yesterday with a client visit. I've not used sublime much, but that is good to know for the future that way stuff looks nice in the editor _and_ online, so thanks for posting the solution :)

PR #556, opened 2016-03-04
Found inconcistancy in settings.xml.dist location

Got-iT-Services-Inc:


dshevtsov:
@Got-iT-Services-Inc thank you!

PR #558, opened 2016-03-04
Added clarification, spell check

Lodewyk:
Just fixed a small spelling mistake, and added some clarification that I think might be helpful for other users. As a complete noob with Magento I was unsure which config.xml was being referred to.


xcomSteveJohnson:
Thanks!

PR #568, opened 2016-03-13
there 3 possible modes: default, developer and production

ClaudiuCreanga:
Sometimes you may want the default one.


xcomSteveJohnson:
@ClaudiuCreanga I'm pretty sure `default` is not an option in this command.

PR #570, opened 2016-03-17
add missing php70u-json

csmarvz:
Without it, run into
[RuntimeException]
PHP's json extension is required to use Monolog's NormalizerFormatter


xcomSteveJohnson:
Thanks! Good catch!

PR #571, opened 2016-03-17
Remove steps for restarting web server

ryebell:
Assuming that "AllowOverride All" is already properly set so that .htaccess files can be read by Apache on each request, changes to .htaccess files shouldn't require a full restart/reload of the web server.

### DIFF
--- a/guides/v2.0/config-guide/secy/secy-cron.md
+++ b/guides/v2.0/config-guide/secy/secy-cron.md
@@ -103,13 +103,8 @@ To add security for cron in Magento's `.htaccess`:
     		Require group <name>
 		</Files>
 4.	Save your changes to `.htaccess` and exit the text editor.
-5.	Restart Apache:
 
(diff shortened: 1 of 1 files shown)


xcomSteveJohnson:
Thank you for your contribution but I'm going to close this without merging. I have always changed AllowOverride from None to All. I have not seen it set correctly by default; however, I will add some text to say that you have to restart only if you did change its value.

ryebell:
Thanks Steve! Mostly just wanted to clear up any confusion in regards to restarting the web server, which is nested in Step 3 alongside the .htaccess modifications, since it can be incorrectly interpreted that the change itself will require a restart to be properly read by the web server.

xcomSteveJohnson:
My fault. Apologies. You're correct. My GitHub mojo isn't sufficient to back this out and merge your PR but I'm going to make the change anyway.

I didn't notice which file you changed. Again, my fault.

PR #572, opened 2016-03-18
Correct order for example replacing core js file

fooman:


almarchenko:
Thank you, good catch!

fooman:
Thanks this was driving me crazy until someone else pointed out where I went wrong.

almarchenko:
Oh I'm sorry to hear that( Then even more reasons to say "thank you" for the pr!

PR #575, opened 2016-03-18
Styles fallback schema Problem

virani06:
when if not found css on direct `<parent_theme_dir>/web/css/` than they find on `<parent_theme_dir>/<Namespace>_<Module>/web/css/` than only find  Module styles for the


almarchenko:
You right! Thank you for catching this!

PR #576, opened 2016-03-21
Add PHP 7.0.2 to supported PHP versions

dank00:
See http://docs.magento.com/m2/ce/user_guide/magento/release-notes-ce-2.0.1.html


xcomSteveJohnson:
Good catch!

PR #577, opened 2016-03-22
Fix filename _extend.less

brendanmckeown:
Changed all references to `_extend.less` to `_extends.less`. See `styles-m.less` and `styles-l.less` in the Magento blank theme for the import statement: 
```
@import (reference) 'source/_extends.less';
```


brendanmckeown:
I later discovered that `_extend.less` actually works, so please ignore this PR.

PR #582, opened 2016-03-28
Added sudo example to Update first-steps-cli.html

pynej:
Added a example using the Magento command line tool in environments with ``sudo`` where the Magento file system owner account may not allow logins directly.


xcomSteveJohnson:
Thanks!

PR #586, opened 2016-03-30
Document running PHP unit tests

Vinai:
This PR does four things:

1. Make the page `mtf/mtf.md` the main entry into testing. I didn't introduce a new page in order to maintain existing external links to the page. This page can now also act as a place from where to link to further testing related documentation yet to be written.
2. Change references to the functional testing framework to from "Magento Testing Framework" to "Functional Testing Framework" to be able to distinguish between the functional testing framework and for example the integration test framework.
3. Add instructions on running PHP unit tests on the command line and in PHPStorm.
4. Add a few more cross references to existing testing related documentation so it is easier to find.


guides/v2.0/test/unit/unit_test_execution.md
```diff
@@ -0,0 +1,31 @@
+---
+layout: default
+group: test-guide
+subgroup: A_Running_Unit_Tests
+title: Running Unit Tests
+menu_title: Running Unit Tests
+menu_node: parent
+github_link: test/unit/unit_test_execution.md
+redirect_from: /guides/v2.0/test/unit/unit_test_execution.html
```
xcomSteveJohnson: Pls remove `redirect_from` because it results in constant redirects when going to this page.

guides/v2.0/test/unit/unit_test_execution_cli.md
```diff
@@ -0,0 +1,91 @@
+---
+layout: default
+group: test-guide
+subgroup: A_Running_Unit_Tests
+title: Running Unit Tests in the CLI
+menu_title: Running Unit Tests in the CLI
+menu_node: parent
+github_link: test/unit/unit_test_execution_cli.md
+redirect_from: /guides/v2.0/test/unit/unit_test_execution_cli.html
```
xcomSteveJohnson: Pls remove `redirect_from` because it results in constant redirects when going to this page.

guides/v2.0/test/unit/unit_test_execution_phpstorm.md
```diff
@@ -0,0 +1,107 @@
+---
+layout: default
+group: test-guide
+subgroup: A_Running_Unit_Tests
+title: Running Unit Tests in PHPStorm
+menu_title: Running Unit Tests in PHPStorm
+menu_node: parent
+github_link: test/unit/unit_test_execution_phpstorm.md
+redirect_from: /guides/v2.0/test/unit/unit_test_execution_phpstorm.html
```
xcomSteveJohnson: Pls remove `redirect_from` because it results in constant redirects when going to this page.

guides/v2.0/test/unit/unit_test_execution.md
Vinai: Removed.

guides/v2.0/test/unit/unit_test_execution_cli.md
Vinai: Removed.

guides/v2.0/test/unit/unit_test_execution_phpstorm.md
Vinai: Removed.

guides/v2.0/extension-dev-guide/test-module.md
```diff
@@ -0,0 +1,37 @@
+---
```
xcomSteveJohnson: There appears to be an issue with this and `validate/validate.md`. Please remove `redirect_from` from `validate/validate.md` and look at the Ext Dev Guide in Jekyll.

here's what it looks like. 

image

guides/v2.0/extension-dev-guide/test-module.md
```diff
@@ -0,0 +1,37 @@
...
+
+{% highlight JSON %}
+"require": {
+    "magento/magento-composer-installer": "*",
+    "magento/product-community-edition": "2.0.0",
+    "foovendor/module-one": "0.1.1"
+},
+{% endhighlight %}
+
+Remember to [register](component-registration.html) your component as well, adding the location of your component. Confirm that your component works as expected and Magento functionality is not compromised.
```
xcomSteveJohnson: Please change the cross-ref to `[register]({{ site.gdurl }}extension-dev-guide/build/component-registration.html)`

guides/v2.0/extension-dev-guide/test-module.md
Vinai: Fixed.

guides/v2.0/extension-dev-guide/test-module.md
Vinai: Changed.

guides/v2.0/extension-dev-guide/test-module.md
```diff
@@ -0,0 +1,37 @@
...
+title: Test your component
+menu_title: Test your component
+menu_order: 2
+menu_node: 
+github_link: extension-dev-guide/test-module.md
+
+---
+
+##{{page.menu_title}}
+
```
xcomSteveJohnson: Pretty minor but the TOC doesn't match the sections right now. Try adding after {{page.menu_title}}:

```
* TOC
{:toc}
```

guides/v2.0/extension-dev-guide/test-module.md
```diff
@@ -0,0 +1,37 @@
+---
+layout: default
+group: extension-dev-guide
+subgroup:
```
xcomSteveJohnson: Two problems. `subgroup` needs to be there; otherwise, the topic displays right under the guide name on the left. Second, there are still two topics, one in `ext-dev-guide` and one in `validate`. Seems like there should be just one file and it should be under `validate`. If I'm offbase, please let me know.

Thanks for the outstanding contribution!!

guides/v2.0/extension-dev-guide/validate/test-module.md
```diff
@@ -16,17 +16,18 @@ redirect_from:
```
xcomSteveJohnson: Please remove `redirect_from` from this topic.

guides/v2.0/extension-dev-guide/test-module.md
Vinai: Yes, you are right. I removed the one and only left the one under `validate`.

guides/v2.0/extension-dev-guide/validate/test-module.md
Vinai: Removed

xcomSteveJohnson: Thanks! Everything looks good now. We’ll have some more feedback for you soon.

xcomSteveJohnson:
Discussed it with our Kyiv MTF writer, CC @dshevtsov and we have a couple of concerns:

*  Even though the name MTF isn't descriptive, we can't eliminate it entirely. MTF has its own GitHub repository after all. This is something we writers can do, suggest we mock something up and show it to you (see questions at the end). 
*  In one place, you should examples for running commands in OS X, which isn't supported for production. AFAIK, we don't document OS X anywhere else in devdocs so we'd like to replace your examples with some for CentOS and Ubuntu. Again, this is something we can do but please give us your thoughts; we won't do anything until we hear from you.
* We're still discussing having the **Developers** > **Magento Testing Guide** land in the How Do I book and the fact that the existing How Do I topic is removed. We might propose an alternative, which we'd let you preview first. (Basically, let that link go to what looks like a guide with the same contents and, on How Do I, have two links--one to the testing guide and the other with the current `mtf.md` topic.

Question: My GitHub mojo isn't much but what I'm hoping to do is:

1.  Create a branch on devdocs from your PR.
2.  Make proposed changes there.
3. You can pull them into your original branch or a new branch, run Jekyll, and let us know what you think about the changes, which will all be visible in the branch.

Do you think that'll work?

Vinai:
Thanks for the update!

> Even though the name MTF isn't descriptive, we can't eliminate it entirely. MTF has its own GitHub repository after all. This is something we writers can do, suggest we mock something up and show it to you (see questions at the end).

My though was that the acronym MTF could remain in use, but in written form could be referred to as "Functional Testing Framework" (since that is what it is).   
Back when I was reading the docs for the first time, I remember that "Magento Testing Framework" confused me a lot. Based on the name I was expecting something quite different (I've read similar comments from others, too).

Given the testing pyramid that usually is recommended (with the number of functional tests being an order magnitude fewer then the number unit and integration tests) plus the trend to not use much functional testing at all (see http://googletesting.blogspot.de/2015/04/just-say-no-to-more-end-to-end-tests.html for example), giving the functional testing framework a more descriptive name then "Magento Testing Framework" would be a good idea.

> In one place, you show examples for running commands in OS X, which isn't supported for production. AFAIK, we don't document OS X anywhere else in devdocs so we'd like to replace your examples with some for CentOS and Ubuntu. 

The examples are for running unit tests, which is something that isn't done on production. It is done on development systems, the majority of which are OS X and Windows. The reason I added those sections are because those are typical issues people encounter when starting out trying to execute the tests.  
The intended purpose of the text I wrote is to help developers start using the tests, so I think it makes sense to address the common issues and pitfalls, even if they don't apply to the supported production environment.

> We're still discussing having the Developers > Magento Testing Guide land in the How Do I book and the fact that the existing How Do I topic is removed

The existing "How Do I" topic isn't removed, I moved it to `mtf/mtf_overview.md`.
The idea behind keeping the existing URL in place was to keep deep links intact. When looking for documentation on testing, that is the most prominently visible place, so it would make a good starting page that links to all other testing articles.

To create a new local branch based on my PR you can run the command

```bash
$ git fetch origin pull/586/head:2.0-unit-test-docs
```
The run

```bash
$ git checkout 2.0-unit-test-docs
```

Then you can modify the branch locally, commit the changes, and when you are done you can push it to the github devdocs repo using

```bash
$ git push origin 2.0-unit-test-docs
```

I'm not sure if that will keep me as the author, but to be honest, that is okay. I care more about pushing for the adoption of testing then the reputation on github.

xcomSteveJohnson:
*  I see your point about OS X so we'll investigate. If Ubuntu and CentOS are the same, great; otherwise, we can add more. Do you know the Windows commands? If not, no big deal.

*  I'll mock something up for you and let's go from there. My feeling about MTF, for example, is to minimize the use of the name. I need to think some more about the suggestion on How Do I.

I want to make sure you get credit because it's the right thing to do. We'll also put your name on all of your topics. If you can suggest blogs and videos that would help, please provide links.

Vinai:
> If Ubuntu and CentOS are the same, great; otherwise, we can add more. Do you know the Windows commands? If not, no big deal.

Under (any) Linux PHP will usually be under `/usr/bin/php` or `/usr/local/bin/php`, depending on the distribution, but there are many variations. Not sure about CentOS and Ubuntu,
I've worked with Windows in the past, but there are so many variations how and where PHP can be installed, I don't know which paths to suggest. It depends on how PHP was installed...

Thanks!

xcomSteveJohnson:
You can always use `which php` to find the binary on Linux correct? If so, that's the solution. Should have thought of it earlier.

Vinai:
The `which php` command will only give you the first PHP in your path. For development systems that often isn't the one that is used with Magento. Or it might return nothing, if PHP is installed but not in a directory that is in `$PATH`. That is one possible reason for the issues that can crop up for developers starting out with testing.

xcomSteveJohnson:
I made some edits to your topics under `/unit` and I did as you suggested and pushed them to `2.0-unit-test-docs`. You'll see your name as a contributor at the top of each topic. I'll work in links to your blogs and videos also.

There are a fair number of text edits, plus I updated OS X and also the PHP Storm topic. Let me know if you have any concerns. I'll make some more changes early in the week.

Vinai:
Thanks very much for the link to my site in the header! That is very generous of you.
Your changes look very good indeed. Thanks for fixing the width on the screenshots, too - I tried to figure out the right way to do that but failed.
For the link to my site, how much effort is it to change that? I'm planing to add an overview page for the Mage2Kata blog topics. Once that is there that might be a better link target.

xcomSteveJohnson:
Changing the link is super simple. There are two values in the Jekyll front matter: `contributor_name` and `contributor_link`. Feel free to change them yourself.

Vinai:
Thank you, I'll do that once I've got a new page to link to :)
Is there any more feedback you would like from me or are there any changes in regards to this PR you would like me to make at this time?

xcomSteveJohnson:
No. I'm going to look at a few more things this morning so stay tuned.

xcomSteveJohnson:
I made some changes to the `howdoi/mtf/mtf.md` page and set the MTF guide back to what it was. I'm going to get feedback from my colleagues. I'm a little torn about the name of the guide so we could change that again.

(It's unfortunate that MTF doesn't have "functional" in the name.) I'll see what everyone thinks, including you, before finalizing it.

Please take a look.

Vinai:
Thanks for the changes. Regarding the name, how about prefixing it with "Functional" whenever it is written fully.  
So the full name would be "Functional Magento Testing Framework" whenever it is spelled out.  
It's not such a large change as changing "Magento Testing Framework" to "Functional Testing Framework", but rather making what already has been established more descriptive. The only downside is the length. But it won't be as confusing as it is currently. I'm sure that readers will appreciate it.

The description of the repository https://github.com/magento/mtf should be changed to include "Functional", too.

xcomSteveJohnson:
@Vinai Our colleagues haven't discussed renaming the MTF guide so I think that for now we'll leave it as Magento Testing Framework Guide. It's super easy to change so it's something that can be done later. I'll try to merge this before Imagine. 

Let you know any other caveats.

xcomSteveJohnson:
Thanks again, @Vinai !!! This is a fantastic addition.

PR #595, opened 2016-04-09
Add instructions on running integration tests

Vinai:


xcomSteveJohnson:
@Vinai Thanks again! BTW, word to the wise ... in your Jekyll front matter, its muuuuuch easier to order topics using numerals than letters.

For example, instead of:

`subgroup: B1_Running_Integration_Tests`

Use something like

`subgroup: 02_Running_Integration_Tests`

PR #598, opened 2016-04-09
2.0 staffan

vaimo-staffan:
Simplified the contribution process by removing the instruction to create a branch after creating a fork. Also small fixes in the table of contents.


xcomSteveJohnson:
Thanks!

PR #599, opened 2016-04-10
Correct inconsistent Hello World / Index string.

pixiesky:
In the example code the instruction is to write "Hello Index!" but the following copy states that we "should see a blank page with the text "Hello World"" on our page.


xcomSteveJohnson:
Thanks!

pixiesky:
You're welcome.

PR #602, opened 2016-04-10
Add article on creating testable code

Vinai:
I've done my best to make it smooth, but I think this PR could still really need some help from a native speaker to make it more readable.


xcomSteveJohnson:
@Vinai As always, many thanks. I'll give it a readthrough on Monday.

xcomSteveJohnson:
@Vinai Wanted to check with you before I made this particular change:

* Your statement: "Consider moving the code in that needs the value into a class that has the data available."
* My revision: "If a class requires data, consider moving code that requires the data into the same class."

What do you think?

xcomSteveJohnson:
@Vinai I went with my wording but let me know if it's incorrect and I'll fix it.

Thanks again!

PR #607, opened 2016-04-13
Expanding file permissions to address servers hosting multiple domains

gamort:
Corrects file permissions when dealing with Virtual Host environments.  In a virtual host environment Apache will be run as 2 different users.  Assuming mydomain.com:

Apache will run as both apache:www-data and mydomain:mydomain

In this case, the closest match to the single server/single domain assumption made in the docs is for the magento user and the mydomain user to share a group which is not shared by the apache user.

This is problematic if the Apache server uses mod_rewrite[and frankly do we expect it ever to NOT use it?].  With mod_rewrite, Apache will scan for .htaccess files before switching to to the mydomain user and after switching to the mydomain user as many other extensions - most notably mod_security2 - need to check for configuration adjustments to global rules.

That means the .htaccess file is an exception to the 660 file permission - it instead must be 664.  Note: Magento actually does set the correct file permission on .htaccess it simply does not document it.

In addition - a frequently missed setting is that in order to perform a recursive check up the directory tree for .htaccess files the apache user will need search/list permissions for the entire directory tree.  
ie an attempt to load the file pub/static/frontend/Magento/luma/en_US/js/theme.js will use the .htaccess file located at pub/static/.htaccess    However, Apache will first check for html/pub/static/frontend/Magento/luma/en_US/js/.htaccess, then html/pub/static/frontend/Magento/luma/en_US/.htaccess, then html/pub/static/frontend/Magento/luma/.htaccess, etc

This requires search permissions - otherwise known as the X bit [for eXtra not eXecute!]

As such, I've modified the documentation with the correct information as well as changing the wording to avoid using the word 'execute' since many people have trouble grasping that it is not possible to execute a directory.

As there is no security implication with enabling the search bit on directories for all users - there really is no need to have a separate set of instructions for the 2 different environments.

### DIFF
--- a/_includes/install/file-system-perms.html
+++ b/_includes/install/file-system-perms.html
@@ -1,22 +1,51 @@
 <h2 id="install-perms-import">Why we recommend you set file system permissions</h2>
 <P>Malicious exploits are an unfortunate reality in the internet age. To help prevent exploits that take advantage of the file system, we recommend you set Magento file system ownership and permissions in a particular way. For more information, see <a href="{{ site.gdeurl }}install-gde/prereq/apache-user.html#install-update-depend-user-over">Overview of ownership and permissions</a>.</p>
+<p>On web servers that host multiple domains, it is common practice for each domain to have a different web server user.  This allows some limited protection in that the web user for one domain is not able to create, and more importantly read, files stored in the other domain(such as the server configuration containing the database logon information!).  This means that there are now two web server users that may need to access files for your system, the domain user and the system user.  In general, the system user will require read access to any .htaccess files as well as search access to directories containing static resources.</p>
+
 <p>The important things:</p>
 <ul><li>The owner of the Magento file system:
 <ul><li>Must have full control (read/write/execute) of all files and directories.</li>
 <li>Must <em>not</em> be the web server user; it should be a different user.</li></ul>
 </li>
-<li>The web server user must have write access to the following files and directories:
-<ul><li><code>var</code></li>
-<li><code>app/etc</code></li>
-<li><code>pub</code></li></ul>
+<li>The web server domain user:
+	<ul>
+		<li>Must have write access to the following files and directories:
+			<ul>
+				<li><code>var</code></li>
+				<li><code>app/etc</code></li>
+				<li><code>pub</code></li>
+			</ul>
+		</li>
+		<li>Must have search[execute] access to the following directories[including any subdirectories]:
+			<ul>
+				<li><code>pub</code></li>
+			</ul>
+		</li>
+	</ul>
 </li>
+	<li>The web server system user:
+		<ul><li>Must have search[execute] access to the following directories[including any subdirectories]
+			<ul>
+				<li><code>pub</code></li>
+			</ul>
+		</li>
+			<li>Must have read access to any .htaccess files in the following directories[including any subdirectories]
+				<ul>
+					<li><code>pub</code></li>
+				</ul>
+			</li>
+		</ul>
+    </li>
 </ul>
-<p>In addition, the web server's <em>group</em> must own the Magento file system so that the Magento user (who is in the group) can share access to files with the web server user. (This includes files created by the Magento Admin or other web-based utilities.)</p>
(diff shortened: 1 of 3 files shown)


xcomSteveJohnson:
Thank you for your contribution. Please expect this to take a little while to validate. We'll get back to you.

sholdowa:
- Magento supports multiple web servers, not all of which use .htaccess
- The stack the site is running on may use different users for (eg) webserver and php
- nobody requires write access to .htaccess in normal use - so 444 ( or 440 with proper use of groups ) is more than adequate.

Also, you do not need the read bit set to test the existence of a file with a known name if walking up a directory tree. The X - execute bit allows you to access a file within that directory, NOT list it's contents ( which is what is required to search ). Please correct your posting.

gamort:
@sholdowa please review within the context given.  The instructions already call for setting file permissions to 660 - including the .htaccess file.  I merely adjusted them to make the .htaccess file usable.   I'm not about to fix everything in the documentation in one step.

gamort:
@sholdowa Magento 2 may support other web servers, however the permissions are documented specifically for Apache/mod_rewrite    Under Apache if the server is not able to search for and read .htaccess files Magento 2 breaks.   Making sure .htaccess files are readable and directories which may contain them are searchable is neccessary.   Under a platform such as nginx setting the X bit on the directory does not cause any problems. 

Also note: for this PR I am limiting it strictly to the use of chmod.   Using setfacl is far preferable.  Then you can provide targeted permissions for each class of Magento 'user' and disable group and other access entirely.  https://wiki.archlinux.org/index.php/Access_Control_Lists

Right now I'm concerned with ensuring things are functional without making security issues worse.  Making them better would depend on the willingness to expand the documentation.

gamort:
@sholdowa  
Search/list is about the best terminology you will get.
 "The X - execute bit allows you to access a file within that directory, NOT list it's contents ( which is what is required to search )."
followed by your example:
"[steve@localhost ~]$ ls -l a
ls: cannot open directory a: Permission denied"

Technically, the X bit is what allows you to change to the directory.   However, being able to change to the directory is a requirement for some, but not all, directory oriented commands - including the searches performed by Apache.

find /path/to/html/pub -type d -exec chmod o+x {} +

Tends to be the safest all purpose command and in my experience actually works when the css files suddenly stop loading after Magento2 has set other access to 0 which it does when switched to production mode or when it creates a new subdirectory for pub/static and is in production mode.

In all honesty, outside of the logfiles, none of the files or directories should be writeable by anyone.  As long as the "Magento" user owns the directories and files - it should make the file writable when it needs to write to it and then switch it back to read only afterwards.  That eliminates a large class of exploits at the cost of a couple extra function calls.   However, since this is the documentation repository - we can't change code here - just document how the code that is there can be made to work.

sholdowa:
"In all honesty, outside of the logfiles, none of the files or directories should be writeable by anyone"
which breaks sessions, caches and media updates

"As long as the "Magento" user owns the directories and files"
Which one? If you mean the one who accesses them via the website, then this is the whole point of my original post. If they own them, then they can override any permissions you put on them ( MACs ). 

So any changes only make the illusion of improved security, as a breach can add / modify any files it wishes.

xcomSteveJohnson:
We're discussing this and I'll provide an update ASAP.

xcomSteveJohnson:
Thank you for your contribution but we're closing without merging for the following reasons:

*  By 2.1 anyway and possibly sooner, Magento is no longer _enforcing_ file permissions at all. We haven't worked out the solution yet so I cannot talk about it here. After being [roundly criticized](http://alanstorm.com/an_open_letter_to_magentos_leaders) by the community for setting permissions in the code, we're going to take a more hands-off approach.  In my personal opinion, we should say _what_ files and directories should be protected or read-write and _why_ but leave _how_ up to the implementer. We'll see what happens.

*  The discussion in this PR and in #609 show there's no one way to advise on security-related permissions. The network administrator or SI is responsible for coming up with a comprehensive security strategy. Magento cannot be responsible for that.

*  Currently, we have [very little documentation](http://devdocs.magento.com/guides/v2.0/config-guide/multi-site/ms_over.html) about how to set up multiple webnodes. We expect to expand it in the future but currently it's not a top priority.

Thank you for thinking of us and for making the contribution. After you see the changes, if you'd like to contribute again, we encourage you to do so.

gamort:
Thank you very much, the new method has allowed us to switch back and forth from developer mode without having to constantly fix permissions!

PR #617, opened 2016-04-20
Changed text to reflect actual content of error msg

sammarcus:
The literal error message (as reflected in the pictures) is "Sorry, we can't take that action right now" (without "we're). In the interest of accuracy and SEO for people searching for the error message verbatim, I removed the word that does not exist in the message being displayed on the platform.


xcomSteveJohnson:
Thanks!

PR #627, opened 2016-04-24
Update to bio page

ryantfowler:
In general, it's not recommended to switch person perspectives in the same paragraph in the English language. It's best to choose one and stick with it consistently. As I read this, it wasn't clear to me which context was intended. If the intent was to speak to the reader, then using second person pronouns consistently would be desirable. If the intent was third person, then using Classy Llama throughout would be desirable. Either way, switching between perspectives within the same paragraph doesn't reflect kindly on the company. I have edited the paragraph to be in first person context...but if it's more desirable to speak to the reader, then modifying the text accordingly would be appropriate. My intentions are only to assist in a positive reflection upon Classy Llama. Thank you


xcomSteveJohnson:
Thanks!

PR #629, opened 2016-04-25
Added the case when the installation hangs while loading data

javiereguiluz:
This just happened to me while installing Magento. I found the solution here: http://magento.stackexchange.com/questions/92976/installation-hangs-while-installing-data


xcomSteveJohnson:
Thanks!

PR #631, opened 2016-04-26
Correct example crontab configuration

davidalger:
Updates include used here: http://devdocs.magento.com/guides/v2.0/config-guide/cli/config-cli-subcommands-cron.html

* Removes redundant `*/1` and uses simpler `*` for minute configuration
* Makes use of log files clearer
* Improves readability of placeholders (following typical snake case format)
* Strips unnecessary output from the log (constant, hey we ran!)
* Removes the trailing `&` from the commands. This is not needed because everything run by crontab is already run in the background, so it's redundant.

### DIFF
--- a/_includes/config/setup-cron.md
+++ b/_includes/config/setup-cron.md
@@ -35,15 +35,14 @@ For example,
 
 A text editor displays. (You might need to choose a text editor first.)
 
-	*/1 * * * * <path-to-binary> -c <ini-file-path> <your Magento install dir>/bin/magento cron:run [>> <log-file>&]
-	*/1 * * * * <path-to-binary> -c <ini-file-path> <your Magento install dir>/update/cron.php [>> <log-file>&]
-	*/1 * * * * <path-to-binary> -c <ini-file-path> <your Magento install dir>/bin/magento setup:cron:run [>> <log-file>&]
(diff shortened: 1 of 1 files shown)


xcomSteveJohnson:
Why did you add `grep -v`? Curious about that. Thanks

davidalger:
Every time `bin/magento cron:run` task executes, it outputs `Ran jobs by schedule` resulting in a meaningless piece of text filling up logs. Normally none of these jobs should output anything (other than this one), so the logs really are for capturing errors if something went wrong. The grep -v silences this particular line of output, but nothing else.

Silencing that green-light message is more important when you set a MAILTO in the crontab, because you don't want an email every minute saying, hey, it works! ;) We use MAILTO on many of our setups, and the one I just configured on a server looks like this actually (notice using tee to both allow it to send an email in addition to logging):

```
MAILTO="user@example.com"

* * * * * cd /var/www/prod/current && bin/magento cron:run | grep -v "Ran jobs by schedule" | tee -a var/log/magento.cron.log
* * * * * cd /var/www/prod/current && bin/magento setup:cron:run | tee -a var/log/setup.cron.log
* * * * * cd /var/www/prod/current && /usr/bin/php -f update/cron.php | tee -a var/log/update.cron.log
```

Note: I've never run across a server where using cd like I do hasn't worked, but I've been told some shells are picky than others. I do it to keep it short since I can.

xcomSteveJohnson:
Done, thanks! I added an explanation of the `grep`.

PR #633, opened 2016-04-27
Add URL to retrieve entire list of service names

aleron75:


keharper:
Thanks for the tip.

aleron75:
my pleasure

PR #644, opened 2016-04-30
Fix backtick typo in PHP memcached extension array

chrone81:
Then:
 'save' => 'memcached`,

Now:
 'save' => 'memcached',


xcomSteveJohnson:
Thanks!

PR #647, opened 2016-05-03
Update version 2.0.4 to 2.0.5

rafaelstz:


xcomSteveJohnson:
We appreciate it but it isn't necessary to change the example every time there's a new patch. Closing.

PR #649, opened 2016-05-04
Update enable-disable-modules.html

ryantfowler:
Removed "you" which didn't seem to fully describe whether or not the sentence should include modules created by a "you" or if that sentence was meant to only include Magento core modules and 3rd party modules....so if I misinterpreted this, my mistake. If that is the case, then the sentence would be more helpful if it included a reference to custom, non-3rd party, modules....the kind that were placed in `app/code/local` codepool in Magento 1. If I was correct in my assumption, then fantastic.

Thanks


xcomSteveJohnson:
Thanks, it was a typo.

PR #651, opened 2016-05-07
Prevent bullets from showing up in ordered lists

netalico:
Before:
screenshot 2016-05-07 07 55 42

After:
screenshot 2016-05-07 08 02 52


xcomSteveJohnson:
Thanks, this fixes the issue you mentioned but another one remains. I'd be very grateful if you could take a look at it: See http://devdocs.magento.com/guides/m1x/install/installing_upgrade_landing.html#upgrade-path, for example. Step 4 has bullets that mash into the text. 

Thank you for your contribution.

PR #655, opened 2016-05-10
Added secion on metadata

vaimo-staffan:
There is still quite a bit of information on how the menu_order works that I have commented out because I'm not sure it's ready to be published. This part could need some more work.

### DIFF
--- a/guides/v2.0/contributor-guide/contributing_docs.md
+++ b/guides/v2.0/contributor-guide/contributing_docs.md
@@ -12,7 +12,6 @@ github_link: contributor-guide/contributing_docs.md
 redirect_from: guides/v2.0/extension-dev-guide/Contribute_edg.md
 ---
 
-
 The following topics are included in the Guide:
 
 * <a href="#contribute">DevDocs contributions</a>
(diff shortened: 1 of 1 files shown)


tanberry:
Whoo hoo, thanks so much Staffan! Our team was just talking last week about the fact that we needed to add info about the meta-data. Perfect timing. ;)
We are having discussions about whether those entries should be called "meta-data" or "front matter." Jekyll calls it front matter, but to me "meta-data" seem more universal, more of an industry standard. Thoughts?
We'll go ahead and merge your contributions, and then we will work on the part you commented out about how the menu_order works. Thanks again!

xcomSteveJohnson:
[Jekyll calls it front matter](https://jekyllrb.com/docs/frontmatter/) but we can alias it as metadata.

vaimo-staffan:
Ah Right! Now I got why it said "front matter"! Feel free to change it if you wish! :)

davidalger:
@xcomSteveJohnson / @tanberry It's not just Jekyll that calls it front matter. :) This seems to be a fairly well known standard known as YFM (or YAML Front-Matter) and is used by many of the static site frameworks out there. My personal choice would be to use only the term front matter. There is plenty out there on the YFM blocks if people go searching for it.

tanberry:
Yep, sounds good @davidalger. Steve had a good suggestion, with "softer" wording... we will use the term front matter, but explain that one can think of front matter as a type of meta-data. But then in remainder of topic, use the term front matter. Thanks all!

PR #656, opened 2016-05-10
Missing endif

macknik:
Added an endif statement to close the wrapping if statement. This code snippet will now work properly when the customer is not logged in.


almarchenko:
Great observation, thanks!

PR #658, opened 2016-05-11
Replaces relative url save string in example with url type

maderlock:
This will construct a valid URL from the action path


xcomSteveJohnson:
Thanks!

PR #664, opened 2016-05-17
Update system-requirements.md

ryantfowler:
Looks like there was a typo for a Php version number, which could get confusing as there is a Magento version number of 2.0.5. If I'm wrong about this assumption, could that line be rewritten so that it's easier to understand?

Thanks


xcomSteveJohnson:
Sorry for the confusion

PR #665, opened 2016-05-17
Update php-ubuntu.md

ryantfowler:
This page should reflect the comments about Php version which are listed here: http://devdocs.magento.com/guides/v2.0/install-gde/system-requirements.html . I have a PR open to possibly clear up some confusion on that page, as well: https://github.com/magento/devdocs/pull/664 . If PR 664 is merged, then my edit above to 7.0.5 is accurate...if it's not, then whatever the outcome of that section of System requirements is, should be reflected here as well.


xcomSteveJohnson:
Merged this with PR#666

PR #671, opened 2016-05-17
Fixed the HTML of the label description table

javiereguiluz:
There were some missing HTML tags that made the table look as follows:

magento-table-before

Now it looks like this:

magento-table-after


tanberry:
Thank you, @javiereguiluz, much appreciated! That looks a lot better.

PR #673, opened 2016-05-18
Add session storage support for Redis

mzeis:
As mentioned here: http://devdocs.magento.com/guides/v2.0/config-guide/redis/redis-session.html


xcomSteveJohnson:
Good one, thanks!

PR #675, opened 2016-05-19
Clarify customers username

Vinai:
Clarify that customers use their email address as a username for authentication.


keharper:
Thanks for the clarification, @Vinai .

PR #676, opened 2016-05-20
Fixed name of configuration-parameter for database

berwa:
The originally parameter 'db' is completely ignored, but this is only apparent if the value of the parameter is changed to anything other than '0' (0 is the Redis-internal default-value).
The proper parameter is 'database' which is also used in the table below.


xcomSteveJohnson:
Sorry about that. Thanks for catching it!

PR #679, opened 2016-05-22
Update config-cli-subcommands-compiler.md

ryantfowler:
Added reference to 2.1 RC1 Release Notes, specifying a fix for the single tenant di compilation command.


xcomSteveJohnson:
I don't think this is a good idea:

- We have no 2.1 documentation yet except for Release Notes; a reference in 2.0 for something available only in 2.1 is confusing

- There's no guarantee that anything in 2.1 RCx will be in 2.1 GA (probably will but not certain)

- An RC is for experimentation only; we provide the means to upgrade but we won't guarantee it will work and might not support it

What do you think?

ryantfowler:
You make valid arguments.

I was just trying to give others insight into the potential upcoming solution for this known issue, since we don't have visibility into the internal issue tracking system. But due to the nature of a release candidate, I see your point. 

I'll close this PR.

Thank you, Sir!

xcomSteveJohnson:
Your comments and input are always appreciated. We agree it's dicey in this case.

ryantfowler:
Pleasure doing business with you ;).

PR #688, opened 2016-05-25
ssh link update

keithbentrup:
duplicate id used, so link to id does not work


xcomSteveJohnson:
I'll get to these shortly. Thanks for your help! I should be back on Cloud today so please keep them coming and I'll do my own checking as well.

PR #702, opened 2016-06-01
replace platform references with magento-cloud

keithbentrup:


xcomSteveJohnson:
Thanks, how embarrassing though!! I thought I had fixed these.

PR #703, opened 2016-06-01
Replaced \Zend\Soap with native \SoapClient

cmuench:
It's more general to use the standard `SoapClient` class of php. It's not needed to use the `\Zend\Soap` component.
The example should show us that it's easy to call the Magento 2 SOAP Service without any special components.


keharper:
Thanks for your input.

PR #755, opened 2016-07-15
fix duplicate observer's name

tawfekov:
I was reading the docs and I found this duplicate observer name


jcalcaben:
It is actually okay to have the same observer name as long as they are defined in different events. 

The paragraph following the example explains this and is meant to reduce confusion when people look into the source code of Magento modules.

See the `events.xml` file on the Reports module for an example:
https://github.com/magento/magento2/blob/2.1/app/code/Magento/Reports/etc/frontend/events.xml

tawfekov:
Hi @jcalcaben , 
I created this pull request because I saw the phrase on the same page : 
**name (required) - The unique observer name.**
do you think it can be updated to be  ?
**name (required)**

jcalcaben:
Yes, I will make some clarifications to the page. Thank you for pointing it out!

PR #756, opened 2016-07-16
Update ui-secondary-uploader.md

ClaudiuCreanga:
Having type array in fieldset and field throws the following error:
1 exception(s):Exception #0 (Magento\Framework\Exception\LocalizedException): Element 'fieldset', attribute '{http://www.w3.org/2001/XMLSchema-instance}type': The type definition 'array', specified by xsi:type, is blocked or not validly derived from the type definition of the element declaration.Line: 37


almarchenko:
Thanks!

PR #757, opened 2016-07-20
Update checkout_edit_form.md

ayodio:
seems like there was a typo in the folder name.


almarchenko:
You are right, thank you!

PR #758, opened 2016-07-20
Clarify some info regarding around plugins

AydinHassan:
Some info was lacking, regarding forwarding arguments so I clarified it with some more examples. Let me know if anything is missing 😄

### DIFF
--- a/guides/v2.0/extension-dev-guide/plugins.md
+++ b/guides/v2.0/extension-dev-guide/plugins.md
@@ -134,6 +134,57 @@ class ProductPlugin
 ?>
 {% endhighlight %}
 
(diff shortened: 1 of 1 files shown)


tanberry:
@AydinHassan  Thank You so much for this contribution!! @jcalcaben Can you take a look and then merge? Also, Aydin, is "variatics" something that the majority of developers will know about? If can you add a link or an explanation?
Thanks again, please keep the contributions coming!

jcalcaben:
Everything looks good. Thanks for the clarification @AydinHassan !

AydinHassan:
thanks guys, @tanberry regarding variadics, it's a core PHP construct, so I don't think there will be an issue there. If we get questions then I'm happy to update.

tanberry:
@AydinHassan , a new word for my internal glossary, thanks so much. BTW... if you are ever interested in helping out with a specific area in the docs, or even our nascent Glossary, please let me know. We have in the past recognized major contributors to our docs by displaying their names on the Jumbotron at events like Australia Live and UK Live, and we also add the name of the contributor to the top of any page for topics that are written completely by the contributor. Give it a thought and let us know! ;-)

PR #759, opened 2016-07-21
Update technology stack

avoelkl:


davidalger:
@xcomSteveJohnson Why would this information be included in the docs for 2.0 when there is a separate page for 2.1?

xcomSteveJohnson:
The pages are identical.

tanberry:
Hi @avoelkl and @davidalger ... yes, ironically we just discussed internally the need to update/improve this page earlier this week, and we have it on our sprint board to be updated... both the 2.0 page and the 2.1 pages, which should not of course be identical. Our plan is to remove any specific version numbers from this page, (in place we will add a link to a Sys Requirements page), and we will talk more generally about the various technologies and how they are implemented in/with Magento. How does that plan sound to you all? Feedback always welcome!

davidalger:
Sounds good to me! Just wanted to make sure it was something that was being considered. Thanks! 👍

tanberry:
The display of ESP by you and Anna is pretty impressive; we had it as an agenda item in our weekly DevDocs meeting just yesterday. I just wish we had gotten to it before. ;-

avoelkl:
Hi @tanberry and @davidalger,
Sounds good to me as well, given some of the information is kind of duplicate also.

PR #782, opened 2016-08-07
Use urn instead of path to file for mq documentation

timneutkens:
In the 2.1 and 2.2 docs this is already done.


keharper:
Thanks for pointing this out.

timneutkens:
@keharper, there are a couple more. I will have a look at them later 👍

PR #798, opened 2016-08-24
Use lowercase elasticsearch.

dmatthew:
elasticsearch needs to start with a lowercase 'e' or else the urls won't work.


xcomSteveJohnson:
Thanks!

PR #803, opened 2016-08-31
Correct error in maintenance mode command

crtsl:
Command had magento twice. Command will not work this way.


xcomSteveJohnson:
Thanks!

PR #804, opened 2016-08-31
Add Github link for "improved the deployment of static assets" to 2.1.1 release notes

erikhansen:
Adding the link will have two benefits:
- Provide more information to developers wanting to know what improvements were made. None of the developers at Classy Llama were aware what the ambiguous "We've improved the deployment of static assets." message meant and only discovered the changes once I stumbled upon the feature.
- Provide attribution to the contributor who created the PR. Thanks @denisristic !

### DIFF
--- a/guides/v2.1/release-notes/ReleaseNotes2.1.1CE.md
+++ b/guides/v2.1/release-notes/ReleaseNotes2.1.1CE.md
@@ -68,7 +68,7 @@ We address the following functional issues in this release.
 <!--- 57410-->* You can now quickly generate or preview multiple variations of a configurable product. (Saving these variations to the database can be time-consuming, if you have several thousand product options, and our efforts to improve performance continue.) Previously, Magento threw an Invalid Form key error is thrown while you tried to save a configurable product with variations. 
 
 
-<!--- 52660 -->* We've improved the deployment of static assets. 
+<!--- 52660 -->* We've improved the deployment of static assets. <a href="https://github.com/magento/magento2/pull/4294" target="_blank">(GITHUB-4294)</a>
 
 <!--- 55300, 55620, 54682-->* We've improved storefront performance when creating 2500 or more product variants. 
 


hostep:
Psssst: they also added parallellized deployment of static assets: https://github.com/magento/magento2/blob/37692c80f41e556e696803bbac8a93b8eb6837e7/app/code/Magento/Deploy/Console/Command/DeployStaticContentCommand.php#L534-L537
In a very quick test (which I'll have to repeat one day when I find more time), it doesn't seem to have a lot of effect on the speed of the deployment unfortunately.

erikhansen:
@hostep Oh wow. You're right. I missed that. I just confirmed that parallization is working and compilation is now blazing fast. 5 themes compiled in 31 seconds. This negates the value that I saw in being able to exclude locales/themes from the compilation process.

@xcomSteveJohnson I'm going to close this PR, as I think the most substantial improvement in 2.1.1 is the parallel compilation, NOT the new flags. The 2.1.1 release notes should mention the fact that static assets are now being compiled in parallel, or at least link to the source Github issue/PR. I don't see any definitive Github issue/PR where the parallel asset compilation is mentioned. All I see in commits like [this](https://github.com/magento/magento2/commit/520a804e27ba7e7708d0d9fb6d6080748917ef34) are references to MAGETWO-50673.

tanberry:
@erikhansen and @denisristic, thank you and thank you. We confess that we (DevDocs) didn't catch the importance of this item, and we agree that " "We've improved the deployment of static assets" is pretty darn ambiguous. 

This is a good kickstart to an initiative to improve our vetting and research on each of the issues in a release. A more carefully curated "Highlights of this Release" will be a goal for future Release Notes. Pointing out the fact that static assets are now compiled in parallel is something we should have been proud to state. (Yes, thanks @denisristic!)

We normally do provide, in the Rel Notes, a link to any GitHub Issue or PR that instigated the fix or better yet resolved the issue; we apparently failed to catch this one. We'll get that fixed in Rel Notes, and keep on keeping us straight; thank you.

erikhansen:
@tanberry I think a more detailed "Highlights of this Release" section in future release notes would be awesome. Those of us who manage many client sites and determine the importance and timing of upgrades will appreciate that.

Thanks for your continued commitment to improve Magento's documentation!

erikhansen:
Also, just to be clear: @denisristic was the one who added the flags for excluding/including compilation based theme/area/locale (see https://github.com/magento/magento2/pull/4294)

It appears that your internal development team was the one who added the parallel compilation.

Those are two separate things, both of which are worth mentioning in the release notes.

tanberry:
@erikhansen Ah, ok, @jfrontain is our heroic Release Notes writer, she was probably just about to straighten me out on that. ;-)

So Yes, for sure, we realize that Release Notes play a critical role in decisions about why and when to upgrade. We will add even more focus to making the content relevant and useful.

PR #807, opened 2016-09-07
Update magento-modes.md

keithbentrup:
I think you meant "dynamically generated" vs "cached" here b/c caching does not cause an adverse performance impact while dynamically generating the files based on an incoming request is.


xcomSteveJohnson:
Thanks!

PR #813, opened 2016-09-09
Update ECE first time setup for accuracy

davidalger:


xcomSteveJohnson:
Thanks. I accepted your edit but I reverted the names of the More information topics because they match the real topic names and I don't think adding the directory name to the title of the topic adds anything.

davidalger:
@xcomSteveJohnson Super, thanks!

PR #817, opened 2016-09-12
Removed unnecessary closing span tag + fixed /i18n code + removed som…

hostep:
…e obsolete whitespace.


hostep:
I also added a second commit which fixes the `_module.less` and `_widgets.less` filenames, they were missing an underscore.

tanberry:
@almarchenko  can you take a look and merge if approved. Thanks @hostep !!

almarchenko:
@hostep Great catch, thanks

PR #819, opened 2016-09-12
Removed an unnecessary code tag.

hostep:


hostep:
@almarchenko: is something wrong with this PR? Should I change something for it to be accepted?
Or is there another reason why it hasn't been merged yet?

I can see that the PR is a bit hard to understand due to all the trailing spaces which my editor removed, but you can append `?w=0` to the url to view the PR without whitespace changes.
I don't think the whitespace changes will cause issues, since only double whitespaces at the end of a line are imported in the Markdown format, right?

Thanks! :)

tanberry:
@hostep , sorry we have been keeping Sasha ( @almarchenko ) very busy with the new UI Components work! I took a quick look at this, and I see it is to remove the extra </code> tag from line 92. Great, thanks you for catching that! (Plus several white spaces at the end of sentences.)

hostep:
Tnx! (I didn't want to push you guys, it was more like a friendly reminder, I'll try to be a bit more patiently in the future 😉 )

tanberry:
;-) No, please don't be too patient. We prefer the pings; they help keep us on top of our GitHub stack! A mere 43 issues to go and 4 PRs.... ;-)

That was an interesting tip about appending ?w=0 to the url... thanks for sharing that!
 
Yes, two blank spaces/white spaces at end of line is like a <br>... and two tabs at beginning give you a bullet. I like markdown; keeps us focused on content, not too-fancy formatting.

PR #820, opened 2016-09-12
Added missing closing li tag.

hostep:
This should hopefully fix the `</ol>` tag which shows up:
screen shot 2016-09-12 at 23 27 26


almarchenko:
@hostep Thanks again!

PR #822, opened 2016-09-13
Changed the file path and made it clear that you need <vendor>_<modul…

edwinbos:
…e-name> instead of a directorty.

### DIFF
--- a/guides/v2.0/howdoi/checkout/checkout_new_step.md
+++ b/guides/v2.0/howdoi/checkout/checkout_new_step.md
@@ -40,7 +40,7 @@ A new checkout step must be implemented as UI component. That is, its JavaScript
 
 The file must be stored under the `<your_module_dir>/view` directory.
  
-A sample `view/my-step-view.js` with comments follows:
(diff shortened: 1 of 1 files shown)


tanberry:
Thanks, @edwinbos! @almarchenko  can you or @guz-anton take a look at this? We want to be careful to make it clear, because some environments might have modules in app/code while others are in /vendor.

edwinbos:
Hmm, that you want to be carefull to make it clear sounds a kind of strange to me, but perhaps we can call it <vendor-name>_<module-name>, so there is less confusion with the vendor folder.

almarchenko:
@edwinbos  You've pointed out really important issues! But after looking closer I realized it should be corrected in a little bit different way. Can I ask you to update the PR (I will give detailed explanations if you will be willing to do it yourself) or I can do it myself (but will close this PR then).
Please let me know what do you prefer!

almarchenko:
@edwinbos  I guess you have more fun things to do) And I was leaving for vacation, so I went ahead and made the changes myself. Thank you very much for pointing this out!

PR #827, opened 2016-09-18
Add documentation for the --theme and --exclude-theme options

mtinsley:
I added documentation for the --theme and --exclude-theme options for the setup:static-content:deploy command, which were added in 2.1.1.

The 2.0 file hasn't been forked yet, I made the following changes

- Update the 'Command Options' line to include the new options
- Add two new rows to the options table for the new options
- Adjusted the column widths of the table slightly
- Bumped the version number


xcomSteveJohnson:
Thanks! I added some more to it based on feedback from engineering.

philwinkle:
@xcomSteveJohnson these should have denoted new options added in new version. 2.0.7 user here, and for me devdocs is wrong

PR #828, opened 2016-09-19
Close XML Highlight before new paragraph

davehearne:
extra paragraph wrapped in the xml highlight option


tanberry:
Thanks, @davehearne !! Good catch. Keep 'em coming... ;-)

PR #834, opened 2016-09-25
Finally retire most wonderful de_DE typo.

benmarks:


almarchenko:
Wow, this was a good catch)

PR #837, opened 2016-09-26
Add missing step `cp package.json.sample package.json`

ajpevers:


almarchenko:
Hi @ajpevers ! Thank you for your input! After the first look at your PR I thought there's some magic going on around this doc. Cause I definitely remember myseldf adding info about Gruntfile and package.json in this topic%) 
By after a small investigation I realized what is going on. 
In Magento v2.0 Gruntfile.js and package.json are present in the root directory out of the box. https://github.com/magento/magento2/tree/2.0
So need to perform any additional steps with them.
Whilst in 2.1 they are Gruntfile.js.sample and package.json.sample https://github.com/magento/magento2/tree/2.1
So the v2.1 doc mentions that renaming is requried: http://devdocs.magento.com/guides/v2.1/frontend-dev-guide/css-topics/css_debug.html#grunt_prereq 
(step 3).

So if you do not mind I would close the PR with many "thank you":)

ajpevers:
Ah there is a v2.1 doc version! I wasn't aware of that. Thanks @almarchenko!

PR #840, opened 2016-09-27
add missing expanded option in menu widget

talalus:
add expanded option which display top level navigational items in mobile menu or all items.


tanberry:
Thanks @talalus , we appreciate that additional info. @almarchenko can you take a look and merge when verified?

almarchenko:
@talalus @tanberry Thank you, I will handle it.

almarchenko:
Thank you for your input!

talalus:
Sure :D

PR #842, opened 2016-09-30
Change the ID to be unique from other IDs

philwinkle:
The `#search-results` ID is already used elsewhere and the id had a style which applied a min-height of 400px causing the rest of the content to shove down below the page.


jcalcaben:
Thanks for the fix @philwinkle!

PR #845, opened 2016-10-01
Changed Solr Package Download Url

atishgoswami:
The old url `http://www.trieuvan.com/apache/lucene/solr/4.10.4/solr-4.10.4.tgz` doesnot work any more changed the url to `http://archive.apache.org/dist/lucene/solr/4.10.4/solr-4.10.4.tgz`


jcalcaben:
@atishgoswami Thanks for the update!

PR #846, opened 2016-10-01
Update theme-apply.md

COTIGA:
lines 34 to 35 do not match my M2 version (2.1.1)

### DIFF
--- a/guides/v2.0/frontend-dev-guide/themes/theme-apply.md
+++ b/guides/v2.0/frontend-dev-guide/themes/theme-apply.md
@@ -31,9 +31,9 @@ After you <a href="{{page.baseurl}}frontend-dev-guide/themes/theme-create.html">
 
 To apply a theme:
 
-2. In Admin, go to **Stores** > **Configuration** > **Design**.
-3. In the **Store View** drop-down field, select the store view where you want to apply the theme.
-4. On the **Design Theme** tab, select your newly created theme in the **Design Theme** drop-down.
+2. In Admin, go to **Content** > **Configuration**.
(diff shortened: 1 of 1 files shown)


tanberry:
@almarchenko can you please take a look? Thanks!

hostep:
Quickly adding a comment that this PR is incorrect.
The documentation for version 2.0 is correct and the documentation for version 2.1 is also correct as far as I can see.
@COTIGA, since you are using version 2.1.x, you should check out the documentation for version 2.1:  http://devdocs.magento.com/guides/v2.1/frontend-dev-guide/themes/theme-apply.html

The place where you apply a theme has been changed in the backend between Magento 2.0 and 2.1, that's why the docs for version 2.0 still mention the old place. So the docs are correct currently, there is no need to fix anything :)

tanberry:
Well, that is a great comment to read, thanks @hostep ! 
Sorry for the confusion between the doc versions, @COTIGA . Please let us know if you have feedback about the ease (or difficulty) or finding the "version toggle" on each DevDocs page. It is in the upper right on every page, and it allows you to toggle between the 2.0.x and the 2.1.x versions of the product. Also, please send feedback about how easy (or difficult) it is to realize what version of the product docs you are on, when scanning a topic.
We are currently doing some redesign of our pages, with help form @Magento UX, and would appreciate your feedback about how easy/hard it is to determine versions. Thanks in advance!

tanberry:
@COTIGA  we will go ahead and close this PR, but we look forward to hearing any feedback that you have on the versioning.

PR #850, opened 2016-10-05
Make the documentation less ambiguous

Vinai:
There is a lot of confusion in the Magento developer community if it is "best practice" or even okay to inject the ObjectManager as a constructor dependency.  
The current documentation says so implicitly already, but it would be helpful if it is stated it explicitly, too.


tanberry:
Hi @Vinai Yes we are tracking these conversations on SE very closely, and James is working on adding even more info to Devdocs, and handling the thorny questions of "When, if ever, is it ok to use the OM directly?" Also... a direct statement explaining that M2 is not fully refactored.

As always, once we get the docs published, we would greatly appreciate your and any community feedback. Thanks!

tanberry:
Tracking internal MAGETWO-59380 ticket.

jcalcaben:
Content merged. There is also ongoing work on revising the content for this topic.

PR #851, opened 2016-10-06
Fix broken link

bassplayer7:
This PR should fix a broken link.

---

I ended up [on this page](http://devdocs.magento.com/guides/v2.1/ui-components/ui_components_js.html) which had a lot of documentation that I hadn't found before. In looking through the site, I'm having trouble finding out where they are. Are they semi-not-public (yet) or am I missing some navigation?

### DIFF
--- a/guides/v2.0/ui-components/ui_components_js.md
+++ b/guides/v2.0/ui-components/ui_components_js.md
@@ -10,7 +10,7 @@ github_link: ui-components/ui_components_js.md
 ---
 
 ## What's in this topic
-This topic is aimed for developers, who need to reuse the [Magento UI Components]({{page.baseurl}}ui-library/ui-library-component.html). 
+This topic is aimed for developers, who need to reuse the [Magento UI Components]({{page.baseurl}}ui_comp_guide/bk-ui_comps.html). 
 
 The topic covers the following:
 


xcomSteveJohnson:
@bassplayer7 We're transitioning from the UI component documentation you cited for 2.0 to a new set of documentation in 2.1: http://devdocs.magento.com/guides/v2.1/ui_comp_guide/bk-ui_comps.html. 

I removed the `guides/v2.1/ui-components` directory and its contents because we'd prefer you use the documentation in `guides/v2.1/ui_comp_guide`. I'm tempted to go farther by redirecting topics like you've done but I'm not sure that's the right thing to do for now.

@tanberry and @almarchenko When should we start redirecting more old UI components topics to the new guide? Let me know because I have an internal PR for it.

Leaving open for the time being.

bassplayer7:
@xcomSteveJohnson, thanks for explaining. Feel free to close if it will be fixed some other way.

xcomSteveJohnson:
Assign to Sasha to complete the work.

tanberry:
Hi @bassplayer7 ; sorry for the confusion. So we are still working on the "new" 2.1 docs for UI Components. We want to keep the "old" docs available in the 2.0 version of devdocs.magento.com, but not do any redirecting or cross-linking between the two, because the new 2.1 docs will be completely new. 
So for now, you might need to move back and forth between the two, apologies. And I still need to handle the other great PR you submitted. ;-)

almarchenko:
@bassplayer7 @tanberry  Did I get it correctly that we are closing this PR for now?

tanberry:
@almarchenko Yes, please, because it works as intended now. We want to keep the old 2.0 book there, and I just verified that if you are in the 2.1 book, you can click the link in the intro paragraph to go to the old 2.0 book. @bassplayer7  thanks for all of your help!

bassplayer7:
Makes sense. 👍

PR #852, opened 2016-10-06
Fixed typo in PHP Version requirements

TiEul:
I think the install guide has a typo in which PHP 7.1.0 and PHP 7.0.10 were confused by the previous author. Officially, PHP 7.1.0 is not supported yet, at least not when doing a composer install without --ignore-platform-reqs.


hostep:
Or you could change it to `7.0.2 up to but not including 7.1.0` or something like that, otherwise you would have to change this line every time a new PHP version is released ;)

xcomSteveJohnson:
The statement is correct as is. I added "up to but not including" for added emphasis.

Closing.

PR #853, opened 2016-10-06
Missing initial code tag

Got-iT-Services-Inc:
Mission initial code tag on line 25 for "magento index:reindex"


dmrachkovskyi:
@Got-iT-Services-Inc Good catch, thanks so much for your input!

PR #854, opened 2016-10-06
Update system-requirements.md

ryantfowler:
The current installation rejects php7.1 (currently php7.1rc3), unless there's some special flag(s)/option(s) that need to be set that I'm not aware of.

This is what I get when installing with php7.1rc3:
```Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for magento/product-community-edition 2.1.1 -> satisfiable by magento/product-community-edition[2.1.1].
    - magento/product-community-edition 2.1.1 requires php ~5.6.0|7.0.2|~7.0.6 -> your PHP version (7.1.0RC3) does not satisfy that requirement.
```
I also tested it out with php7.0.11 and it worked just fine.

### DIFF
--- a/guides/v2.0/install-gde/system-requirements.md
+++ b/guides/v2.0/install-gde/system-requirements.md
@@ -37,7 +37,7 @@ MariaDB and Percona are compatible with Magento because we support MySQL 5.6 API
 
 *	5.6.x
 *	5.5.x, where x is 22 or greater
-*	7.0.2 up to 7.1.0, except for 7.0.5
+*	7.0.2 up to 7.0.11, except for 7.0.5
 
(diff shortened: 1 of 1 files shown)


ryantfowler:
Also, it looks like Php7.1 will be [deprecated](https://wiki.php.net/rfc/mcrypt-viking-funeral) mcrypt, which is currently a required php module for M2...so yea, until there's an update to the M2 codebase to support Php7.1, this should get removed from the docs :D.

xcomSteveJohnson:
7.0.10 is not the same as 7.1. The current requirements are correct. I'm not sure the status of mcrypt but will keep track of that.

Closing.

ryantfowler:
@xcomSteveJohnson I guess I read ```7.0.2 up to 7.1.0, except for 7.0.5``` as: "from 7.0.2 up to and including 7.1, except for 7.0.5" and that's what motivated my PR. My apologies for the misinterpretation. Thanks

xcomSteveJohnson:
Hush. You don't miss much, thanks for contacting us.

ryantfowler:
lol ;) np

PR #856, opened 2016-10-07
Update css_debug.md

iivashchenko:


hostep:
This `grunt-config.json.sample` file only exists in the `develop` branch of the Magento2 project. It isn't available in 2.0.x or 2.1.x, so I think this documentation is only meant for Magento 2.2.x, is that correct @iivashchenko?
If yes, then @almarchenko probably has to remove it from the 2.1 docs to not confuse the audience.

iivashchenko:
@hostep, thanks, you right, `grunt-config.json.sample` is only for 2.2.x. It's my fault, sorry for that.

hostep:
Ok thanks for the info. Should I create an issue for this? Or are you guys already working on fixing this?

almarchenko:
@hostep Thanks!! I'll fix this!
@iivashchenko I trusted you...;)))

almarchenko:
@hostep  @iivashchenko  I moved this info to the 2.2 doc (not published yet)
@hostep Thank you again!

PR #878, opened 2016-10-26
Fix minor formatting problems on DataSource tpic

bassplayer7:
Jekyll was not displaying `{{totalRecords}}` on the site and there was no line break before an unordered list. This should fix those problems.


tanberry:
@bassplayer7 thanks for catching that!

PR #885, opened 2016-11-02
Replace 'paymentData' parameter with 'method'

GaryStanton:
This parameter appears to be incorrect in the documentation. The correct name for the paymentData param is 'method' as per the WSDL.


keharper:
Thanks!

PR #887, opened 2016-11-03
Supported browsers list update

Igloczek:
Due to lack of support for IE9 by storefront themes, documentation should inform users about that.
Lack of IE9 support description -> https://github.com/magento/magento2/issues/5126

EOL of IE9 was at February 2016 (same as IE10), so investing time in supporting dead browsers are not so good idea for all of us (Magento team, our teams and clients).


xcomSteveJohnson:
Thanks, the Product team is reviewing this request.

PR #895, opened 2016-11-07
Fix grammar in CatalogStaging.md

brendanWeb:
Fixes a grammatical error that could confuse readers.

### DIFF
--- a/guides/v2.1/mrg/ee/CatalogStaging.md
+++ b/guides/v2.1/mrg/ee/CatalogStaging.md
@@ -111,7 +111,7 @@ The Magento_CatalogStaging module makes irreversible changes in a database durin
 
 ## Dependencies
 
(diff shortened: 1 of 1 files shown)


tanberry:
Hi @brendanWeb  thanks for your PR. The original sentence is referring to the the list of modules that have dependencies upon the Magento_CatalogStaging module... not the modules upon which the Magento_CatalogStaging module itself depends. 

The `require` section of the `composer.json` file lists those modules that depend on Magento_CatalogStaging, or "require" the Magento_CatalogStaging module.

Please let us know if this makes sense; it is quite possible that we still have it worded in a confusing way. Thanks again!

tanberry:
We are closing the PR, but please do let us know if the wording is still confusing. Thanks!

PR #897, opened 2016-11-08
Case sensitivity

GaryStanton:
I have discovered that the identifier type in this context is case sensitive and will only work with a lower case value of 'sku', rather than 'SKU'.
Other API methods that use an 'identifierType' param do not have this issue and will work with 'SKU' in upper case, so I feel this is a bug. I cannot however, find a Github repo for Magento and I'm not sure where I would post an issue.
Ideally, this should be fixed in the API but failing that I think it important to be noted in the documentation.


xcomSteveJohnson:
Created issue MPERF-9267 to track the underlying issue.

PR #898, opened 2016-11-12
Added definition for <block>, <arguments> and <argument>

riconeitzel:


almarchenko:
@riconeitzel Thank you for your input!
Looks like this table only contains the instructions specific for page configuration files. All other instructions are described in the [Layout instructions](http://devdocs.magento.com/guides/v2.1/frontend-dev-guide/layouts/xml-instructions.html) topic.
(see the text right before the table). 
<block> and <argument> are described there, though <arguments> is not.
If you are willing, you can change the PR. But also we can close it, and I will add <arguments> in xml-instructions.md myself.
Please let me know what do you prefer! Thank you.

riconeitzel:
@almarchenko hi oleksandra, I'm currently slightly out of time. If you get it done in the next 2 weeks, feel free to close the PR and migrate it yourself. If not: Let it open and I'll check back in a week or two. Rico :-)

almarchenko:
@riconeitzel  
Thank you! 
Actually, i've re-read my comment and found that I didn't use the correct formatting, so the names of the instructions were lost%)
It was " and are described there, though is not."
instead of
`<block>` and `<argument>` are described there, though `<arguments>` is not.
I wonder what did you think about this strange wording:))
Anyway, thank you for your reply, I will close this once I add it quickly myself.

PR #899, opened 2016-11-12
Change escapeXssInUrl to escapeUrl

avoelkl:
escapeXssInUrl is not called from any blocks in the Magento 2 Core, esacpeUrl is. escapeUrl calls escapeXssInUrl internally, see https://github.com/magento/magento2/blob/1ade3b769a937f19682bbbe4e51b6c956147952f/lib/internal/Magento/Framework/Escaper.php#L206


tanberry:
Thanks und Danke, Anna! And thanks for including link to code.

PR #910, opened 2016-11-30
Add Centos 7 config file location

mtinsley:
In Centos 7, the varnish configuration file is in /etc/varnish/varnish.params rather than /etc/sysconfig/varnish


xcomSteveJohnson:
Thanks!

PR #911, opened 2016-12-01
corrected link for Enterprise Cloud Edition

JonyD:
3 dashes instead of 2 (according to previous instruction)


xcomSteveJohnson:
Thanks!

PR #912, opened 2016-12-01
Corrected step 6; removed image.

JonyD:
There is no such "Add Store Code to Urls" option in Store View. This option is in "Default Config".
Also, the image is misleading and should be removed. One can get an idea where to click (in **Store View**) by following the image displayed after step 5.


xcomSteveJohnson:
Thanks! I disagree with you about the screen capture so I added one that shows **Default Config**. The problem with removing it is that in the previous step, the scope was different.

PR #913, opened 2016-12-01
Removed wrong instruction; corrected link

JonyD:
The deleted instruction was wrong because there is no such checkbox in the **Default View**; corrected link same as in commit d5d08f4.

### DIFF
--- a/guides/v2.0/config-guide/multi-site/ms_websites.md
+++ b/guides/v2.0/config-guide/multi-site/ms_websites.md
@@ -156,17 +156,15 @@ You must perform this step last because after it's done, you'll lose access to t
 1.	In the Admin, click **Stores** > Settings > **Configuration** > General > **Web**.
 2.	From the **Store View** list at the top of the page, click **Default Config** as the following figure shows.
 
-
 	![Select the default config scope]({{ site.baseurl }}common/images/config_multi-site-default.png){:width="250px"} 
 
 3.	In the right pane, expand **Base URLs**.
-4.	In the Base URL section, clear the **Use Default** check box.
 4.	Enter `http://mysite.mg` as the following example shows.
 
 	![Set a base URL]({{ site.baseurl }}common/images/config_multi-site_base-url-default2.png){:width="650px"}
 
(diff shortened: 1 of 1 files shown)


xcomSteveJohnson:
Unable to automatically merge this one, closing.

JonyD:
@xcomSteveJohnson "Unable to automatically merge this one, closing."? Can you please at least look at the changes and edit the "unmergable" commit? You are throwing away a correction to something that is published/wrong.

xcomSteveJohnson:
I took a look, I don't see anything. What am I missing?

JonyD:
look here: https://github.com/magento/devdocs/pull/913/files.
the instruction 4. was removed because, like my first comment says, "The deleted instruction was wrong because there is no such checkbox in the Default View". Other change is in the link: has 3 dashes instead of 2..

xcomSteveJohnson:
Look [here](http://devdocs.magento.com/guides/v2.1/config-guide/multi-site/ms_websites.html#multi-storecode-baseurl) and [here](http://devdocs.magento.com/guides/v2.1/config-guide/multi-site/ms_websites.html#step-7-change-the-default-store-views-base-url)

JonyD:
I looked and I stand my point: I did everything right. Please read carefully:
- in point 7.  the link is wrong. It needs 3 dashes instead of 2 (according to the text: "you must replace the first period with three dashes" ) so, instead of "http://french--branch-sbg" it should be "http://french---branch-sbgW"
- point 7.4 needs to be removed because that procedure to "change the default store view's base URL", does not exist. If you follow instructions steps you will see that there is no default checkbox (you can even see that in the image after 5.). So, the instruction that says "In the Base URL section, clear the Use Default check box." needs to be removed from this step-by-step guide.
I cannot explain any better.

JonyD:
I made a new commit so you can see it better. commit: https://github.com/magento/devdocs/commit/6becad16f761ba0427c5700f6988ef4659389141

xcomSteveJohnson:
@JonyD Was a busy week, finally got a chance to look. You're correct. Fixed now.

PR #925, opened 2016-12-13
Customize email templates corrections

zanilee:


almarchenko:
@zanilee thank for making docs better!

PR #930, opened 2016-12-14
Add /etc for better understanding

denchev:
config.xml must reside in the /etc folder inside %Vendor_Module%. It was a bit misleading otherwise.


almarchenko:
@denchev sure, you are right! Thank you for your input!

PR #933, opened 2016-12-15
Update css-practice.md

ryantfowler:
There is a lack of detail regarding the need to clear the static files cache....I spent hours following this specific section of the documentation (while in Developer mode) with *all* cache disabled, with zero effect...until I clicked on "Flush Static Files Cache". There should be clearer notification regarding this additional cache needing to be cleared at the beginning of all sections that related to css customization.

### DIFF
--- a/guides/v2.0/frontend-dev-guide/css-topics/css-practice.md
+++ b/guides/v2.0/frontend-dev-guide/css-topics/css-practice.md
@@ -75,8 +75,6 @@ To do this, they added an overriding <code>_theme.less</code> file in the <code>
 
 </pre>
 
-Next, you need to flush the static files cache. This can be done in a few ways, please refer [here](http://devdocs.magento.com/guides/v2.1/config-guide/cli/config-cli-subcommands-static-view.html).
-
-After the Grey theme is applied, store pages will look like following:
+After the Grey theme is applied (and [static files cache cleared]({{page.baseurl}}howdoi/clean_static_cache.html)), store pages will look like following:
 
 <img src="{{ site.baseurl }}common/images/css_practice.png" alt="product page when Grey applied">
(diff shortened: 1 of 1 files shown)


xcomSteveJohnson:
@almarchenko Suggest you also add a link to the User Guide because you can do this from the Admin as well.

ryantfowler:
@xcomSteveJohnson The link I added to the PR has some decent notes on static files cache and modes. There's a section for `Developer mode only` which provides a few ways to clear the static files cache. In that section it provides details for also clearing the static files cache vie the Admin panel. The page I linked isn't specifically "how to clear static files cache" but it does provide more details. 

The closest user guide link that I could find is this: http://docs.magento.com/m2/ce/user_guide/system/cache-management.html?Highlight=clear%20static%20files%20cache and it's not really clear on the static files cache clearing...it has the button listed, but that's it. The link I have in this PR already is closer to a clearer explanation, in my opinion. Otherwise, there could be a separate "troubleshooting css styles not being applied" page which has details specifically for clearing cache and could include information from the page that I linked.

Ideas?

xcomSteveJohnson:
@ryantfowler @almarchenko Yes, I agree but some merchants have no command-line access. Given that the User Guide doesn't make it clear, I would suggest to put the steps in this topic as well.

There could also be troubleshooting but I'm not familiar with the Frontend Dev Guide so I'll let Sasha decide.

almarchenko:
@ryantfowler  Thank you for your input! 
I will double check with developers what is the correct behavior, and merge your PR if this necessity to clean cache is not a bug.
the internal issue 63021

almarchenko:
@ryantfowler 
Here's what I have having discussed this with our devs:
In general case, in Developer mode, there is no necessity to clean cache. But it also depends on your previous actions and settings, so might become necessary at the end (like it was in your case).
And I see that we actually mention this in the procedural topics:
http://devdocs.magento.com/guides/v2.1/frontend-dev-guide/css-guide/css_quick_guide_mode.html
(we suggest manually cleaning the directories, but it is only one of the possible options).
I think we didn't add it in the topic you mentioned, cause it was a high-level illustration. But looks like it still should be there!

ryantfowler:
Thanks for the feedback!

PR #938, opened 2016-12-17
Updates the Magento 2.1.3 release notes with a link to issue 5101, which was missing.

hostep:
I'm not sure if that number in front is your internal MAGETWO number, but that might be incorrect.
It should probably be 57026 or 54786, mentioned in this PR: https://github.com/magento/magento2/pull/5116#issuecomment-252547488
If you want me to fix that number too, let me know, or you can change it yourself, your choice :)


jfrontain:
Hi Pieter, thanks for adding a link to  GITHUB issue 5116. I've checked that the internal MAGETWO issue number is correct. Thanks for your close reading of our notes, and please keep those corrections and enhancements coming!

PR #952, opened 2016-12-21
add Use ObserverInterface in class

kandarp26:
Added ObserverInterface in class which would required for implement the interface


jcalcaben:
Thanks for the update!

PR #953, opened 2016-12-21
Clarify strucure of searchCriteria GET parameters

Vinai:


keharper:
Good catch. And I had so many examples showing that wasn't the correct syntax.

PR #954, opened 2016-12-21
New Topic on Template Literals in UI Components

bassplayer7:
This has been long in the making. The main challenge I had was keeping this topic focused on template literals and not take rabbit trails that would probably better be in other topics. This is because there are many concepts that I touch on that are not (yet) covered in other topics. As new topics are created, it would probably be good to add links to them: like [here](https://github.com/magento/devdocs/compare/develop...bassplayer7:develop#diff-f4a04aa7a12f831e492bc7de475e064dR26) [here](https://github.com/magento/devdocs/compare/develop...bassplayer7:develop#diff-f4a04aa7a12f831e492bc7de475e064dR37), and [here](https://github.com/magento/devdocs/compare/develop...bassplayer7:develop#diff-f4a04aa7a12f831e492bc7de475e064dR56).

Due to size of the scope and subsequent complexity of this concept in Magento, the topic is a bit long. Perhaps it would be better to split it up a little? Although, the concepts all seem integral to the use of template literals. 

I am very open to whatever suggestions you have for improvement! I tried to avoid a tutorial feel and match other topic's style as much as possible.

### DIFF
--- a/guides/v2.1/ui_comp_guide/concepts/ui_comp_template_literals.md
+++ b/guides/v2.1/ui_comp_guide/concepts/ui_comp_template_literals.md
@@ -1,0 +1,96 @@
+---
+layout: default
+group: UI_Components_guide
+subgroup: concepts
+title: Template Literals in UI Components
+menu_title: Template Literals
+menu_order: 10
+contributor_name: SwiftOtter Studios
+contributor_link: https://swiftotter.com/
+version: 2.1
+github_link: ui_comp_guide/concepts/ui_comp_template_literals.md
+---
+
+# Template Literals in Magento
+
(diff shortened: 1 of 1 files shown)


tanberry:
@bassplayer7 , thank you so much! I have seen the recent research and work that you and several others have been doing in this area, especially the ES2015 / ES6 specification background. We really appreciate you writing up the results and sharing it with the whole community via DevDocs!

This topics looks great, is clearly a conceptual topic (believe me, we would be happy with a tutorial as well!). I don't think it is too long, not at all, and it is very focused on template literals. Agreed it is hard to not go off down rabbit holes; there are plenty of them. And yes, as we add new topics, we should add cross-reference links between the concepts, for sure.

Let us run it through our normal process of internal review, and then we look forward to merging it and getting it published! Again, thanks so much!

bassplayer7:
Wow! That was fast.

tanberry:
Thanks @bassplayer7  for the contribution, and @olhanakonechna for the speedy review. Olha is one of the core frontend devs, so if your article passes her review, all is good. ;-)

We are still working on some plans for a second Doc Sprint, specifically about UI Comps, with our core devs and community devs. Jesse, I hope you will consider joining us when we get the location/event finalized.

Thanks again for the article!

bassplayer7:
That's great. I saw a couple new topics in there which is nice - but a lot more to go. 

Let me know on the doc sprint. I'd be interested in helping.

PR #962, opened 2016-12-30
Fix typo

denchev:
Fixing a misspelled word "Model".


jcalcaben:
Nice catch on the typo!

PR #965, opened 2017-01-02
Update ui-form.md

chamalC:
corrected the path to definition.xml


almarchenko:
@chamalC Thank you for your input!

PR #967, opened 2017-01-03
Debug UI components JavaScript - Fix links URLs

ihor-sviziev:


almarchenko:
@ihor-sviziev  Thank you for your input!

PR #973, opened 2017-01-05
Remove redundant phrase

scottsb:
Not sure what this was supposed to be saying but option A and option B are identical, so this cleans it up.


xcomSteveJohnson:
D'OPE! Thanks!

PR #977, opened 2017-01-06
Fix incorrect indentation

scottsb:
These items are not subitems--they should be parallel with the overall list.


xcomSteveJohnson:
Thanks!

PR #984, opened 2017-01-11
Update checkout_order.md

11mb:
Fix the directory of the custom validator


almarchenko:
@11mb Thank you for your input!

PR #985, opened 2017-01-12
Add information about required rewrite rule for Nginx

cmtickle:
A required rewrite rule has been added to pub/static/.htaccess this causes 404 responses for users of Nginx. This update details the workaround

### DIFF
--- a/guides/v2.1/release-notes/ReleaseNotes2.1.3CE.md
+++ b/guides/v2.1/release-notes/ReleaseNotes2.1.3CE.md
@@ -559,8 +559,21 @@ This release introduces the `Magento\Vault\Block\TokenRendererInterface::getToke
 
 	**Workaround**: You can reduce processing time and performance by increasing your [PHP `memory_limit` setting]({{ page.baseurl }}install-gde/prereq/php-settings.html) to 1 GB.
 
+* **Issue** If using Nginx to serve static content you will see 404 responses for CSS/JS. This is due to requirements for a  rewrite which has been added in pub/static/.htaccess
+```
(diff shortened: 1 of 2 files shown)


xcomSteveJohnson:
Thanks, we're confirming this with engineering.

cmtickle:
This requirement is covered in the sample Nginx configuration of the 2.1 branch introduced in this commit: https://github.com/magento/magento2/commit/b6f19a8d02fcae7d4e21e353e423d1c240166d1a
However, relying on this to communicate the change would assume everyone updates their Nginx configuration based on the sample with each new release.

tanberry:
Hi @cmtickle . Apologies that it has taken so long to reply. After looking at this PR, we would like to suggest that rather than adding the info to the Release Notes for a specific version, we instead add a new topic in devdocs.magento.com, in the Configuration Guide, with the guidance that one should always add version-specific rewrites to the nginx.conf file, and the basic instructions for doing so. And add a link to the nginx.conf.sample file in our repo.

Please let us know your thoughts on that proposal, and, if you are willing to write this topic that would be fantastic. We acknowledge our devdocs contributors by adding their name and a link-back at the top of the topic. We even have templates, to make it very easy. ;-) Please take a look at our [Contributor's Guide](http://devdocs.magento.com/guides/v2.1/contributor-guide/contributing_docs.html).

tanberry:
@mazhalai Can you please take a quick look and let us know if you agree with this proposal; we want to add a new topic to the Config Guide, advising that one should always add version-specific rewrite configs to the nginx.conf file.

PR #987, opened 2017-01-12
Fixed incorrect callable type hints.

hostep:
The `callable` type hint shouldn't have a `\` prefix, otherwise you get errors like:
```
PHP Parse error:  syntax error, unexpected 'callable' (T_CALLABLE), expecting identifier (T_STRING)
```


jcalcaben:
Thanks for the correction!

PR #990, opened 2017-01-13
Added info about payment token factories

joni-jones:
Added more details about payment token factories and they usage.


almarchenko:
@joni-jones  Thank you for your input. 
I've edited the text a little bit. If you confirm that my edits are ok, I will merge the PR

PR #999, opened 2017-01-20
Added example how to add CLI to $PATH

joni-jones:


xcomSteveJohnson:
@joni-jones I edited your text a bit. Thanks!

PR #1001, opened 2017-01-20
Correct the filename in the flow

zanilee:
There is a wrong filename in the flow. Seems like a typo.


dshevtsov:
@zanilee you are right, thanks!

PR #1005, opened 2017-01-20
simplify permission fixing commands

scottsb:
Two changes:
1. Consolidate `find` commands to handle directories and files together when there is no difference in `-exec`.
2. Remove extra direction about how to run the two commands together. This just clutters the doc and makes it harder to read for a fraction of a second of effort saved.


xcomSteveJohnson:
Thanks! @scottsb I collapsed all the commands into single commands. Now that we have a Copy button, it's easier. Verified they all work.

Issue #1013, opened 2017-01-24
Missing Page: js_init.html


dmatthew:
The following page is missing: http://devdocs.magento.com/guides/v2.1/frontend-dev-guide/javascript/js_init.html

It is linked to at the bottom of this section: http://devdocs.magento.com/guides/v2.1/javascript-dev-guide/javascript/custom_js.html#extend_js_component

dmatthew:
It looks like this may be the page: http://devdocs.magento.com/guides/v2.1/javascript-dev-guide/javascript/js_init.html

PR #1014, opened 2017-01-24
Update link to js_init.js file

dmatthew:
The correct path for frontend-dev-guide/javascript/js_init.html should be javascript-dev-guide/javascript/js_init.html.

Fixes https://github.com/magento/devdocs/issues/1013


almarchenko:
@dmatthew  Great catch, thank you!

PR #1017, opened 2017-01-26
Added example how to avoid of code duplication

joni-jones:
Added example of plugin, which allows avoiding code duplication if entity doesn't have default implementation of fetching extension attributes


jcalcaben:
Thanks for the contribution!

PR #1018, opened 2017-01-27
Added list of common payment operations

joni-jones:


almarchenko:
@joni-jones  Thanks for your input!

PR #1020, opened 2017-01-30
Add a note about contacting support if download fails

gdoug-guidance:
If you magento account is not authorized to download the EE version, you will receive the  Could not find package magento/project-enterprise-edition with version 2.0.0 error message. You have to contact EE support, and have them open it.


xcomSteveJohnson:
Thanks!

PR #1021, opened 2017-01-31
Update form ui component path with the correct one

asubit:
Tested on Magento CLI version 2.1.3


almarchenko:
@asubit You are right! Thank you for your input!

asubit:
@almarchenko You're welcome. It's always a pleasure to help and contribute!

PR #1022, opened 2017-01-31
fix modal link

slackerzz:


almarchenko:
@slackerzz  Great catch!! Thank you!

PR #1023, opened 2017-01-31
Moved inheritdoc to Other Tags

buskamuza:
As `@inheritdoc` is applicable not only to methods, moved it to Other Tags


xcomSteveJohnson:
Thanks!

PR #1026, opened 2017-02-02
Update theme-apply.md

atai:
Error on page. Wrong path in admin panel.


almarchenko:
@atai 
Thank you for being that attentive to our docs! 
Your correction is absolutely right, but only for Magento v2.1 (so we have it in v2.1 docs) http://devdocs.magento.com/guides/v2.1/frontend-dev-guide/themes/theme-apply.html

And in Magento 2.0 it is the way we describe in docs. 

I will close this PR, but please, keep it coming!

PR #1028, opened 2017-02-03
Specify admintml area for di.xml when applying theme

jameshalsall:


almarchenko:
Hi @jameshalsall!

Thank you for your input!
Actually, both ways must work (the way it is in docs, and the way you suggest). And I think we should add the option you suggest. 
But first I wanted to double-check:
Did you have problems applying a theme the way it is described in the doc? If yes, did you specify the dependency as advised in the doc: 
http://devdocs.magento.com/guides/v2.1/frontend-dev-guide/themes/admin_theme_apply.html#prerequisites

jameshalsall:
Hey @almarchenko - yes I followed the docs exactly. As soon as I moved it to `adminhtml` area it worked as expected.

almarchenko:
@jameshalsall  Thanks! interesting feedback, i'll discuss this with our devs!

almarchenko:
@jameshalsall  Thank you for your input!
Developers confirmed that using the admin-area di.xml is the best way.

PR #1045, opened 2017-02-16
Fix incorrect sign of migration timestamp correction

scottsb:
If the time zone for the database field is `UTC-7`, then it needs to be corrected by *positive* 7 hours to match UTC.


jcalcaben:
Hi @scottsb 

Thanks for pointing this out! We have incorporated this fix when the file was moved/renamed. Unfortunately, we cannot merge this PR because of the merge conflict.

PR #1046, opened 2017-02-17
Fix Gateway Command url on payment action page

mikeoloughlin:
Just fixing some broken markdown for a URL in the first section of this page:
http://devdocs.magento.com/guides/v2.1/payments-integrations/base-integration/payment-action.html


xcomSteveJohnson:
Thanks!

PR #1052, opened 2017-02-22
Proposal: explicitly mention PSR standards

buskamuza:
Many developers know about PSR standards, so I don't see a reason to confuse them with hiding the fact that the references we mention are PSR standards.


xcomSteveJohnson:
Thanks!

PR #1053, opened 2017-02-23
Update php-ubuntu.md

awei01:
Fixing order of installation instructions for php 5.6. need to `update` before `install`


xcomSteveJohnson:
Thanks!

PR #1055, opened 2017-02-24
Change Filetype of DB Backup from tgz to .sql

osrecio:
in File :  [BackupRollback.php](https://github.com/magento/magento2/blob/develop/lib/internal/Magento/Framework/Setup/BackupRollback.php#L207)  you can see:`$dbBackup->setBackupExtension('sql');`


xcomSteveJohnson:
Thanks!

PR #1060, opened 2017-03-02
Change template link to one that is updated

jamescowie:
After working through the deploy process guide. There is a section for a template to use [here](http://devdocs.magento.com/guides/v2.0/cloud/discover-deploy.html). However, this template is a fork of the main magento one that gets updated more regular. The linked GitHub repo is very out of data in terms of Magento version etc. I propose we with merge this PR with link to the upstream main template or start to update the forked template.


xcomSteveJohnson:
Thanks!

PR #1077, opened 2017-03-25
How to create and display a category attribute

bassplayer7:
This is a short and reasonably simple tutorial on how to create and display a category attribute - something that is easy to do but does not appear to have much information related to it. Perhaps more important, it gives a bit of an overview on how the configuration is passed into the Javascript classes.

A few notes:

- There doesn't appear to be any official guidance on install scripts. While the sample below works well, if there are better semantics, it can be adjusted as needed.
- I don't currently have placeholder notation. It appears that DevDocs uses `%` to denote the beginning and end of strings that should be replaced by the user. However, I didn't want any confusion on whether those were necessary. That said, I think it could be good to included them. Thoughts?

### DIFF
--- a/guides/v2.1/ui_comp_guide/howto/add_category_attribute.md
+++ b/guides/v2.1/ui_comp_guide/howto/add_category_attribute.md
@@ -1,0 +1,102 @@
+---
+layout: default
+group: UI_Components_guide
+subgroup: howto
+title: How to add a Category Attribute
+menu_title: Creating and displaying a category attribute with UI Components
+menu_order: 3
+contributor_name: SwiftOtter Studios
+contributor_link: https://swiftotter.com/
+version: 2.1
+github_link: ui_comp_guide/howto/add_category_attribute.md
+---
+
+# How to add a Category Attribute
+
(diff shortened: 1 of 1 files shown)


keharper:
First of all, thank you for your contribution. This is great stuff.

Can the category attribute that you created in this example be used in REST calls? If so, how?

bassplayer7:
@keharper good question. To be honest, I didn't check that. If it was, I wonder if that would perhaps be a bit beyond the targeted scope of this topic? I was hoping that there was a topic on setup scripts which could be simply linked to, but there doesn't appear to be one at the moment. Either way, I can take a look if necessary.

keharper:
REST information would definitely be outside the scope of this topic. If it's realistic that a developer would add an attribute to the front end and carry that attribute forward as part of an order or send it to a 3rd-party app, then having a multi-part how-to would be pure gold.

keharper:
We appreciate your contribution, @bassplayer7 . This is a great article. I'll be making a few tweaks, mostly to ensure the links resolve for each release.

PR #982, opened 2017-01-10
Add Layout XML Standard

bassplayer7:
The goal of this is to formalize a standard for writing layout XML. There hasn't been a standard set that the community follows when authoring layout XML. By setting a precedent for layout XML naming conventions, I am hopeful that it will lead to XML declarations that are easier to read and more consistent which could decrease maintenance time.

PR #1078, opened 2017-03-25
Add Layout XML Standard (new branch)

bassplayer7:
The goal of this is to formalize a standard for writing layout XML. There hasn't been a standard set that the community follows when authoring layout XML. By setting a precedent for layout XML naming conventions, I am hopeful that it will lead to XML declarations that are easier to read and more consistent which could decrease maintenance time.

---

Note: this is **identical** to #982. I found that I did that PR on the wrong branch of my fork and needed to move over to this one to provide flexibility.

### DIFF
--- a/guides/v2.1/coding-standards/xml-coding-standard.md
+++ b/guides/v2.1/coding-standards/xml-coding-standard.md
@@ -1,0 +1,59 @@
+---
+layout: default
+group: coding-standards
+subgroup: Coding standards
+title: XML Coding Standards
+menu_title: XML Coding Standards
+menu_order: 100
+contributor_name: SwiftOtter Studios
+contributor_link: https://swiftotter.com/
+version: 2.1
+github_link: coding-standards/xml-coding-standard.md
+---
+
+# XML Coding Standard
+
+### Name attribute
+
+The `name` attribute should adhere to the guidelines listed below. The `name` attribute should:
+
+- Be unique to the project.
+- Use a namespace approach where each child block's name contains its parent's name as well.
+- Have segments separated by a `.` (period).
+- Be all lowercase.
+- Never use `_` (underscores).
+
+*Example:*
+
+```xml
+<block name="header">
+  <block name="header.right">
+    <block name="header.right.search">
+      <block name="header.right.search.autocomplete" />
+    </block>
+  </block>
+</block>
+```
+
+### As attribute
+
+The `as` attribute should follow the guidelines below. The value of `as`:
+
+- Only needs to be unique to the block that contains it.
+- Should be as concise as possible, but with enough clarity to be understood within the block.
+- Can use an `_` (underscore) as a separator.
+- Should only be added if necessary.
+- Should be all lowercase.
+
+
+*Example:*
+
+```xml
+<block name="header">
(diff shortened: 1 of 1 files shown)


almarchenko:
Hi @bassplayer7!

Thank you for your input! 
We have the "name" and "as"  attributes described in the layout-related chapter of Frontend Developer Guide:
http://devdocs.magento.com/guides/v2.1/frontend-dev-guide/layouts/xml-instructions.html#fedg_layout_xml-instruc_ex_block
But in fact, of course, they are useful not only for frontend developers. And also your descriptions contain more info. 
I've created an internal ticket (67490) to consult with our internal developers.

almarchenko:
@bassplayer7 What I found out is that currently we don't have these internal standards. 
We need the opinion and decision of architects on this. 
@antonkril  perhaps you can help? Should Magento start following these standards?

antonkril:
@bassplayer7, this is great. We'll discuss it internally.

bassplayer7:
@almarchenko thanks for checking into this. 

@antonkril thanks for picking this up as well. The exact details don't much matter to me, but over time, I've felt that having at least a naming recommendation could help us identify blocks more instinctively.

guides/v2.1/coding-standards/xml-coding-standard.md
```diff
@@ -0,0 +1,59 @@
...
+    </block>
+  </block>
+</block>
+```
+
+### As attribute
+
+The `as` attribute should follow the guidelines below. The value of `as`:
+
+- Only needs to be unique to the block that contains it.
```
benmarks: I have an internal goal of having this complete tonight (12-Jun), but it will be a couple more days. Thanks again for proposing.

guides/v2.1/coding-standards/xml-coding-standard.md
```diff
@@ -0,0 +1,59 @@
...
+---
+
+# Layout XML Coding Standard
+
+### Name attribute
+
+The `name` attribute should adhere to the guidelines listed below. The `name` attribute should:
+
+- Be unique to the project.
+- Use a namespace approach where each child block's name contains its parent's name as well.
```
benmarks: Adding for the record (this is the salient point): 

> Use a namespace approach where each child block's name contains its parent's name

Because block names exist in the global layout scope, this isn't a "namespace approach"; rather, it's a hierarchy approach.

jcalcaben:
@benmarks @antonkril 
what is the status of this PR? is it still in development or ready to merge?

guides/v2.1/coding-standards/xml-coding-standard.md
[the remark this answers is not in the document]
buskamuza: What do we expect in this scenario? I'd leave the statement as is.

buskamuza:
Created another PR as I can't edit this one - https://github.com/magento/devdocs/pull/2051

benmarks:
I haven't gotten to this in a year, so consider me removed.

buskamuza:
Closing the proposal based on the discussion https://github.com/magento/architecture/wiki/October-31,-2018

PR #1080, opened 2017-03-28
Added a related topic

ekuusela:
The "Adding extension attributes to entity" is a relevant guide since it describes one way to persist extension attributes. It also serves as a hint that values set to the extension attributes actually have to be separately persisted.


xcomSteveJohnson:
Thanks!

PR #1105, opened 2017-04-17
Typo config:set -> setup:config:set

LukeHandle:
None of the other examples give the `php` prefix either: removed. Slightly adjusted the description.

Error given on 2.1:
```
$ magento config:set

  [InvalidArgumentException]
  There are no commands defined in the "config" namespace.
  Did you mean one of these?
      setup:config
      setup:store-config
```


jcalcaben:
Thanks for the typo fix @LukeHandle !

PR #1111, opened 2017-04-23
Fix indentation on 2.0 redis-session conf example

LukeHandle:
The current config example has 2 issues:
1. It uses tabs part way through, this does not format well when copying/pasting.
2. It adds indent to the `array (` declaration, whereas the default `env.php` does not, nor does the `redis-pg-cache` page.


jcalcaben:
Thanks for the format fix @LukeHandle !

PR #1112, opened 2017-04-23
Fix extra <tr> tags

lfolco:


tanberry:
Thanks @lfolco !! We appreciate the help with keeping the errant <tr>s down.

PR #1121, opened 2017-04-26
Updated Consumer Runner class

rev42:
The class has been changed on Magento 2.1

### DIFF
--- a/guides/v2.0/config-guide/mq/manage-mysql.md
+++ b/guides/v2.0/config-guide/mq/manage-mysql.md
@@ -22,7 +22,7 @@ The following shows a `crontab` group entry:
 
 {% highlight xml %}
 <group id="default">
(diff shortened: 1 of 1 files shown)


keharper:
Thanks!

hostep:
@rev42: if this class was changed in version 2.1, why was the 2.0 version of the documentation updated? You should update version 2.1 of the docs, if a version 2.1 of this particular page doesn't exist yet, copy 2.0 to 2.1 and modify the new file.

(Sorry if what I'm saying is incorrect, I have no way of verifying this, since I don't have access to the EE sourcecode.)

keharper:
@hostep  I didn't notice that. I'll break the symlink between the 2 versions and make sure that only the 2.1 version contains this info.

rev42:
@hostep Good point. My intent was indeed to update 2.1 docs but I didn't notice I was on 2.0 docs.
Nevertheless I just checked Magento 2.0.x EE source code and the class ConsumerRunner was already into Magento\MessageQueue\Model namespace.
So I guess we're safe.

/cc @keharper

PR #1122, opened 2017-04-27
Update css_quick_guide_approach.md

ktan88:
The section is about overriding button styles, instead the text says, it's a way to **extend** the parent theme's styles for buttons


ccondra:
Thank you @ktan88!  We really appreciate your help. Keep them coming!

ktan88:
Ur welcome @ccondra =)
If there is a little more time, maybe I can add some more How-tos about Frontend Dev, f.e. Theme Creation

PR #1134, opened 2017-05-04
Added provider to layout

wildcard27:
Without setting a provider for the step, you will not have access to certain properties, such as 'source` which is required for validation. Copying the code above and running it as-is will result in console errors.

### DIFF
--- a/guides/v2.0/howdoi/checkout/checkout_new_step.md
+++ b/guides/v2.0/howdoi/checkout/checkout_new_step.md
@@ -174,6 +174,7 @@ A sample `checkout_index_index.xml` follows:
                                             <!-- The new step you add -->
                                             <item name="my-new-step" xsi:type="array">
                                                 <item name="component" xsi:type="string">%Vendor%_%Module%/js/view/my-step-view</item>
+						<item name="provider" xsi:type="string">checkoutProvider</item>
                                                     <!--To display step content before shipping step "sortOrder" value should be < 1-->
                                                     <!--To display step content between shipping step and payment step  1 < "sortOrder" < 2 -->
                                                     <!--To display step content after payment step "sortOrder" > 2 -->


aakimov:
Hello wildcard27,

It is not necessary for the step component to declare _provider_ property (but of course you can do it if you want). You can add this node to your child-components under <children/> node. Just like it is done in the _checkout_index_index.xml_ of the *Magento\Checkout* module.
Moreover your components can have a custom data provider. That is why this node is not included in the code snippet in documentation.

Hope this helps,
Regards.

wildcard27:
A possible solution would be to explain what the provider node is for and where it should go. The documentation is pretty rubbish when it comes to things like this.
As the code provided does not work without the _provider_ property, it should either be documented or included. Had this been the case, it would have saved me hours trying to figure it out myself.
For the record, I am still not completely sure what this does as there is **no** documentation at all.

wildcard27:
@aakimov I see that this has been added to the devdocs but without using my pull request? Shouldn't my PR be counted as a contribution?

jeff-matthews:
@wildcard27, we can definitely give you credit for this contribution on our [Quarterly contributors](http://devdocs.magento.com/guides/v2.2/contributor-guide/quarterly-contributors.html) page! 

The only place I see the `<item name="provider" xsi:type="string">checkoutProvider</item>` example is in the Static forms section here: http://devdocs.magento.com/guides/v2.2/howdoi/checkout/checkout_form.html#static-forms

Is that the example you're saying was added outside your PR?

jeff-matthews:
Closing due to inactivity

wildcard27:
@jeff-matthews Apologies for not reply to this sooner and also for commenting on a dead issue. But in answer to your question, yes, this is the code that has been merged in outside of my pull request.

Basically, it looks like the code has been added elsewhere in order to combat the same issue that resulted in me opening the pull request.

It would have been good to have received a contributer badge for this as it was most likely my thought that led to the inclusion of the `provider` reference.

PR #1138, opened 2017-05-07
Instead of saying "more installation" i changed it to "more information"

Vasilisdm:


jcalcaben:
Good catch! Thanks @Vasilisdm !

PR #1140, opened 2017-05-07
changing the "is the theme" to "if the theme"

Vasilisdm:


jcalcaben:
Thanks @Vasilisdm !

Vasilisdm:
np glad to helped just a little!

PR #1143, opened 2017-05-08
Update xml-instructions.md

Vasilisdm:
Changed to value of the remove attribute from "true" to "false" in order to go hand in hand with the description that says "to cancel removal of a block or container in your layout by setting remove attribute value to false".


almarchenko:
@Vasilisdm Thanks for drawing our attention to this! The actual mistake was in the text, not in the code. So I've changed the PR slightly and merged it.

Thank you for your input!

Vasilisdm:
@almarchenko no need to thank me at all! Keep up the good work!

PR #1144, opened 2017-05-08
Update xml-manage.md

Vasilisdm:
changed the "is specified relatively to one the following locations" to "is specified relatively to one of the following locations", so the meaning of the sentence to be more clear.


dshevtsov:
Good catch @Vasilisdm! Thank you!

PR #1147, opened 2017-05-09
Update the orderItemIdQty Def

gdoug-guidance:
orderItemIdQty Array is Keyed with Order Item ID, with Value of qty to invoice
See: Mage_Sales_Model_Service_Order::prepareInvoice()


keharper:
Thanks for noticing this.

PR #1148, opened 2017-05-09
Fixed related topic links on javascript-dev-guide/javascript/custom_j…

BenSpace48:
Fixed related topic links on javascript-dev-guide/javascript/custom_js.html

Fix the related topic links at the bottom of the page, Markdown seems to be causing issues (is it even supported on the Dev Docs?) - http://devdocs.magento.com/guides/v2.0/javascript-dev-guide/javascript/custom_js.html

### DIFF
--- a/guides/v2.0/javascript-dev-guide/javascript/custom_js.md
+++ b/guides/v2.0/javascript-dev-guide/javascript/custom_js.md
@@ -180,8 +180,8 @@ $(mage.apply);
 {% endhighlight %}
 
 
-#### Related topic
+<h2 id="related_topic">Related topic</h2>
 
-- [JavaScript resources in Magento]({{ page.baseurl }}javascript-dev-guide/javascript/js-resources.html)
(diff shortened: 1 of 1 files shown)


almarchenko:
Hi @BenSpace48 
Thank you for the PR!
Markdown actually is supported on DevDocs. And even more, we tend to use MD instead HTML wherever possible.  But in the docs created earlier, there's still HTML mostly, and sometimes the mix of HTML and MD causes problems. 
I will take a look at this page, if I do not find the solution with MD, I will happily merge your PR and leave it in HTML.
Thank you!

BenSpace48:
Ah right, I did think it was strange how there was a mixture. Thanks for the update.

almarchenko:
@BenSpace48  Looks like I've managed to fix the mix of HTML and MD we have in this topic, and the Related topics header (and everything) else is displayed correctly now
http://devdocs.magento.com/guides/v2.0/javascript-dev-guide/javascript/custom_js.html

If you do not mind, I will close this PR.
Thank you for drawing our attention to this and for the input!

BenSpace48:
Cool, thanks for fixing - I've closed this PR.

PR #1150, opened 2017-05-10
Update xlate.md

Vasilisdm:
Just changed the following "so that a system integrator can **translations** variations..." to "so that a system integrator can **translate** variations..."


almarchenko:
@Vasilisdm Thanks for a good catch!

PR #1159, opened 2017-05-12
Include reference to SnowdogApps effort

nshiff:


almarchenko:
@nshiff Thanks for your suggestion, it does look logical and relevant. I will double-check if it complies with our rules/policies. Thanks!

almarchenko:
@nshiff  Currently the decision is not to reference any 3rd party themes in our documentation. 
Thank you for the suggestion!

PR #1170, opened 2017-05-20
Fix broken link v.22

sgnsajgon:
Related to https://github.com/magento/devdocs/pull/1168 but this case is for v2.2


dshevtsov:
Thanks @sgnsajgon !

Issue #1178, opened 2017-05-24
Ambiguous test command


QwertyZW:
Feedback on page: /guides/v2.0/test/integration/integration_test_setup.html

The page concludes with 

>Such a test suite configuration can then be executed using the --testsuite command option, for example --testsuite "Third Party Integration Tests"

What command is it talking about?

It doesn't seem to be

```
php bin/magento dev:tests:run
```

nor

```
php bin/magento dev:tests:run integration
```

It's probably 

```
php ../../../vendor/bin/phpunit --testsuite <name>
```
from within `dev/tests/integration`

dshevtsov:
@QwertyZW thank you for your detailed feedback!
Please check how it looks with changes - http://devdocs.magento.com/guides/v2.1/test/integration/integration_test_setup.html

QwertyZW:
Well done!

PR #1181, opened 2017-05-25
Disambiguate command by being explicit

QwertyZW:
For the case when the user lands on said page without having gone through the preceding sections

For issue https://github.com/magento/devdocs/issues/1178


QwertyZW:
Fixed

PR #1182, opened 2017-05-26
Update adding-attributes.md

mtbottens:
Typo error fix found while browsing m2 docs


almarchenko:
Thanks for your input!!

PR #1190, opened 2017-05-31
Tweak acceptable semantic HTML markup example

tomreece:
The example for:

**You must use semantic HTML markup only, and must not use presentation markup.**

...included presentation markup (`<b>` and `<strong>`) which I believe violates the rule.

I replaced it with `<span class="...">` as an example.

### DIFF
--- a/guides/v2.0/coding-standards/code-standard-demarcation.md
+++ b/guides/v2.0/coding-standards/code-standard-demarcation.md
@@ -127,14 +127,14 @@ The following list will help you make a distinction between the actual meaning o
 **Acceptable**:
 
 {%highlight html%}
-<p>HTML has been created to <span class="font-weight-bold">semantically</span> represent documents.</p>
-<p><span class="font-weight-bold">Warning:</span> Following the procedure described below may irreparably damage your equipment.</p>
+<p>HTML has been created to <strong>semantically</strong> represent documents.</p>
+<p><strong>Warning:</strong> Following the procedure described below may irreparably damage your equipment.</p>
 {%endhighlight%}
 
 **Unacceptable**:
 
 {%highlight html%}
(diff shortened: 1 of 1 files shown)


almarchenko:
@tomreece Thanks for a good catch!
But I think the changes you suggest should be in the Unacceptable sample. 
Do you want to change your PR yourself, or we can do it instead.

Thank you.

tomreece:
I thought that `<b>` and `<strong>` were considered presentation markup since they affect style.

Is that an incorrect assumption? Is it ok to use `<b>` and `<strong>`?

hostep:
`<b>` is only presentational, but `<strong>` has a semantic meaning: https://www.w3.org/TR/html5/text-level-semantics.html#the-strong-element

I think that the examples as they are now, are correct, but I'm not 100% sure.
But the suggestion from @tomreece is very incorrect in my opinion.

tomreece:
Thanks for sharing that link @hostep. I may have misinterpreted.

However, the Acceptable vs Unacceptable examples as they were before my change only switch where the `<b>` and `<strong>` are used. It was unclear to me how one could be acceptable and the other could not.

If `<b>` is only presentational, perhaps it just needs to be excluded from the Acceptable example? I am updating my PR.

hostep:
Your new commit makes much more sense 👍 

Even though I'm still not convinced if the original or your new commit is the correct implementation.
It depends on if the purpose of the example is to show that the word 'semantically' needs to be emphasized only visually, or also for blind people/search bots/...

Maybe it's best if someone from Magento who's working on accessibility can take a look at this...

tomreece:
Upon further investigation, w3 uses a similar example, seen towards the bottom of:
https://www.w3.org/TR/html5/text-level-semantics.html#the-b-element

So maybe no change necessary. Feel free to close this PR. :)

PR #1210, opened 2017-06-09
FIX admin_apply_theme logic for EE versions

mgeoffray:
With last EE versions defining a custom admin theme is override by the `Magento_Enterprise` extension.    
For that we have to load `Magento_Enterprise` extension before the custome one.


jcalcaben:
Good catch! Thanks @mgeoffray !

PR #1216, opened 2017-06-14
Change index management submenu

surfer190:
These are the changes pertaining to a `2.1.5` magento version


jcalcaben:
Good catch! Thanks @surfer190 !

PR #1218, opened 2017-06-14
Fix mistake in code

springimport:


jcalcaben:
Good catch! Thanks @springimport !

PR #1221, opened 2017-06-15
Adding the number of the PSR specification

rafaelstz:


lorikrell:
Accepted and merging. Thank you!

PR #1236, opened 2017-06-21
1196: add the shared group to the Magento file system owner user acc…

jeff-matthews:
…ount as a secondary--not primary--group

### DIFF
--- a/_includes/install/file-system-perms-twouser.md
+++ b/_includes/install/file-system-perms-twouser.md
@@ -12,6 +12,10 @@ To enable the web server to write files and directories in the Magento file syst
 
 This section discusses how to create a new Magento file system owner and put that user in the web server's group. You can use an existing user account if you wish; we recommend the user have a strong password for security reasons.
 
+<div class="bs-callout bs-callout-info">
+	Skip to <a href="#install-update-depend-user-findgroup">step 2</a> is you plan on using an existing user account.
+</div>
+
 ### Step 1: Create the Magento file system owner and give the user a strong password {#mage-owner-create-user}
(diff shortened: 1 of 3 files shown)


_includes/install/file-system-perms-twouser.md
```diff
@@ -12,6 +12,10 @@ To enable the web server to write files and directories in the Magento file syst
 
 This section discusses how to create a new Magento file system owner and put that user in the web server's group. You can use an existing user account if you wish; we recommend the user have a strong password for security reasons.
 
+<div class="bs-callout bs-callout-info">
+	Skip to <a href="#install-update-depend-user-findgroup">step 2</a> is you plan on using an existing user account.
```
keharper: ~~is~~ > if

_includes/install/file-system-perms-twouser.md
```diff
@@ -29,13 +33,13 @@ Follow the prompts on your screen to create a password for the user.
     <p>If you don't have <code>root</code> privileges on your Magento server, you can use another local user account. Make sure the user has a strong password and continue with <a href="#install-update-depend-user-add2group">Put the Magento file system owner in the web server group</a>.</p>
 </div>
 
-For example, to create a user named `magento_user` and give the user a password, enter:
+For example, to create a user named `magento` and give the user a password, enter:
```
sgnsajgon: This line should be reverted, because in two commands below we use account named `magento_user`.

jeff-matthews: @sgnsajgon: My mistake! You're absolutely right. I'll get this updated.

PR #1258, opened 2017-06-28
Updated varnish version as it was incorrect

iancassidyweb:
Varnish 3.5 is not a Varnish 3 version, this should be 3.0.5 in accordance with the VCL file produced in the admin panel.


keharper:
Thanks for pointing this out.

PR #1260, opened 2017-06-29
Add additional requirements for cron run process

vgoncharenko:


dmrachkovskyi:
May I suggest to separate the sentences — to make them better for understanding? The result would be:
> You must run cron twice: the first time to discover tasks to run, then the second time — to run the tasks themselves. The second cron run must occur on or after the `scheduled_at` time for every task.

PR #1265, opened 2017-06-30
Add HTML attribute translation example for UI Components

claireparker:
An example for when you want to translate an element attribute, e.g. `<input type="text" placeholder="Email Address" />` in a Knockout file.


jcalcaben:
Thanks for the example @claireparker !

PR #1269, opened 2017-07-03
remove reference to partner accounts

mbrinton01:
Updated guide to remove reference to accepting partner accounts. GitHub accounts must be individual accounts to grant access.


dshevtsov:
@mbrinton01 thanks for the clarification!

PR #1271, opened 2017-07-04
Specify magento2ce directory when cloning

careys7:
Following step 1 from "Clone the Magento Github repositories" gives the user the repository inside a directory "magento2" by default, whereas the rest of the guide describes this as "magento2ce"


jeff-matthews:
Thanks for pointing out the inconsistency @careys7. The correct repo is `magento2`, so I need to change all references to `magento2ce` to `magento2`.

PR #1273, opened 2017-07-05
Update theme-apply.md

chickenland:
Reflects movement of the theme configuration settings from Store | Configuration

### DIFF
--- a/guides/v2.0/frontend-dev-guide/themes/theme-apply.md
+++ b/guides/v2.0/frontend-dev-guide/themes/theme-apply.md
@@ -25,11 +25,11 @@ After you <a href="{{page.baseurl}}frontend-dev-guide/themes/theme-create.html">
 
 To apply a theme:
 
-2. In Admin, go to **Stores** > **Configuration** > **Design**.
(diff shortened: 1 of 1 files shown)


hostep:
Hi @chickenland, sorry, you certainly aren't the first one to report this and you probably can't help it.

But the instructions for [Magento 2.0](http://devdocs.magento.com/guides/v2.0/frontend-dev-guide/themes/theme-apply.html) are correct, you should check out the instructions for [Magento 2.1](http://devdocs.magento.com/guides/v2.1/frontend-dev-guide/themes/theme-apply.html).

The only way I can see from stopping people to try to fix this, is to make the version picker much more obvious in some way, @ devdocs people, I'm sure you are already aware of this?

chickenland:
@hostep Most definitely - something bigger, and more red shouting at me would have helped me realise!  Thanks.

jcalcaben:
Thanks for pointing that out @chickenland and @hostep !

Yes, we are currently working with our UX team to come up with a better solution for version confusion in our docs.

PR #1275, opened 2017-07-10
Fix name of PhpStorm

ihor-sviziev:
This blocks is shown on following page:
http://devdocs.magento.com/guides/v2.1/install-gde/docker/docker-phpstorm-debug.html

It looks ugly.


keharper:
Thanks!

PR #1280, opened 2017-07-11
Add missing closing `<config>` tag.

dersam:


dshevtsov:
Good catch, thank you @dersam !

PR #1282, opened 2017-07-12
Add hint to preserve .htaccess file in pub/static

jerome-breton:
to prevent breaking url rewrite of assets


jcalcaben:
Thanks for the hint @jerome-breton !

PR #1290, opened 2017-07-15
Remove SSH key info from HTTPS section

sgnsajgon:
If we use HTTPS connection to clone git repository, SSH key is not used. I removed misleading information.


jeff-matthews:
Thansk @sgnsajgon!

PR #1295, opened 2017-07-19
Update css-preprocess.md

mateo-occitech:
Fix typo


jcalcaben:
Good catch @mateo-occitech ! Thanks!

PR #1298, opened 2017-07-19
Fixed Link issue

RakeshJesadiya:
Column ink redirect to 404 page.Its create confusion for user so remove link.


keharper:
Fixing the link instead.

PR #1300, opened 2017-07-19
Fixed ColumnAction,Column link redirect to 404 page.

RakeshJesadiya:
Fixed ColumnAction and Column link in a page redirect to 404 page. This Might confused to user which contains exist for a link so remove link from above both word.


keharper:
This appears to be a duplicate of PR1298, which you opened. As with PR 1304, I'm fixing the links instead of deleting them.

PR #1304, opened 2017-07-19
Fixed ColumnAction with link redirect to 404 page.

RakeshJesadiya:
ColumnAction with link redirect to 404 page so remove link from ColumnAction from the table.


keharper:
I'm going to fix the broken link, rather than delete it. The problem was that the link incorrectly pointed to `ui-column.md` instead of `ui-column.html`. Thanks for letting us know about the problem.

PR #1305, opened 2017-07-20
Fixed link with typo mistakes.

RakeshJesadiya:
Fixed target link with typo mistakes in jQuery widget instead of Query.


jcalcaben:
Thanks for the fix @RakeshJesadiya !

PR #1311, opened 2017-07-23
Added listening port configuration for Varnish 4.*

rogyar:
The previous description "If necessary, comment out the following:" was not clear enough. Moreover, for Varnish 4.* VARNISH_LISTEN_PORT is not actual anymore, you need to edit the daemon parameters for this purpose (https://varnish-cache.org/docs/4.0/tutorial/putting_varnish_on_port_80.html). I've clarified the correct way to set the listening port for varnish 4.*


jeff-matthews:
Thanks for the contribution @rogyar!

PR #1312, opened 2017-07-24
How to use Tabs in template file.

RakeshJesadiya:
Set example for use tabs in template file.


jcalcaben:
Thanks again @RakeshJesadiya !

PR #1314, opened 2017-07-24
Update config-cli-subcommands-cache.md

convenient:
Just to _really_ clarify that this cache contains the merged configuration from both the file system as well as the database.


jeff-matthews:
Thanks for the contribution @convenient!

PR #1315, opened 2017-07-25
Fixed link issue.

RakeshJesadiya:
Php Modifier link not found for magento 2.2


keharper:
The problem here was that the target file didn't get added to the 2.2 directory structure. I'll be checking in that file and keep the link intact.

PR #1316, opened 2017-07-25
Set PoolInterface.php file link.

RakeshJesadiya:
Set dynamic Poolinterface.php file link from github Magento 2.1


keharper:
Thanks!

PR #1318, opened 2017-07-25
Massactions Link redirect to 404 page.

RakeshJesadiya:
Set Proper link for MassActions Components.


keharper:
It looks like we restructured the directory structure of the ui_components book at some point but forgot to update all of the links. I'll merge this instance and push updates to the other files later today.

PR #1322, opened 2017-07-26
Checkbox set link not working.

RakeshJesadiya:
Checkbox set link not working.


keharper:
Thank you for your many contributions. 

You've found several instances of a weird problem which requires a fix that is different than what you checked in. 

In this case, the "c" in "ui-сheckboxset.html" wasn't really a "c". It was a character from a non-English character set. In some cases, I think, the special character was in the file name. In general, I've been re-typing the link and renaming the file to make sure I've gotten rid of the the special character. 

So, I won't be merging this PR, and perhaps others, but I do appreciate your efforts.

PR #1326, opened 2017-07-26
Fix formatting and add updated info to v2.2

jeff-matthews:


jeff-matthews:
Supplements PR [#1311](https://github.com/magento/devdocs/pull/1311)

PR #1335, opened 2017-07-30
<vendor-name>/module-<component-name> is unnecessary restrictive

fooman:
The naming convention `<vendor-name>/module-<component-name>` only makes sense for Magento internal packages as there is some processing happening with it here
https://github.com/magento/magento2/blob/2.0.0/lib/internal/Magento/Framework/Module/PackageInfo.php#L153

For 3rd party module names there is no such processing happening and therefore module- does not need to be part of name.


jcalcaben:
Good point. Thanks @fooman !

PR #1338, opened 2017-08-02
Removes superfluous 'a' in the menu title

digitalbricklayer:
Removes a superfluous 'a' in the menu title on the "Install a third-party storefront theme" page.


jcalcaben:
Good catch @digitalbricklayer !

PR #1363, opened 2017-08-19
Update Magento Opensource and Commerce namespace

RakeshJesadiya:
Update Magento Opensource and Commerce namespace


keharper:
Thanks!

PR #1369, opened 2017-08-19
Magento Label Update

RakeshJesadiya:
Update Magento Commerce instead of EE.


dshevtsov:
Thanks @RakeshJesadiya !

PR #1378, opened 2017-08-20
Add note on how to keep code backward compatible

Vinai:
Code using the `SerializerInterface` as a constructor dependency will be incompatible with Magento 2.0 and 2.1.  
Having to maintain separate branches for different Magento versions is a burden for extension developers. The new paragraph suggests a way to deal with this issue.

### DIFF
--- a/guides/v2.2/extension-dev-guide/framework/serializer.md
+++ b/guides/v2.2/extension-dev-guide/framework/serializer.md
@@ -120,7 +120,7 @@ The `SerializerInterface` interface and its implementations only exist since Mag
 Because of this, it is not posssible to use these classes in code that has to be compatible with Magento 2.1 or 2.0.  
 
 In code that is compatible with earlier versions of Magento 2, constructor dependency injection can not be used to get an instance of `SerializerInterface`.  
-Instead, a runtime check against the Magento 2 version has to be made, and if the version is 2.2.0 or newer, the `SerializerInterface` can be instantiated by directly accessimg the object manager using a static method.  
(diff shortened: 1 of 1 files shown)


guides/v2.2/extension-dev-guide/framework/serializer.md
```diff
@@ -113,3 +113,36 @@ public function loadDataFromCache()
...
+{% highlight php startinline %}
+
+```
+/**
+ * @param mixed $data
+ * @return string
+ */
+ private function serialize($data)
+ {
+    if (class_exists(\Magento\Framework\Serialize\SerializerInterface::class)) {
```
hostep: Wouldn't this throw a php warning when the class `\Magento\Framework\Serialize\SerializerInterface` doesn't exist in 2.1 or 2.0? It would probably be better to use a string representation of the class name I think?

Vinai: No, it does not throw an exception or raise a warning.
The PHP "compiler" will apply namespace rules to `::class` constants and then uses it as a string, so there is literally zero difference between
`if (class_exists(\Magento\Framework\Serialize\SerializerInterface::class))`
and
`if (class_exists('\Magento\Framework\Serialize\SerializerInterface'))`
except that the former allows static code analysis to do a better job.

hostep: Hi @Vinai, I just tested this, you are absolutely correct, learned something new today, thanks! :)

PR #1380, opened 2017-08-21
Update prod_perf-optimize.md

kandy:
Other command does not need for enable production mode especially on deployment on production

### DIFF
--- a/guides/v2.2/config-guide/prod/prod_perf-optimize.md
+++ b/guides/v2.2/config-guide/prod/prod_perf-optimize.md
@@ -117,7 +117,7 @@ Switching to production mode improves store front responsiveness and prevents lo
 Run the following commands to switch to production mode:
 
 ~~~
-bin/magento deploy:mode:set -s production
+bin/magento deploy:mode:set production
 ~~~
 
 ### Optimizing Asset Performance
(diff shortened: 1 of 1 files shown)


jeff-matthews:
@kandy: Do these changes apply to versions 2.0 and 2.1 as well? If so, please update the following files:

-  guides/v2.0/config-guide/prod/prod_perf-optimize.md
-  guides/v2.1/config-guide/prod/prod_perf-optimize.md

hostep:
@kandy: when the `-s` flag is used, you need to regenerate the static assets and run the compiler, unless they are already up to date, right?
The docs say:
```
➜ php bin/magento deploy:mode:set --help
Usage:
  deploy:mode:set [options] [--] <mode>

Arguments:
  mode                    The application mode to set. Available options are "developer" or "production"

Options:
  -s, --skip-compilation  Skips the clearing and regeneration of static content (generated code, preprocessed CSS, and assets in pub/static/)
  ...
```

(Unless this will be changed in 2.2? I haven't tested this out yet.)

Also, slightly relevant, maybe while you're at it, I think the command: `bin/magento deploy:mode:set -s production` is missing on this page: http://devdocs.magento.com/guides/v2.1/config-guide/prod/single-machine-deployment.html
I noticed this some days ago but failed to create an issue or PR for this, sorry!

jeff-matthews:
@kandy: Please confirm whether this affects all versions or just v2.2.

jeff-matthews:
I confirmed with an internal product manager that this behavior is consistent across all version.

PR #1383, opened 2017-08-22
fixed some whitespaces and changed link 

roman204:
- i  have found some mising whitespaces
- changed a link from an 404 page to the js-video


jcalcaben:
Thanks @roman204 !

PR #1394, opened 2017-08-28
Update theme-create.md

Draeli:
Actually, composer.json is require to saw theme in configuration theme.
I add section "trouble" because I saw some time create file and clear cache isn't enough.


jfrontain:
Thanks, @Draeli! We'll merge your change.

PR #1395, opened 2017-08-28
nginx config did not match command

krebstar2000:
The doc was checking for $MAGE_ROOT/var/maintenance.flag whereas the command said "touch <your Magento install dir>/maintenance.enable".

Alternatively the command could be changed to match the nginx config. "touch <your Magento install dir>/var/maintenance.flag"


jeff-matthews:
Thanks for the contribution @krebstar2000!

PR #1408, opened 2017-09-11
Update tshoot-varnish-503.md

klakovskiy:
Then you extends http_resp_size you need to extend workspace_backend, because http_resp_size  used this memory, but default workspace_backend = 64k. 

* The memory for the request is allocated from the backend workspace (param: workspace_backend) and this parameter limits how much of that the request is allowed to take up.

* link: https://varnish-cache.org/docs/4.1/reference/varnishd.html#http-resp-size


jeff-matthews:
Thanks @klakovskiy

PR #1412, opened 2017-09-13
I think this is a mistake

sanuter:
I think this is a mistake


osrecio:
This Pull Request is submitted [here](https://github.com/magento/devdocs/pull/1411)

jcalcaben:
Good catch @sanuter ! This has been fixed in the previously linked PR.

PR #1413, opened 2017-09-13
Rename Enterprise Edition to Magento Commerce

osrecio:
Small change following the renaming criteria about platforms


dshevtsov:
Good catch, thanks @osrecio !

PR #1415, opened 2017-09-13
Update using_grunt.md

oddyssey:
Correct the section on using the new grunt-config.json file.


bdenham:
@oddyssey, Thanks for the update!

PR #1418, opened 2017-09-15
Correct Code Example

pmclain:
This change corrects the parent class name used in the code example. Constants were added, making the example consitent with cache type code used in core modules.


jeff-matthews:
Thanks @pmclain!

PR #1419, opened 2017-09-15
Fix the links to EE

rafaelstz:


jfrontain:
Thank you for catching that error, @rafaelstz. Clearly I am copying and pasting too quickly. I appreciate your attention to detail

PR #1420, opened 2017-09-16
Set dynamics links in bootstrap.php file

RakeshJesadiya:
Set dynamics links in bootstrap.php file


jeff-matthews:
I'm not sure if this is correct... They way I understand it, `index.php` contains a reference that pulls in `[app/bootstrap.php](https://github.com/magento/magento2/blob/2.0/index.php#L22)`. It looks like you're replacing the first link to `index.php` with a link to `app/bootstrap.php` and replacing the second link to `app/bootstrap.php` with a link to `index.php`.

PR #1421, opened 2017-09-16
Theme defination with default theme elaborate.

RakeshJesadiya:
Theme defination with default theme elaborate.


jfrontain:
Thanks, @RakeshJesadiya! I've edited your text and will merge to devdocs. Thanks for submitting to the doc set!

PR #1433, opened 2017-09-30
Remove extra </tr> on module packaging guide

renttek:
On the page "/guides/v2.2/extension-dev-guide/package/package_module.html" is an extra "</tr>" which is rendered as content instead of html, because there is no matching opening tag.

Exact position of the is beneath the headline "Create a Magento Composer file", right above the table.


keharper:
The problem was that there was a missing `<tr>` immediately before the `<td>` containing`autoload`. Since the table also contained a markdown cross-reference that was resolving, I converted the whole table to markdown, bypassing the original problem all together.

Thanks for making us aware of the problem. I'll be pushing out the fix soon.

PR #1435, opened 2017-10-02
Fixing link to the B2B Quick Tour

brobie:


keharper:
It looks like we've fixed that already. Thanks for your diligence.

PR #1443, opened 2017-10-09
Add config file to the app:config:dump command

boldbart:
The current description only mentions the update to app/etc/env.php after running app:config:dump, but it also updates app/etc/config.php with shared configuration for all Magento instances.


jeff-matthews:
Thanks @boldbart!

PR #1444, opened 2017-10-09
Remove tooltip from code block

robinj:
The tooltip declaration is causing the code block to mess up and should be removed. Tooltips and plain text don't mix well


jcalcaben:
Good catch! Thanks @robinj !

PR #1449, opened 2017-10-11
Fix typo: grount to grunt

lazyguru:
The left sidebar nav currently says

> Using Grount for Magento tasks

It should read:

> Using Grunt for Magento tasks

This fixes it by removing the extraneous "o" in Grount


jcalcaben:
Good catch! Thanks @lazyguru !

PR #1450, opened 2017-10-11
Fix typo in regular expression string

jahvi:
It should say "regular expressions" instead of "regular exceptions".


keharper:
Chuckling at this typo. Thanks for fixing it.

hostep:
@jahvi, @keharper: this fix is incorrect.

The line which needs to be changed is:
`In the Search String box specify the user-agent using either normal strings or regular exceptions (PCRE).`

The line which is now changed was originally correct.

@jahvi: can you create a new PR which fixes this? Thanks!

jahvi:
@hostep My bad, I'll create a new PR with the correct fix.

PR #1453, opened 2017-10-12
Fixed missing letter in dev docs about Cron

rickkuilman:
Fixed missing letter in dev docs about Cron


jeff-matthews:
Thanks @rickkuilman!

PR #1456, opened 2017-10-14
Revert "Fix typo in regular expression string"

jahvi:
Reverts magento/devdocs#1450 and fixes the right string.


keharper:
Thanks!

PR #1460, opened 2017-10-17
Version changed from 2.1.2 to 2.2 in examples

rogyar:
The change is about changing Magento version in examples from 2.1 to 2.2. Since this document should be actual for Magento 2.2 it makes sense to use the actual version (might help with copy-paste troubles for some cases)


jeff-matthews:
Thanks @rogyar!

PR #1462, opened 2017-10-17
Moved misplaced text outside of example code block

rickkuilman:
It looks like the text "Assignment operations to constructor prototypes creating temporal coupling and sometimes other unwanted side effects." should be outside (after) the example code block instead of in the code block


jcalcaben:
Good catch! Thanks @rickkuilman !

PR #1466, opened 2017-10-18
Paths for migration configuration changed

rogyar:
In the migration tool versions after 2.2 the configuration paths like "ce-to-ce" have been changed to "opensource-to-opensource" in order to match the actual version name. The adjustment brings the actual paths to the documentation.


jeff-matthews:
Thanks @rogyar!

PR #1489, opened 2017-10-28
Update config-cli-subcommands-static-view.md

likemusic:
Rename "<lang>" "<list of locales>" to "languages" (according to command help output).


jeff-matthews:
@likemusic: Thanks for the contribution! Please update the v2.0 and v2.1 topics so that the content is consistent across all versions.

likemusic:
@jeff-matthews Updated v2.0 and v2.1.

PR #1508, opened 2017-10-30
Update tech-stack.md

likemusic:
Update links to "System Requirements" pages.

whatsnew
Replaced the list of minor releases in the [Technology Stack](http://devdocs.magento.com/guides/v2.0/architecture/tech-stack.html) topic with a generic reference to system requirements


keharper:
@likemusic Thanks for all your contributions. In this case, rather than adding a new link, it would be better to delete that bulleted list of releases and change the end of the preceding paragraph to 

...see the [System Requirements]({{page.baseurl}}install-gde/system-requirements-tech.html).

We don't need to maintain a list of ever-growing list of releases here. Readers can select a different release if they need to quickly navigate to a different release than the current one.

likemusic:
Updated

PR #1510, opened 2017-10-30
Fix indexes classes

ihor-sviziev:
Update classes for indexes according to https://github.com/magento/magento2/blob/2.2/app/code/Magento/Catalog/etc/mview.xml


ihor-sviziev:
@jeff-matthews could you process my PR?

jeff-matthews:
Sorry for the delay @ihor-sviziev. Thanks for the contribution!

PR #1512, opened 2017-10-30
Fix directory pathes in framework.md

likemusic:


likemusic:
> * `/lib/internal` contains some non-PHP as well as PHP components. Non-PHP framework libraries includes {% glossarytooltip 312b4baf-15f7-4968-944e-c814d53de218 %}JavaScript{% endglossarytooltip %} and LESS/CSS.

This text should be actualized too, but I don't now how. In my current installation v2.2 CE directory `/lib/internal` contains just `LinLibertineFont` directory with ttf-font-files and some text files. It not contains any PHP files, but maybe it has in other version or edition.

Issue #1515, opened 2017-10-31
Outdated info about module's code placement


likemusic:
Feedback on page: /guides/v2.2/architecture/archi_perspectives/components/modules/mod_intro.html

Seems that section "Where do modules live?" must be updated because modules installed by composer placed in `vendor` directory. Customer module of Magento can be found at `vendor/magento/module-customer`.

Most likely that module autoloader supports loading modules from `app/code/<Vendor>/<ModuleName>` but seems that this is an outdated way. If it's true that should be reflected in the documentation.

PR #1518, opened 2017-10-31
Update mod_intro.md for v2.0

likemusic:
Update "Where do modules live?" according to http://devdocs.magento.com/guides/v2.2/architecture/archi_perspectives/components/modules/mod_intro.html


likemusic:
#1515 Update info according to http://devdocs.magento.com/guides/v2.2/extension-dev-guide/prepare/prepare_file-str.html

PR #1519, opened 2017-10-31
Separate Deploy Code and Migrate static

kimbthomas:
Instructions were missing the actual command to deploy code.
Deploy code and migrate static data should be separate tasks.
Deploy code is the task that will be used the most, especially after going live.


lorikrell:
Reviewing the PR for updating documentation. Thanks!

PR #1524, opened 2017-11-02
Update mod_intro.md

likemusic:
Add <type> - placeholder to module path. Use Customer Import/Export module instead of Customer module to show hyphen using in module name.

### DIFF
--- a/guides/v2.0/architecture/archi_perspectives/components/modules/mod_intro.md
+++ b/guides/v2.0/architecture/archi_perspectives/components/modules/mod_intro.md
@@ -34,9 +34,14 @@ See <a href="{{page.baseurl}}architecture/archi_perspectives/components/modules/
 
 ## Where do modules live?
 
-Modules typically live in the `vendor` directory of a Magento installation, in a directory with the following PSR-0 compliant format: `vendor/<vendor>/<type>-<module-mame>`. For example, the Customer Import/Export module of Magento can be found at `vendor/magento/module-customer-import-export`.
+Modules typically live in the `vendor` directory of a Magento installation, in a directory with the following PSR-0 compliant format: `vendor/<vendor>/<type>-<module-mame>`, where `<type>` can be one of values:
+ - `module` - for modules (`module-customer-import-export`);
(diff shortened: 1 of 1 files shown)


guides/v2.0/architecture/archi_perspectives/components/modules/mod_intro.md
```diff
@@ -34,7 +34,7 @@ See <a href="{{page.baseurl}}architecture/archi_perspectives/components/modules/
 
 ## Where do modules live?
 
-Modules typically live in the `vendor` directory of a Magento installation, in a directory with the following PSR-0 compliant format: `vendor/<vendor>/module-<module-mame>`. For example, the Customer module of Magento can be found at `vendor/magento/module-customer`.
+Modules typically live in the `vendor` directory of a Magento installation, in a directory with the following PSR-0 compliant format: `vendor/<vendor>/<type>-<module-mame>`. For example, the Customer Import/Export module of Magento can be found at `vendor/magento/module-customer-import-export`.
```
jeff-matthews: `<module-mame>` should be `<module-name>`

Also, why replace `module` with `<type>`? Doesn't this contradict Magento's module naming convention?

likemusic: `<type>`  can be one of values: 
 - "module" - for modules (`module-customer-import-export`);
 - "theme" -  for front and admin themes (`theme-frontend-luma`);
 - "language" - for language packs (`language-de_de`);

May be add this info to doc?

jeff-matthews: Good point! You're absolutely right. Sorry, I was too focused on the word "module".  Yes, I think it would be beneficial to add these to the docs as well. Do you mind adding them?

likemusic: Ок, I'll add this info to doc.

likemusic:
Info about module types added. But I have not good English level, therefore you can check it and fix before merging :)

jeff-matthews:
Thanks @likemusic!

PR #1525, opened 2017-11-02
Update discover-deploy.md

andriyShevtsov:
Lori, i found some wrong info and change it. Could you please check it


lorikrell:
@andriyShevtsov  Thanks! I'll get the information added to all of the versions of the file and another matching file for Starter. I'll also get a ticket created to track.

PR #1530, opened 2017-11-05
Vagrant in root

likemusic:
Seems that there is no sense to have separated subdirectory `vagrant` in witch cloned `develop` branch. More obvious and simple to start vagrant on current sources version, without duplicating sources and managing routines for syncing.


dshevtsov:
@likemusic , thanks for your feedback!
Vagrant project will be moved to a separate repo to avoid duplication. `devdocs/README.md` will be updated according to this change.

Created internal issue: MAGEDOC-2664

PR #1533, opened 2017-11-06
Suggestion - bold headers

likemusic:
Currently evaluated styles for fonts in headers: 
 - h1 - 38px normal
 - h2 - 26px normal
 - h3 - 22px; bold

Visualy h3 seems had more weight than h2, because it had bold font weight.

Perhaps it would be better to set font weight for all header the same?
Imho, bold headers seems better.


jeff-matthews:
Closing since we're using a theme gem now. All style updates should be made there. 

https://github.com/magento-devdocs/devdocs-theme

PR #1536, opened 2017-11-06
Fix about 4000 broken links

likemusic:
I understand that most likely this pool-request will be rejected. But still decided to offer it, since already did.

It reduces the number of broken links by about 4000 (and not introduce new).

After that pull-request'and will be approximately 1443 broken links of which 1148 are internal.


jeff-matthews:
Thanks for your effort @likemusic. We recently added a rake task that checks links to the devdocs repo and are actively working on resolving errors, which number about 1100 the last time I ran it.

likemusic:
There is still 1121 broken internal links in devdocs site.
List of broken links: https://docs.google.com/spreadsheets/d/1wpCPj8sQQIKimBhHMLCuu17xN-it2vHnKGZqVJFlT1o/edit?usp=sharing

I think this issue should be kept opened until all broken links will not be fixed.

PR #1577, opened 2017-11-29
Update module-load-order.md

JosephMaxwell:
While the description is mostly helpful, I keep forgetting who is dependent on who. The goal is that this clears it up and makes it very easy to understand.


jcalcaben:
Thanks for the clarification @JosephMaxwell !

PR #1582, opened 2017-12-04
Update layout-override.md

lijithmg:
Corrected the filename of the overridden file from layout1.xml to layout2.xml


jcalcaben:
Good catch! Thanks @lijithmg !

PR #1599, opened 2017-12-12
Fix the sed filter in mysqldump command

yyevgenii:
The actual version of sed filter in mysqldump command doesn't work, it fails with the following error:
```
sed: -e expression #1, char 20: Invalid preceding regular expression 
mysqldump: Got errno 32 on write
```

The suggested fix works fine :)


yyevgenii:
You can use the following Magento code as a reference:
https://github.com/magento/magento2ee/blob/develop/app/code/Magento/Support/Console/Command/BackupDbCommand.php#L126

There you can see the same sed filter but with an -e option which is not necessary in our case, so I've skipped it in my command. Also I've added the "g" modificator which tells sed to filter all occurrences in the line (default behavior is to filter the only first matched occurrence in each line).

jeff-matthews:
Thanks @yyevgenii!

PR #1603, opened 2017-12-12
Correct spelling mistake

adam-paterson:


jfrontain:
Thanks, @adam-paterson, for catching this!

PR #1608, opened 2017-12-14
Change varnish config path for Ubuntu LTS 16.04

sadon:
/lib/systemd/system/varnish.service


sadon:
for able to use 
`service varnish restart`

PR #1610, opened 2017-12-15
Correct spelling error

georanma:
Step 5 currently says "You need to be carfull". I've updated the error. In general, this could be worded better. 
Possible suggestion: 
It is is important to practice caution while working through manual data changes. Mistakes will create errors in the incremental data migration step that follows.


jeff-matthews:
Thanks @georanma. See https://github.com/magento/devdocs/pull/1611.

PR #1614, opened 2017-12-19
Fix a mistake in the SSH port forwarding command

yyevgenii:
The OpenSSH official documentation says:

The -R option:
Specifies that connections to the given TCP port or Unix socket on the
remote (server) host are to be forwarded to the local side.

The -L option:
Specifies that connections to the given TCP port or Unix socket on the
local (client) host are to be forwarded to the given host and port, or
Unix socket, on the remote side.

The reference: https://man.openbsd.org/ssh

So, the -R option redirects a local port to a remote machine, and -L
redirects a remote port to a local machine. When we use Xdebug we need
the remote port to be forwarded to the local machine. That's why we need
the -L option instead of -R one.

I've checked other cloud articles and found that we use SSH port
forwarding only in the following article (except Xdebug):
guides/v2.0/cloud/project/project-conf-files_services-rabbit.md
and that article has correct "-L" flag.

### DIFF
--- a/guides/v2.0/cloud/howtos/debug.md
+++ b/guides/v2.0/cloud/howtos/debug.md
@@ -109,7 +109,7 @@ To set up port forwarding on a Mac or in a Unix environment, you will enter a co
 1. Open a terminal.
 2. Enter the following command:
 
-        ssh -R 9000:localhost:9000 <ssh url>
+        ssh -L 9000:localhost:9000 <ssh url>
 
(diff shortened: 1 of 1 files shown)


jeff-matthews:
Thanks @yyevgenii!

yyevgenii:
@jeff-matthews my fault, I thought that the PHP listens on the 9000 port and IDE connects to it for debugging. But it works in opposite way! Could you please revert this change? Thank you and sorry for inconvenience.

PR #1617, opened 2017-12-20
Fixed PHPUnit version

buskamuza:
See https://github.com/magento/magento2ce/blob/0e445495f1471738fb5f3b09e119a47ef4b78fd2/composer.json#L76


jeff-matthews:
Thanks @buskamuza!

PR #1620, opened 2017-12-20
Changed parameters order for non-default browser

roman-snitko-vaimo:
It's better to change parameters order, because in some cases this command throws the following error: "Exception in thread "main" com.beust.jcommander.ParameterException: Unknown option: -Dwebdriver.chrome.driver=..."


dshevtsov:
@roman-snitko-vaimo 
Thanks for your contribution!

PR #1621, opened 2017-12-22
Update requirements

isitnikov:
Because mhash is obsolete and since 5.3.0 php emulates mhash functions through the HASH extension I decided to change it in our official requirements.
More that we had an internal ticket MDVA-7831 where the merchant was confused by this fact.
I checked composer.json and also found that ext-hash exists in requirements but ext-mhash is absent.
Thank you.


jeff-matthews:
Thanks @isitnikov!

PR #1647, opened 2018-01-10
Add information about Elasticsearch index prefix to the ES configuration page

yyevgenii:
I've added information about index prefixes in Elasticsearch configuration. It's very important thing if you share single ES instance, like on the Magento Cloud staging/production environments.

whatsnew
Added information about the index prefix when [configuring Elasticsearch](http://devdocs.magento.com/guides/v2.2/config-guide/elasticsearch/configure-magento.html), which is necessary when using a single Elasticsearch instance with multiple Magento installations, like Staging and Production environments.


yyevgenii:
Thank you @jeff-matthews ! A lot of Magento users have fallen into this trap :)

PR #1650, opened 2018-01-11
Fixed small 'redeploy' typo.

hostep:
I don't think the whitespace changes will screw something up, but if they do, let me know and I can remove them from the commit.


jeff-matthews:
Thanks @hostep

PR #1656, opened 2018-01-19
resolve grammar issue

neeta-wagento:
Hello, 
I have resolved some grammar issues like it should be 

1 - All required configuration and entities are added in the module on the payment provider integration 
[ All required configuration and entities are added to the module of the payment provider integration. ].

2 - Using stored tokens to place order from Admin panel.
 [it should be Using stored tokens to place an order from Admin panel]


jcalcaben:
Good catch! Thanks @neeta-wagento !

PR #1660, opened 2018-01-23
Improve install PHP 7 command line

dankhrapiyush:
php7.0 and php7.1 were added two times in install php7 command line


jeff-matthews:
Thanks @dankhrapiyush!

PR #1661, opened 2018-01-23
Correct information about Fastly in "Go live checklist"

yyevgenii:
The first commit merges two paragraphs related to the TLS setup when Fastly service is in use. They were duplicating each other.

The second one corrects the DNS setup section. The previous version didn't outline correct DNS setup on production environment. It could fit to the new Starter plan, but I think we have to have information about Pro plan for first as most of Cloud clients use this plan, not Starter, and add the separate section for Started plan later.

Let's discuss the DNS setup topic if you have any thoughts about this.

### DIFF
--- a/guides/v2.0/cloud/live/go-live-checklist.md
+++ b/guides/v2.0/cloud/live/go-live-checklist.md
@@ -57,11 +57,8 @@ Many other providers also offer workarounds to accomplish this goal. The most co
 
 For the format, we support `www.domain.tld CNAME <environment>-<project>.<region>.magentosite.cloud`. We don't support a domain without www: `domain.tld CNAME <environment>-<project>.<region>.magentosite.cloud`.
 
-### Configure DNS for Fastly {#fastly-dns}
-We provide a Domain Validated SSL certificate with Subject Alternative Name enabled, issued by GLobalSign. The domain validation process is executed by Fastly. When you are ready to go, you will need to provide your domain names in a Support ticket to us. We will then provide a DNS TXT record to add to your apex domain to confirm ownership.
-
 ### TLS and Fastly {#fastly-tls}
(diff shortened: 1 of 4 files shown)


guides/v2.0/cloud/live/go-live-checklist.md
```diff
@@ -55,13 +55,10 @@ The following is a possible list of DNS providers strictly for your information.
 
 Many other providers also offer workarounds to accomplish this goal. The most common is to add a CNAME record for the `www` host on the domain and then use the DNS provider's redirection service to redirect the apex over to the `www` version of the domain. Consult your DNS provider to see how they support this.
 
-For the format, we support `www.domain.tld CNAME <environment>-<project>.<region>.magentosite.cloud`. We don't support a domain without www: `domain.tld CNAME <environment>-<project>.<region>.magentosite.cloud`.
-
-### Configure DNS for Fastly {#fastly-dns}
-We provide a Domain Validated SSL certificate with Subject Alternative Name enabled, issued by GLobalSign. The domain validation process is executed by Fastly. When you are ready to go, you will need to provide your domain names in a Support ticket to us. We will then provide a DNS TXT record to add to your apex domain to confirm ownership.
+Another option for <em>apex domain</em> is to add an A record which maps a domain name to the Fastly IP address: `150.101.113.124`.
```
hguthrie: Is this IP address the same for everyone or is there a more generic way to present this, such as where to find the Fastly IP address?

yyevgenii: @hguthrie This IP is the same for every Magento Cloud customer, it's not a mistake. Actually, it's the IP address where the DNS name prod.magentocloud.map.fastly.net points to. Despite it looks like a single host and single point of failure, it isn't, because uses an IP multicast technology.

PR #1665, opened 2018-01-24
Logs

yyevgenii:
I've replaced the location of deploy.log on the Integration environment: /tmp/log/deploy.log => /var/log/deploy.log

Actually this log is located in the /tmp/log directory, it isn't a mistake, and /var/log is a symbolic link to the /tmp/log. I think it less confuses users if they can find logs in a usual UNIX location: /var/log directory. Also it's easier to remember where to find system logs: only in the /var/log.

Second commits unifies the location of var/log/cloud.log: the /app/var/log/cloud.log is a valid path only on the Integration environment.

### DIFF
--- a/guides/v2.0/cloud/project/project-start.md
+++ b/guides/v2.0/cloud/project/project-start.md
@@ -88,7 +88,7 @@ You can review these logs via SSH into the environment. Change to the directorie
 
 Logs from the deploy hook are located on the server in the following locations:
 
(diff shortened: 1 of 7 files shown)


yyevgenii:
@jeff-matthews so fast! Thank you :)

jeff-matthews:
Thanks for opening PRs @yyevgenii!! They're very helpful. Please keep it up.

PR #1666, opened 2018-01-25
Improve UI Components Guide outline

Yogeshks:
Fieldset and Field mentioned two times in the list of Magento UI components


dshevtsov:
Good catch!
Thanks @Yogeshks !

PR #1667, opened 2018-01-25
Fix formatting of code blocks in stage-prod-migrate

yyevgenii:
Please see the "Migrate static files"section in the [stage-prod-migrate article](http://devdocs.magento.com/guides/v2.0/cloud/live/stage-prod-migrate.html): some code blocks are prepended with spaces and are shown as code blocks, but in the same time they contain apostrophe signs which outputs to the page as is. This commit removes unneeded apostrophes in two code blocks.


jeff-matthews:
Thanks @yyevgenii!

PR #1669, opened 2018-01-25
Changed from 'couple' to 'following' as there are 3 steps

asrar7787:
Changed from 'couple' to 'following' as there are 3 steps

### DIFF
--- a/guides/v2.0/test/unit/unit_test_execution_phpstorm.md
+++ b/guides/v2.0/test/unit/unit_test_execution_phpstorm.md
@@ -11,7 +11,7 @@ functional_areas:
   - test
 ---
 
-Running tests in PHPStorm requires a couple steps to setup the system.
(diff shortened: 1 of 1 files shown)


dshevtsov:
Thanks, @asrar7787

guides/v2.0/test/unit/unit_test_execution_phpstorm.md
```diff
@@ -11,7 +11,7 @@ functional_areas:
   - test
 ---
 
-Running tests in PHPStorm requires a couple steps to setup the system.
+Running tests in PHPStorm requires following steps to setup the system.
```
jeff-matthews: This sentence needs an article between "requires" and "following". I also think it can be simplified. For example,
>Running tests in PHPStorm requires **the** following steps:

PR #1677, opened 2018-01-29
replace the extension the extension with the module as it is in v2.0

neeta-wagento:
Hello , 

There should be "the module" instead of "the extension the extension"
Refer below screenshot.
install manage and upgrade extensions magento 2 developer documentation


jeff-matthews:
Thanks for the contribution @neeta-wagento!

PR #1679, opened 2018-01-30
Proper Understanding of Dependencies exact meaning

triabhishek95:
I did this changes just to make the sentence proper & make it better to be understood


jcalcaben:
Hi @triabhishek95 ,
Can you explain how the original sentence was confusing?

Would the following sentence make more sense?
>The dependencies usually declared by A are class interfaces, and the dependencies provided by B are concrete implementations for those interfaces.

triabhishek95:
The sentence should be interdependent to each other as of we are talking about dependency over here. As you can see in the present sentence it is not presenting that how A & B relates each other? It is looking like that A is used for the declaration purpose & B is providing the dependencies implementation (But to whom it will provide this?).

jcalcaben:
You are correct, **A** declares the interfaces and **B** provides the actual implementation classes. The connection between the two is implied where it says, "for those interfaces", as in the interfaces declared by **A** mentioned in the first clause of the sentence.

PR #1681, opened 2018-01-30
Updated naming convention for Plugin classes.

joni-jones:
The "Plugin" word is redundant for a class name because the namespace already contains it. All plugins located in the same folder for each module and it's not necessary naming classes with "Plugin" word.

It's more important to create meaningful names instead `ModelNamePlugin`.


jcalcaben:
Nice contribution! Thanks @joni-jones !

PR #1682, opened 2018-01-31
Remove recommendation to use Memcache instead of Redis for session du…

chris-pook:
…e to locking issues

The session locking issue was fixed for the 2.1 release with the updated "colinmollenhour/php-redis-session-abstract" dependency to 1.3.4, containing this fix: https://github.com/colinmollenhour/php-redis-session-abstract/commit/6f005b2c3755e4a96ddad821e2ea15d66fb314ae.

Redis should be used for session over Memcache if available as the better back-end, now with no locking issues.

### DIFF
--- a/guides/v2.1/config-guide/memcache/memcache.md
+++ b/guides/v2.1/config-guide/memcache/memcache.md
@@ -1,1 +1,26 @@
-../../../v2.0/config-guide/memcache/memcache.md
+---
+layout: default
+group: config-guide
+subgroup: 10_mem
+title: Use memcached for session storage
+menu_title: Use memcached for session storage
+menu_order: 1
+menu_node: parent
+version: 2.0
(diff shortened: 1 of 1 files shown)


jeff-matthews:
@dshevtsov, please verify with internal devs.

jeff-matthews:
@chris-pook, it doesn't appear that you've made any changes to the text to recommend Redis over memcache for session storage. It looks like you've only removed the symlink between 2.0 and 2.1, but copied the text from the 2.0 topic to create the new 2.1 topic. Am I missing something?

jeff-matthews:
We haven't heard from you regarding this issue in two weeks. Due to inactivity, we are closing this issue. If you have further comments and feedback, please create a new issue. Thanks for taking the time to create a GitHub issue for docs improvement!

PR #1686, opened 2018-02-01
Corrected technical inaccuracy of plugin documentation

davidalger:
Final public methods cannot be intercepted, but public methods on a class which also has final public methods can be. Also unified the note about interception bootstrapping.


keharper:
Thanks @davidalger

PR #1694, opened 2018-02-06
Add Video to keep update Repo

osrecio:
Add video to keep update repo: https://www.youtube.com/embed/mJDCL0uzIpY

I don't know if is the best option to add directly a iframe. If you have any suggestion we can discuss about it

whatsnew

Added a [descriptive video](http://devdocs.magento.com/guides/v2.0/contributor-guide/contributing.html) showing how to keep your fork updated with the latest changes.


jeff-matthews:
Thanks @osrecio! Instead of using an iframe, add a front matter tag to the top of the markdown file. For example: 

```
youtube_id: mJDCL0uzIpY
```

osrecio:
Hi @jeff-matthews Thanks for Suggestion. If I add to top of markdown file. How can I add a Title of video or similar

jeff-matthews:
Good question. After reviewing your PR more closely, the front matter tag may not be the best solution since this isn't a standalone topic about keeping your repo up to date. I also don't think it's possible to specify a location for a video on the page using a front matter tag.

@shrielenee, please verify that the embedded video via iframe (original submission) looks good and merge.

shrielenee:
Hey @osrecio - the embedded video looks good. @jeff-matthews is right: using the front matter places the video at the very top of the page, which doesn't allow for a specifying a location. Thanks for adding this helpful tidbit!

PR #1698, opened 2018-02-08
Small changes

yyevgenii:
First commit moves the GENERATED_CODE_SYMLINK environment variable from a list of build variables to a list of deploy variables, because it's used on the deploy step, not on build.

Second commit copies editing made by @jeff-matthews from the main Fastly article into [go-live-checklist.html(http://devdocs.magento.com/guides/v2.2/cloud/live/go-live-checklist.html)


jeff-matthews:
Thanks @yyevgenii! Is the `GENERATED_CODE_SYMLINK` environment variable available in 2.2 as well or 2.1 only?

yyevgenii:
@jeff-matthews the `GENERATED_CODE_SYMLINK` is available only in 2.0.x and 2.1.x. The new 2.2.x doesn't support it.

yyevgenii:
@jeff-matthews Thank you!

PR #1699, opened 2018-02-08
Proposed partial cleanup to Layout overview-Introduction

bdenham:
I tightened up the first part of the introduction of the Layout overview section to introduce and focus on the key concepts: layout, containers, and blocks. I also updated the container and block images by consolidating them into one numbered image that illustrates the three concepts side by side and details those concepts by reference those numbers below the image. I also removed the MVC information from the beginning as it was a bit out of place and distracted from the explanation of the main topics.

Reorganization of the other topics within this overview is also needed, but not made in this commit.


jeff-matthews:
Thanks @bdenham! We might start asking all candidates to open PRs as part of the recruiting process.

bdenham:
Yeah, well, you just remember who did it first! ;-)

jeff-matthews:
@bdenham, can you add these updates to the `/Users/jmatthews/git/devdocs_internal/guides/v2.0/frontend-dev-guide/layouts/layout-overview.md` file as well?

bdenham:
Done. Let me know if you need me to do anything else. Next step would be to tackle the rest of the layout section by providing various XML file examples (layout handles included) to help cement the concept with the practical.

Issue #908, opened 2016-11-22
webapi_rest event area isn't mentioned in the Events and observers page


hostep:
Hi

I just wasted 2 hours trying to figure out why a certain event wasn't being observed on the frontend while defining it in my `etc/frontend/events.xml` file.
Apparently, certain 'frontend' events now should be declared in the `etc/webapi_rest/events.xml` file.
There is no mentioning of this on the [Events and observers page](http://devdocs.magento.com/guides/v2.1/extension-dev-guide/events-and-observers.html)

Is it possible to add this and also explain which events should be declared in that event area?
So other people waste less time then myself when they have to get some work done :)

Thanks!

References from the Magento2 project: https://github.com/magento/magento2/issues/7524 & https://github.com/magento/magento2/issues/1581

tanberry:
Ach, sorry @hostep  for the wasted time, and a big thanks for letting us know how to prevent others from wasting time in the future. @jcalcaben can you take a look and add the requested info to the  Events and observers page and wherever else it might be needed? Pieter, it would be great if you are willing to take a look after James updates the docs, and let us know if it is more clear. Thanks!

hostep:
Tnx @tanberry!

@jcalcaben, I see there are even more area's to be aware of, see:
https://github.com/magento/magento2/blob/develop/lib/internal/Magento/Framework/App/Area.php#L18-L24

But I'm not sure if all of these can and are being used for defining observers...

jcalcaben:
@hostep Thanks for pointing this out. I will also add **Areas** as a topic that needs to be written in addition to fixing the doc.

hostep:
Hi @jeff-matthews, actually PR https://github.com/magento/devdocs/pull/1700 only talks about the `frontend` and `adminhtml` area's, not about the `webapi_rest` area or any of the [other area's](https://github.com/magento/magento2/blob/2.3-develop/lib/internal/Magento/Framework/App/Area.php#L18-L24).

PR #1700, opened 2018-02-09
Update observers-bp.md

luckyduck:
Make sure the adminhtml option is listed as well. Already described somewhere else, but showing all options on the point of use avoids further research.


jeff-matthews:
Thanks for the contribution @luckyduck!

hostep:
Guys, don't forget that there are other area's as well which are being used on the frontend, see this issue which still hasn't been addressed yet: https://github.com/magento/devdocs/issues/908 ;)

jeff-matthews:
Thanks for the reminder @hostep. We're doing our best to address issues. Sometimes, a PR is easier for us to process given our current time and resource constraints.

PR #1702, opened 2018-02-12
Run grunt command before ece-build hook

yyevgenii:
The grunt tasks prepare Magento themes and they should be run before setup:static-content:deploy command. I've added a note that grunt should be placed before our build hook.


yyevgenii:
@jeff-matthews sure, it's fine for me

PR #1712, opened 2018-02-17
change .xml to di.xml

chandranm:
`<your module root dir>/etc/frontend/di.xml filename is not specified.


keharper:
Good catch! Thanks.

PR #1713, opened 2018-02-19
Fixed default number of jobs

buskamuza:
Described default number of jobs, "4", is incorrect. It is "0" for the command. Though default number is "4" for `\Magento\Deploy\Process\Queue` which is used by the command.


keharper:
Thanks for the correction.

PR #1723, opened 2018-02-23
Update component-root.md

JosephMaxwell:
There doesn't seem to be a good page discussing `etc/module.xml`. I'm not sure there is much to say about it other than at least pointing people to details about each piece of configuration in there. Having a sample file would be helpful.


jeff-matthews:
Thanks @JosephMaxwell!

JosephMaxwell:
My pleasure. You all are doing a great job with the DevDocs!

jeff-matthews:
Thanks for the words of encouragement! That means a lot to us!

PR #1731, opened 2018-02-27
add informormation about MariaDB to MySQL Page

jonashrem:
added 

Magento is also compatible with MySQL NDB Cluster 7.4.*, MariaDB 10.0, 10.1, 10.2, Percona 5.7 and other binary compatible MySQL technologies.

from http://devdocs.magento.com/guides/v2.2/install-gde/system-requirements-tech.html


jeff-matthews:
Thanks @jonashrem!

rsidhaarth:
@jonashrem @jeff-matthews 

NDB cluster doest support FULLTEXT Index, then how Magento 2 is compatible with NDB Cluster?

I also created an issue for the same.

PR #1733, opened 2018-02-28
Replace na region with us

yyevgenii:
Replace na region with us and remove mention about plans to support other regions (these regions are already available).


hguthrie:
Would the SSH line in this topic also need replacing `na` ? : ../guides/v2.0/cloud/project/project-conf-files_services-rabbit.html#cloud-rabbitmq-conn-loc

yyevgenii:
@hguthrie I've added changes into  ../guides/v2.0/cloud/project/project-conf-files_services-rabbit.html#cloud-rabbitmq-conn-loc as well. Could you please check? Thank you.

yyevgenii:
Thank you :)

PR #1739, opened 2018-03-05
presentation markup used in acceptable example

iainhubbard:
"You must use semantic HTML markup only, and must not use presentation markup." 

The acceptable example uses presentation markup `<p>semantically</p>`. Replaced with `<strong>`. Removed `<strong>` in the unacceptable example to make it clear what is, and is not acceptable.


jeff-matthews:
Thanks @iainhubbard!

PR #1748, opened 2018-03-11
Update gulp-sass.md

zanilee:
Gulp is a task runner and SASS is a preprocessor. Fixed some inconsistent terms


shrielenee:
Thanks for facilitating this change @zanilee!

PR #1750, opened 2018-03-12
Update package_module.md

ryantfowler:
Toran is a service which is being phased out. The creators of Packagist.org have now created another service called "Private Packagist" which offers all the features of Toran Proxy, plus more. To date, the majority of account holders are Magento users. This service is the successor to Toran.


jcalcaben:
Thanks for the update @ryantfowler !

PR #1754, opened 2018-03-15
Remove minor version

mzeis:
Otherwise, this would have to be updated for every minor release (at the moment it says 2.0 where the current version is 2.2).


jeff-matthews:
Thanks @mzeis!

PR #1755, opened 2018-03-15
Remove https:// from hosts file entry

mzeis:
To my knowledge, protocols cannot be specified in the `hosts` file.


jeff-matthews:
Thanks @mzeis!

PR #1758, opened 2018-03-20
Add Customer EAV attribute example including `used_in_forms` parameter.

lbajsarowicz:
*Magento 2.1* marked several commonly used methods (including `$attribute->save()` and `$attribute->getResource()`) deprecated. As a result, lots of commonly used resources (StackOverflow / Tutorials) spread the deprecated way of persisting attributes like `used_in_forms`.

To avoid these misleading information, I've created extra EAV page section focused on creating new attribute for Customer.

After my changes it looks like the one attached:
zrzut ekranu z 2018-03-21 00-20-55

### DIFF
--- a/guides/v2.0/extension-dev-guide/attributes.md
+++ b/guides/v2.0/extension-dev-guide/attributes.md
@@ -44,54 +44,6 @@ In this case, when `getCustomAttributes()` is called, the system returns only cu
 
 The `Customer` module does not treat its EAV attributes in a special manner. As a result, the `getCustomAttributes()` method returns all EAV attributes.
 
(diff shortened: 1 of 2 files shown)


jeff-matthews:
Thanks for the contribution @lbajsarowicz! This is great! We'll start processing this and let you know if we have any questions or feedback.

lbajsarowicz:
How to make change only to 2.1+? I see that's symlinked one to another, but when someone update 2.0 does it mean that someone has to update 2.1 also?

jeff-matthews:
That was the right way to do it @lbajsarowicz, thanks! We just need to do a little light copy-editing, then I'll merge this.

PR #1759, opened 2018-03-21
Fix directory name

Vinai:


keharper:
:thumbsup: Thanks for the correction, @Vinai

PR #1770, opened 2018-03-24
Update package_module.md

ryantfowler:
Correcting a previous misspelling of "Private Packagist".


jeff-matthews:
Thansk @ryantfowler!

PR #1771, opened 2018-03-24
Update package_module.md

ryantfowler:
Included a link to the documentation which does include more steps than simply adding in a repositories entry for private packagist, such as actually running a composer command to configure your local composer with credentials (which are mandatory) to access your private packagist account.

whatsnew
Added a link to the Private Packagist documentation in the [Package a component](http://devdocs.magento.com/guides/v2.2/extension-dev-guide/package/package_module.html) topic.


jeff-matthews:
Thanks @ryantfowler!

PR #1776, opened 2018-03-28
Improved clarity of MAJOR and/or MINOR dependency

josephmcdermott:
`MAJOR and/or MINOR` suggests that a dependency can be added on the MINOR alone, re-worded this to `MAJOR and optionally MINOR`.


jcalcaben:
I agree. This is a lot more clear. Thanks @josephmcdermott !

PR #1777, opened 2018-03-28
Incorrect word "Customer" used instead of "Factory"

chris-pook:


jcalcaben:
Thanks for the fix @chris-pook !

PR #1778, opened 2018-03-28
incorrect file name for generated global file

chris-pook:


jcalcaben:
Thanks for the fix @chris-pook !

PR #1779, opened 2018-03-28
Adding the missing semi-colons

kalpmehta:


jcalcaben:
Good catch. Thanks @kalpmehta !

kalpmehta:
Thanks for merging it @jcalcaben !

PR #1780, opened 2018-03-28
Updated to correct filesystem paths for commerce and opensource

markoshust:


markoshust:
Ah... it appears I was looking at the 2.0 branch. all ok in the 2.2/2.3 branch http://devdocs.magento.com/guides/v2.3/migration/migration-tool-configure.html

PR #1793, opened 2018-04-03
Update example of requirejs-config.js file

mrodespin:
There was an error on the example, one missing } for close the config


bdenham:
Good catch, @mrodespin. Thanks for the update!

PR #1808, opened 2018-04-13
Remove incorrect information related to dig command

yyevgenii:
I've removed the dig commands which don't work. There's no need to fix this sections, we can just drop them.

I think the following articles could be merged into one, but I'm not sure that it doesn't break anything if I do it:
* guides/v2.0/cloud/trouble/trouble_fastly.md
* guides/v2.1/cloud/trouble/trouble_fastly.md


jeff-matthews:
Thanks @yyevgenii! Re: merging these files. We broke symlinks between 2.0 and 2.1 files because Magento 2.0 has reached end of life. In most cases, that means we no longer update 2.0 docs.

PR #1811, opened 2018-04-17
Update mage-for-dev-4.html

classaxe:
A couple of typos and file name for app/design/frontend/base/default/template/page/1column.phtml was incorrect.


jeff-matthews:
Thanks @classaxe!

PR #1833, opened 2018-04-24
SCD dump command fix

NickKost:
Replaced php bin/magento magento-cloud by php vendor/bin/m2-ece-scd-dump for 2.2.


hguthrie:
Thank you for catching that.

PR #1834, opened 2018-04-25
Update README.md

ryantfowler:
### Description
- I've rewritten the section for building the Magento DevDocs using Docker so that it will be: clearer to the reader, accurate, and comprehensive.
- The existing instructions are not helpful at this time, as there is simply a link to another repository which does not contain clear instructions on using Docker to build out the Magento DevDocs locally.


jcalcaben:
Awesome! Thanks @ryantfowler !

PR #1836, opened 2018-04-25
Info block adjustment

zanilee:
Corrected broken markup


keharper:
Thanks again. The 2.3 version of the file is symlinked, so you won't have to do this three times. :)

PR #1838, opened 2018-04-26
Update indexing-custom.md

chris-pook:
A `sales_order_flat` table doesn't exist in M2, renamed to `sales_order` table which would be used in this context.


keharper:
Thanks for noticing and correcting these details, @chris-pook . I verified that the `sales_order_flat` table doesn't exist and that `sales_order.entity` does.

PR #1839, opened 2018-04-26
Update lifecycle.md

chris-pook:
The `DataUpgrade::upgrade()` function is called only when the `data_version` is lower than the module version, it is not called when the `schema_version` is less than the module version.


jcalcaben:
Good catch! Thanks @chris-pook !

PR #1841, opened 2018-04-27
Update attributes.md

chris-pook:
Changed code example to use the correct factory DI. As per this very page:

> You should not directly inject `EavSetup` in extension code.


jcalcaben:
Thanks @chris-pook !

PR #1850, opened 2018-04-30
Update view.xml link to point to exact location

erikhansen:


bdenham:
Hey @erikhansen, it looks like the link you mention is supposed to be to the luma theme view.xml as a whole because the code snippet directly under it shows the exclude node within the file. But having said that, linking directly to the exclude node of Magento's version-specific luma view.xml is probably more accurate, than the code shown in the snippet as shown here: 

image

I'm inclined to remove the snippet code in favor of your direct links to the code base theme view. Is that what you're getting at here?

erikhansen:
@bdenham My goal with this PR was to make the link take the user directly to the section of the `view.xml` file that showed the `<exclude>` code, as you noted. I think that having the `<exclude>` code ALSO included inline in the documentation makes sense, as a user can read through the documentation without having to click on a link to see the example code. So I would recommend leaving the documentation as it is.

PR #1854, opened 2018-05-01
Remove link related to itself

springimport:


jeff-matthews:
Thanks @springimport!

PR #1862, opened 2018-05-02
Update ui_comp_linking_concept.md

Vinai:
The documentation was missing important information and the docs on `listens` where wrong.

### DIFF
--- a/guides/v2.1/ui_comp_guide/concepts/ui_comp_linking_concept.md
+++ b/guides/v2.1/ui_comp_guide/concepts/ui_comp_linking_concept.md
@@ -169,7 +169,7 @@ As a result, if the component's property is the variable for the template string
 
     '${ $.provider }:foo' 
     
-If the string would be built at runtime it would be equivalent to `this.provider + '.foo'`.
+If the string would be built at runtime it would be equivalent to `this.provider + ':foo'`.
 
 We can also build complex templates strings using this syntax, as follows:
 
--- a/guides/v2.1/ui_comp_guide/concepts/ui_comp_linking_concept.md
+++ b/guides/v2.1/ui_comp_guide/concepts/ui_comp_linking_concept.md
@@ -26,10 +26,10 @@ Linking properties are set in [UI components configuration files]({{page.baseurl
 
 ### `exports`
 
-The `exports` property is used to notify some external entity that a property has changed. `exports`'s value is an object, composed of the following:
+The `exports` property is used to copy a local value to some external entity. If the local value is an observable, the external entity will also be updated whenever the local property changes. `exports`'s value is an object, composed of the following:
 
   - `key`: name of the internal property or method that is tracked for changes.
-  - `value`: name of the property or method that receives the notification. Can use [string templates](#string_templ).
+  - `value`: name of the property or method that receives the value. Can use [string templates](#string_templ).
 
 Example of setting `exports` in a component's `.js` file:
 
@@ -41,7 +41,7 @@ Example of setting `exports` in a component's `.js` file:
 }
 {% endhighlight js%}
 
-Here `visible` is the `key`, `${ $.provider }:visibility` is the `value`. The value of the local `visible` property is assigned to the `visibility` property of the `provider` component. The latter is changed automatically if the value of `visible` changes.
+Here `visible` is the `key`, `${ $.provider }:visibility` is the `value`. The value of the local `visible` property is assigned to the `visibility` property of the `provider` component. The latter is changed automatically if the value of `visible` changes if the local `visible` property is observable (which it isn't given only the code example above).
 
 Example of setting `exports` in a component's configuration `.xml` file:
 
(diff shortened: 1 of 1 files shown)


bdenham:
Hey @Vinai, thanks for the update. We'll get it reviewed and out there soon.

guides/v2.1/ui_comp_guide/concepts/ui_comp_linking_concept.md
```diff
@@ -41,7 +42,7 @@ Example of setting `exports` in a component's `.js` file:
 }
 {% endhighlight js%}
 
-Here `visible` is the `key`, `${ $.provider }:visibility` is the `value`. The value of the local `visible` property is assigned to the `visibility` property of the `provider` component. The latter is changed automatically if the value of `visible` changes.
+Here `visible` is the `key`, `${ $.provider }:visibility` is the `value`. The value of the local `visible` property is assigned to the `visibility` property of the `provider` component. The latter is changed automatically if the value of `visible` changes if the local `visible` property is observable (which it isn't given only the code example above).
```
omiroshnichenko: >if the local `visible` property is observable (which it isn't given only the code example above)

This is not correct, you can change value of `visible` with `set` method and it will trigger export.

guides/v2.1/ui_comp_guide/concepts/ui_comp_linking_concept.md
```diff
@@ -157,15 +159,17 @@ For example of `listens` usage in Magento code see [`new_category_form.xml`, lin
...
 So if we put a variable name in `'${...}'`, it is processed into a string representation of the variable’s value.
 
 When working with UI components, we often need to use the string representation of a certain property of the {% glossarytooltip 9bcc648c-bd08-4feb-906d-1e24c4f2f422 %}UI component{% endglossarytooltip %}. To address a property of the UI component in the scope of this component, the `$.someProperty` syntax is used.
 
 As a result, if the component's property is the variable for the template string, we get notation similar to the following:
 
-    '${ $.provider }' 
+    '${ $.provider }:foo' 
+    
+If the string would be built at runtime it would be equivalent to `this.provider + '.foo'`.
```
omiroshnichenko: It would be equal to `this.provider + ':foo'`

guides/v2.1/ui_comp_guide/concepts/ui_comp_linking_concept.md
```diff
@@ -147,7 +149,7 @@ Example of using `listens` in a component's configuration `.xml` file:
 <argument name="data" xsi:type="array">
     <item name="config" xsi:type="array">
         <item name="listens" xsi:type="array">
-            <item name="visible" xsi:type="string">sample_config.sample_provider:visibility</item>
+            <item name="sample_config.sample_provider:visibility" xsi:type="string">visibilityChanged</item>
```
omiroshnichenko: I think better to show listens functionality example like "local listeners". You listen for local property and handle local method. Because, previous example it's more imports and new(your) example inverted imports.

guides/v2.1/ui_comp_guide/concepts/ui_comp_linking_concept.md
```diff
@@ -26,10 +26,11 @@ Linking properties are set in [UI components configuration files]({{page.baseurl
```
omiroshnichenko: Please, remove everywhere phrase about observable or add information about method `set` and `track` property.

guides/v2.1/ui_comp_guide/concepts/ui_comp_linking_concept.md
Vinai: It is correct based on the code sample. If one of the parent classes declares the property as observable (using default.tracks or a manual call) then it works, but that is not included in the displayed code example.

guides/v2.1/ui_comp_guide/concepts/ui_comp_linking_concept.md
Vinai: Right, fixed.

guides/v2.1/ui_comp_guide/concepts/ui_comp_linking_concept.md
Vinai: The new example listens for changes in an external component  sample_config.sample_provider` and reacts locally via the callback method `visibilityChanged`. 

That is actually more like an import and also that is the most common use case I found as a third party developer for `listens`.

The previous example was very confusing to me, as it set the property `visibility` of the external component when the local property `visible` changes. That is more like an export.

guides/v2.1/ui_comp_guide/concepts/ui_comp_linking_concept.md
Vinai: The fact that the imports, exports, links and listens only are triggered for changed values on observables is so important!
After reading the original documentation it was extremely frustrating that my code examples didn't work. With the new information I think they will be much more helpful to developers who want to learn about this.

Vinai: Oh, I don't want to add more information about how to configure `default.tracks` or `track()` or `ko.observable` here since that is another topic, I don't have the time to do that. All I want to do with this PR is fix the obvious omissions.

bdenham:
Thanks @Vinai! I will add docs for the `tracks` property and `set` method as part of the UI component doc updates.

Vinai:
Thanks!

dshevtsov:
@bdenham please add `whatsnew` to the PR description

PR #1865, opened 2018-05-03
ext-bcmath is required since Magento 2.2.4

thomas-blackbird:
The php extensions "ext-bcmath" is required now in Open Source (since 2.2.4).


keharper:
@thomas-blackbird  Thank you for making us aware of this change. I couldn't accept the change you made in `system-requirements-tech.md` because this is still a live document for all of 2.2.x. I updated the topic to make it clear that `math` is required for both versions as of 2.2.4.

The release note update is coming shortly.

PR #1867, opened 2018-05-03
Update plugins.md

likemusic:
Copy info about around plugin from https://github.com/magento/devdocs/blob/develop/guides/v2.1/ext-best-practices/extension-coding/common-programming-bp.md.

Update php-code example to show that around-method should be used only when required to prevent other plugins and original method call.


likemusic:
I think that would be fine to delete the same info from source page to prevent dissynchronization the same text block in future or split it to common included block?

jcalcaben:
Thanks @likemusic !

PR #1871, opened 2018-05-04
Update mod_depend.md

ryantfowler:
Included a link to a page which provides further details relating to defining the load order of config files in module.xml ... as the previous page content is ambiguous about how this is done, even though there is reference to changing the load order using sequence. The included link to Component load order page is more verbose and clearer as to the how/why of changing the load order of files to the uninitiated.

*Also, I'm not clear on what the "redirect_from" links at the top of this page are ... so if I'm making a mistake by not including one that refers to M2.2, please let me know what action is required to update the file correctly.. I didn't find any mention of this in the devdoc contributor documentation. Thanks

### DIFF
--- a/guides/v2.2/architecture/archi_perspectives/components/modules/mod_depend.md
+++ b/guides/v2.2/architecture/archi_perspectives/components/modules/mod_depend.md
@@ -1,1 +1,60 @@
-../../../../../v2.1/architecture/archi_perspectives/components/modules/mod_depend.md
+---
+layout: default
+group: arch-guide
+subgroup: Components
+title: Module dependencies
+menu_title: Module dependencies
+menu_order: 6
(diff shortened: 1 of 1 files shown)


jeff-matthews:
Thanks @ryantfowler! The `redirect_from` metadata is what we use to create redirects on devdocs. It's a Jekyll plugin that makes managing redirects easier if we move a file. For more info, refer to https://github.com/jekyll/jekyll-redirect-from. 

I don't think you need to include a `redirect_from` link in this instance.

@jfrontain, assigned to you because this PR affects the Architecture Guide.

ryantfowler:
@jeff-matthews understood, thank you (and thank you for the link).

ryantfowler:
@jfrontain just following up on this. Thanks

PR #1878, opened 2018-05-05
Update ui_comp_template_literals.md

ryantfowler:
Broke symlink to v2.1 and created v2.2 page instead.

### DIFF
--- a/guides/v2.2/ui_comp_guide/concepts/ui_comp_template_literals.md
+++ b/guides/v2.2/ui_comp_guide/concepts/ui_comp_template_literals.md
@@ -1,1 +1,102 @@
-../../../v2.1/ui_comp_guide/concepts/ui_comp_template_literals.md
+---
+layout: default
+group: UI_Components_guide
+subgroup: concepts
+title: Template Literals in UI Components
+menu_title: Template Literals
+menu_order: 50
+contributor_name: SwiftOtter Studios
+contributor_link: https://swiftotter.com/
+version: 2.2
+github_link: ui_comp_guide/concepts/ui_comp_template_literals.md
+---
+
+# Template Literals in Magento
+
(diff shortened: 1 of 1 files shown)


ryantfowler:
@jeff-matthews I just noticed that just about every page in this [directory](https://github.com/magento/devdocs/tree/develop/guides/v2.2/ui_comp_guide/concepts) has symlinks to the v2.1 documentation. That said, does it make sense for me to go through and create v2.2 pages (copying over the v2.1 content and updating the links to reflect v2.2)? As it doesn't make sense for me to only update one page within this directory. If it does make sense to update all the pages that have v2.1 symlinks, I can do that. Thanks

jeff-matthews:
@ryantfowler, that shouldn't be necessary. To ensure that a reference in a symlinked file goes to the correct version of a file, we use Jekyll variables in the link syntax. For example, consider the following markdown link in a topic:

`[.xml declaration files]({{page.baseurl}}/ui_comp_guide/concepts/ui_comp_xmldeclaration_concept.html)`

The `{{page.baseurl}}` variable ensures that when we build versions 2.1 and 2.2 of a topic, the reference points to version 2.1 and 2.2 of the referenced topic.

The only time we need to add an absolute path to a reference is if that reference doesn't exist in all versions. For example, if a topic only exists in 2.2, then we must use an absolute path to that topic when creating a cross reference.

bdenham:
Hey @ryantfowler, If I understand you right, you are referring to the fact that all 2.2 and 2.3 files in the directory you mentioned use symlinks back to 2.1 files, which means that either (1) The file contents for 2.2 and 2.3 have not changed since version 2.1, or (2) the contents have changed, but the documentation has not been updated yet. 

Only in the second case would it be necessary to break the symlink and create an update file for 2.2 and/or 2.3.

PR #1880, opened 2018-05-06
Update README.md

likemusic:
Run `bundle install` required before `bundle exec` command.


likemusic:
Step 7 in the README is for Mac environment. I added short instruction for Linux environment.

I tested building site on Windows 7 and its work as well. I think that it would be good to add short instruction for Windows environment.

dshevtsov:
Unix symlinks and Ruby can be very tricky on Windows and cause unexpected failures.
We recommend to use virtual environments like Docker or Vagrant on Windows.

PR #1883, opened 2018-05-07
change to correct git remote name

jonpday:


hguthrie:
Technically, `origin` is correct. The remote name is subjective. You can use `git remote -v` to see the remote reference for `origin`, or `magento`, if you have it.

PR #1894, opened 2018-05-09
Update lifecycle.md

teodorpopa:
Fixed incorrect ; placement


jcalcaben:
Good catch! Thanks @teodorpopa !

Issue #1897, opened 2018-05-10
Magento v2.1.10 Modal Widget: isn't closed an option too?


zack6849:
I was going to open a PR to document that you can set a function as the value of the `closed` option, and that, as one would hope, it's called when the modal closes, your documentation on the modal shows that it does in fact have a closed **EVENT** but not that you hook into it by setting the closed attribute to a function.

Is this behaviour documented somewhere, and all events are hooked into that way, or is closed an undocumented functionality for the Modal?

Feedback on page: /guides/v2.1/javascript-dev-guide/widgets/widget_modal.html

omiroshnichenko:
Hi, @zack6849. This is how jQuery UI widget factory method `_trigger` works. Thats not Magento special behaviour. But I agree with you that this must be present in documentation. It's would be very nice if you will open PR to devdocs. Thank you.

zack6849:
@omiroshnichenko 

I'm happy to open a PR for this, but for further clarification, can those trigger events be listened to in a different way?

I don't want to document this saying **this is how you can listen to them** if this is more of a side effect than intended functionality.

omiroshnichenko:
Okay. We have two options how to assign callback to event:
1. Listen on event
```js
var modal = $('#some-element').modal({...});
modal.on('modalclosed', function() {
//Do some action when modal closed
});
```
2. Assign callback as property when creating modal instance
```js
var modal = $('#some-element').modal({
    ...
    closed: function (event, modalInstance) {
        //Do some action when modal closed
    }
});
```

zack6849:
Haha, I actually had already searched it and made a PR for it :) 
See https://github.com/zack6849/devdocs/commit/dfcf1e8e2c6646125a6af9930b56cb12f8fd29b4

PR #1906, opened 2018-05-14
Further documentation on events for modal widget (#1897)

zack6849:
## This PR is a:
[ ] New topic
[X] Content fix or rewrite
[ ] Bug fix or improvement
 
Changes the modal widget's documentation to give further information on event handling
 
## Additional information
The documentation on modal events is lacking, and the "close" property working as an event handler was undocumented entirely, which isn't ideal.

Not sure if this should be elaborated on elsewhere in the documentation as well, maybe in the future you guys should have a master page on how to hook into events and for each component declare a list of events and link to a page explaining how to handle events on another dedicated page, so the docs on the components don't vary much in that way

### DIFF
--- a/guides/v2.1/javascript-dev-guide/widgets/widget_modal.md
+++ b/guides/v2.1/javascript-dev-guide/widgets/widget_modal.md
@@ -184,7 +184,7 @@ $('#modal_content').modal({
     &lt;option2&gt;: &lt;value2&gt;,
     ...
 });
-$( "#modal_content" ).on( "close", function() {
+$( "#modal_content" ).on( "modalclosed", function() {
   foo(bar);
 });
 </pre>
@@ -196,7 +196,7 @@ $('#modal_content').modal({
     &lt;option1&gt;: &lt;value1&gt;,
     &lt;option2&gt;: &lt;value2&gt;,
     ...
-    close: function(){
+    closed: function(){
        foo(bar);
     }
 });
--- a/guides/v2.1/javascript-dev-guide/widgets/widget_modal.md
+++ b/guides/v2.1/javascript-dev-guide/widgets/widget_modal.md
@@ -176,6 +176,32 @@ The modal widget is subscribed to the following events:
 <li><a href="#modal_opened">opened</a></li>
 </ul>
 
+These can be listened to a couple of different ways, the primary being listening to it with jQuery's [`on`](http://api.jquery.com/on/) function, or creating a property with the name of the event that you want to listen to, like so:
+
+<pre>
+$('#modal_content').modal({
+    &lt;option1&gt;: &lt;value1&gt;,
+    &lt;option2&gt;: &lt;value2&gt;,
+    ...
+});
+$( "#modal_content" ).on( "close", function() {
+  foo(bar);
+});
+</pre>
+
+Which is functionally equivalent to the following:
+
+<pre>
+$('#modal_content').modal({
+    &lt;option1&gt;: &lt;value1&gt;,
+    &lt;option2&gt;: &lt;value2&gt;,
+    ...
+    close: function(){
+       foo(bar);
+    }
+});
+</pre>
+
 <h3 id="modal_closed"><code>closed</code></h3>
 Called when the modal window is closed.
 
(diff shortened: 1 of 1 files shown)


guides/v2.1/javascript-dev-guide/widgets/widget_modal.md
```diff
@@ -176,6 +176,32 @@ The modal widget is subscribed to the following events:
...
 
+These can be listened to a couple of different ways, the primary being listening to it with jQuery's [`on`](http://api.jquery.com/on/) function, or creating a property with the name of the event that you want to listen to, like so:
+
+<pre>
+$('#modal_content').modal({
+    &lt;option1&gt;: &lt;value1&gt;,
+    &lt;option2&gt;: &lt;value2&gt;,
+    ...
+});
+$( "#modal_content" ).on( "close", function() {
```
omiroshnichenko: Event must be `modalclosed`, not simply 'closed'

zack6849: That's interesting, how do you determine how to convert the two?

For example, the documentation says the modal listens and fires the 'close' event, how does jquery convert close -> modalclosed? The property is also close and it seems to work for me.

Am i missing something?

omiroshnichenko: You can read about it here https://api.jqueryui.com/jquery.widget/.
> widgetEventPrefix: The prefix prepended to the name of events fired from this widget. For example the widgetEventPrefix of the draggable widget is "drag", therefore when a draggable is created, the name of the event fired is "dragcreate". **By default the widgetEventPrefix of a widget is its name.**

zack6849: Ah, I see, i mis-remembered what the property was called, so this documentation needs to be changed a bit, my apologies!

https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Ui/view/base/web/js/modal/modal.js#L327

Looking at this it makes sense, sorry for all the questions!

guides/v2.1/javascript-dev-guide/widgets/widget_modal.md
```diff
@@ -176,6 +176,32 @@ The modal widget is subscribed to the following events:
 <li><a href="#modal_opened">opened</a></li>
 </ul>
 
+These can be listened to a couple of different ways, the primary being listening to it with jQuery's [`on`](http://api.jquery.com/on/) function, or creating a property with the name of the event that you want to listen to, like so:
```
bdenham: "...can be listened to **in** a couple of different ways..."

guides/v2.1/javascript-dev-guide/widgets/widget_modal.md
```diff
@@ -176,6 +176,32 @@ The modal widget is subscribed to the following events:
...
+$('#modal_content').modal({
+    &lt;option1&gt;: &lt;value1&gt;,
+    &lt;option2&gt;: &lt;value2&gt;,
+    ...
+    closed: function(){
+       foo(bar);
+    }
+});
+</pre>
+
```
bdenham: Revise as follows:
You can listen to these events in two ways:
1. Use jQuery's [`on`](http://api.jquery.com/on/) function:
```
var modal = $('#some-element').modal({...});
    modal.on('modalclosed', function() {
    // Do some action when modal closed
});
```
2. Assign a callback as a property when creating a modal instance:
```
$('#modal_content').modal({
    ...
    closed: function (event, modalInstance) {
        // Do some action when modal closed
    }
});
```

zack6849: I've updated with the requested changes, but what i'm curious about is, you're able to directly edit my pull request, and i'm not complaining, but why have me edit in and replace what i wrote with what you just wrote when you're able to just edit it to what you want it to be anyways?

bdenham:
Hey @zack6849! You're right, I could have just added what I wrote in the review, but I wanted to give you a chance to add to it either with your additional shorthand example or with something else additional. Sorry if I sounded too dictatorial in specifying _exactly_ what you should change. All that said, thanks for adding the extra documentation. It definitely helps!

PR #1923, opened 2018-05-18
Database and media dump transferring update

NickKost:
Cloud customers can no longer transfer dumps from one remote environment to another, only via local machine. 

Also, added -h <database.host> for zcat command as required.

## This PR is a:
[ ] New topic
[ ] Content fix or rewrite
[ ] Bug fix or improvement
 

## Summary
 
### DIFF
--- a/guides/v2.0/cloud/live/stage-prod-migrate.md
+++ b/guides/v2.0/cloud/live/stage-prod-migrate.md
@@ -117,15 +117,13 @@ For additional options, see the [rsync man page](http://linux.die.net/man/1/rsyn
 
 To migrate static files:
 
-1.	Open an SSH connection to your Staging or Production environment:
+1.	Copy media folder to your local environment:
 
-			*	Staging: ssh -A <project ID>_stg@<project ID>.ent.magento.cloud
-			*	Production: ssh -A <project ID>@<project ID>.ent.magento.cloud
+		rsync -azvP <environment_ssh_link@ssh.region.magento.cloud>:pub/media/ /local_machine/pub/media/ 
+
 2.	rsync the `pub/media` directory from your local Magento server to staging or production:
 
-		rsync -azvP pub/media/ <developmemt machine user name>@<development machine host or IP>:pub/media/
-
-  The IP is for the Magento Commerce VM or container you created when setting up the local.
+		rsync -azvP local_machine/pub/media/ <environment_ssh_link@ssh.region.magento.cloud>:pub/media/ 
 
 ## Migrate the database {#cloud-live-migrate-db}
 
@@ -161,26 +159,31 @@ To migrate a database:
 	For Pro Staging and Production environments, the name of the database is in the `MAGENTO_CLOUD_RELATIONSHIPS` variable (typically the same as the application name and user name):
 
 		mysqldump -h <database host> --user=<database user name> --password=<password> --single-transaction --triggers <database name> | gzip - > /tmp/database.sql.gz
+		
+4. 	Enter `exit` to terminate the SSH connection.
 
-4.	Transfer the database dump to Staging or Production with an `rsync` command:
+5.	Transfer the database dump to local machine with an `rsync` command:
 
-	*	Staging: `rsync -azvP /tmp/database.sql.gz <project ID>_stg@<project ID>.ent.magento.cloud:/tmp`
-	*	Production: `rsync -azvP /tmp/database.sql.gz <project ID>@<project ID>.ent.magento.cloud:/tmp`
-8.	Enter `exit` to terminate the SSH connection.
-9.	Open an SSH connection to the environment you want to migrate the database into:
+		rsync -azvP <environment_ssh_link@ssh.region.magento.cloud>:/tmp/database.sql.gz /local_folder/
+		
+6. 	Transfer the database dump from local machine to remote cloud environment (staging or production):
+
+		rsync  -azvP /local_folder/database.sql.gz <environment_ssh_link@ssh.region.magento.cloud>:/tmp
+	
+7.	Open an SSH connection to the environment you want to migrate the database into:
 
 	*	Staging: `ssh -A <project ID>_stg@<project ID>.ent.magento.cloud`
 	*	Production: `ssh -A <project ID>@<project ID>.ent.magento.cloud`
 	* To SSH into the `master` branch of your Integration environment:
 
 			magento-cloud environment:ssh
-10.	Import the database dump with the following command:
+8.	Import the database dump with the following command:
 
-		zcat database.sql.gz | mysql -u <username> -p<password> <database name>
+		zcat database.sql.gz | mysql -h <database_host> -u <username> -p<password> <database name>
 
 	The following is an example using information from step 2:
 
-		zcat database.sql.gz | mysql -u user main
+		zcat database.sql.gz | mysql -h database.internal -u user main
 
 ### Troubleshooting the database migration
 If you encounter the following error, you can try to create a database dump with the DEFINER replaced:
--- a/guides/v2.0/cloud/live/stage-prod-migrate.md
+++ b/guides/v2.0/cloud/live/stage-prod-migrate.md
@@ -115,15 +115,26 @@ Options:
 
 For additional options, see the [rsync man page](http://linux.die.net/man/1/rsync){:target="\_blank"}.
 
-To migrate static files:
 
-1.	Copy media folder to your local environment:
+To migrate static files from your local machine:
 
-		rsync -azvP <environment_ssh_link@ssh.region.magento.cloud>:pub/media/ /local_machine/pub/media/ 
-
-2.	rsync the `pub/media` directory from your local Magento server to staging or production:
+	*	rsync the `pub/media` directory from your local Magento server to staging or production:
 
 		rsync -azvP local_machine/pub/media/ <environment_ssh_link@ssh.region.magento.cloud>:pub/media/ 
+						
+To migrate static files from remote-to-remote environments directly (fast approach):
+
+**Note** In order to transfer media from remote-to-remote environments directly you need to enable ssh agent forwarding, please refer to this [github guidance](https://developer.github.com/v3/guides/using-ssh-agent-forwarding/)
+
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/devdocs?pullRequest=1923) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/devdocs?pullRequest=1923) before we can accept your contribution.<br/>

dmrachkovskyi:
Hi @hguthrie , 

Please kindly try to review and merge this PR at your earliest convenience since many Customers submit support tickets stating they cannot transfer dumps from one Cloud environment to another—as referenced in the existing documentation. 

Here's the [ticket example](https://magento.zendesk.com/agent/tickets/81216). To view the ticket, please make sure you can access the Zendesk Support app from your OneLogin account.

Environment-to-environment transfer using `rsync` worked on pre-WINGS projects, but for WINGS it's no longer valid. 

Please let me know if you have any questions or need additional info.
Thank you!

dmrachkovskyi:
@NickKost , sorry for suggesting the wrong MD file to edit. 

The exact same changes should affect this topic in v2.1:
https://github.com/magento/devdocs/blob/develop/guides/v2.1/cloud/live/stage-prod-migrate.md

Please kindly update your PR when possible. Thanks!

hostep:
I don't agree with the changes to the rsync command for static files. Because that still works.
I performed that yesterday on a Cloud project where Wings is enabled. Works just fine.

However, I'm not entirely sure if the rsync command which is currently in the documentation works because we have our own version which uses slightly different params and paths, but you definitely don't have to first rsync to local and then back to the other environment. That's just a waste of time and bandwidth.

So please test this first before making those changes.

Thanks!

NickKost:
Hi Pieter,

Could you please share the rsync syntax you’re currently using for your environments?

Your ssh keys were added manually to all environments before we enabled Wings this might be reason.


Regards,

Nick
________________________________

hostep:
Ah hi Nick, didn't realize it was you :)

So, suppose you are logged in via ssh in the staging environment, and you want to copy all the pub/media files over from the production environment, we do something like this (redacted, so our project id's aren't in here):
```
rsync -e 'ssh -p 22' --exclude="/catalog/product/cache/" --delete --stats -rvth ent-abcdefghijklm-production-abcdefg@ssh.eu.magentosite.cloud:"/app/abcdefghijklm/pub/media/" "/app/abcdefghijklm_stg/pub/media/"
```

This command is build dynamically in our case, since it depends on what environment you are syncing from/to, we have a configuration file which contains all the data for each environment (ssh user, ssh domain, ssh port, magento root directory on the server, domainname, branch, ...)

You probably don't need the `-e 'ssh -p 22'` flag, but we tend to write our scripts so that should the ssh port change one day, we can just change it very easily in our config file.

All those other flags are also what we think is right, everybody should read the rsync manual and decide for themselves what they think is right, but the above flags work fine for us.

> Your ssh keys were added manually to all environments before we enabled Wings this might be reason.

Ah, I think I know what you mean, and now I start to understand it. But we have SSH agent forwarding enabled, that is probably very important, otherwise you can't authenticate to the other server from the first server.
Ok this is beginning to make sense now, you can leave it like this then. Other devs might not have SSH agent forwarding enabled. I'm ok with the changes, forgot about authentication from the one server to the other :)

NickKost:
Hey Pieter!

It’s extremely important for our team to receive reviews from our partners, I appreciate this!

True, ssh agent is very convenient approach, I think we need to test this and add to devdocs.

For the average users this might be an advanced level so I suggested rsync to local machine and then to the remote environment in order to decrease amount of support tickets with complains to the current documentation.

Thank you again and have a great weekend!

Regards,

Nick
________________________________

guides/v2.0/cloud/live/stage-prod-migrate.md
```diff
@@ -117,15 +117,13 @@ For additional options, see the [rsync man page](http://linux.die.net/man/1/rsyn
 
 To migrate static files:
 
-1.	Open an SSH connection to your Staging or Production environment:
+1.	Copy media folder to your local environment:
```
JacobBrownAustin: I think we should recommend using ssh agent forwarding instead.  As other said, it is a lot quicker, and easier.  We just need to explain how to do it (or link to third party documentation on how to do it).

guides/v2.0/cloud/live/stage-prod-migrate.md
```diff
@@ -161,26 +159,31 @@ To migrate a database:
 	For Pro Staging and Production environments, the name of the database is in the `MAGENTO_CLOUD_RELATIONSHIPS` variable (typically the same as the application name and user name):
 
 		mysqldump -h <database host> --user=<database user name> --password=<password> --single-transaction --triggers <database name> | gzip - > /tmp/database.sql.gz
```
JacobBrownAustin: We should probably recommend using `vendor/bin/ece-tools db-dump` here instead.  That way you don't have to lookup the username, databasename, hostname, and password.  Though, I did recently file a bug report on our tool because one of the options we are using fails on large database tables, so we may want to wait on changing this line of devdocs.  https://magento2.atlassian.net/browse/MAGECLOUD-2215

guides/v2.0/cloud/live/stage-prod-migrate.md
```diff
@@ -161,26 +159,31 @@ To migrate a database:
...
 	*	Staging: `ssh -A <project ID>_stg@<project ID>.ent.magento.cloud`
 	*	Production: `ssh -A <project ID>@<project ID>.ent.magento.cloud`
 	* To SSH into the `master` branch of your Integration environment:
 
 			magento-cloud environment:ssh
-10.	Import the database dump with the following command:
+8.	Import the database dump with the following command:
 
-		zcat database.sql.gz | mysql -u <username> -p<password> <database name>
+		zcat database.sql.gz | mysql -h <database_host> -u <username> -p<password> <database name>
```
JacobBrownAustin: :+1:

JacobBrownAustin:
So, @michaelcasey316 and I have recently been tasked with transferring a lot of staging environments.  We both use ssh agent forwarding because that way the transfers are super quick, and we don't have to waste a lot of time waiting for transfers that may be slow.  He uses ssh-agent in his virtual machine, while I'm using gpg-agent as my ssh-agent so that I can use my yubikey with with physical touch confirmation.  Both methods work great.  I know that some weird local operating systems environments may not support this, but I'm guessing that most do.  I have never tried it on Mac OS, but I would imagine that there is support for it.  I'm pretty sure that the popular putty client on MS Windows supports it because I feel like I've done that before.  

I'm not against also documenting how to transfer files locally and back, because that may be important for some users that have weird or restricted local machines.  I just think the first recommendation for transfering files should be the fastests one because I don't think it is that hard to set up ssh agent forwarding with either ssh-agent or gpg-agent (or pageant for MS Windows users).

NickKost:
Thanks Jacob.

We will test ssh agent and add this option to the devdocs as well as remote-local-remote option.

Regards,

Nick
________________________________

hostep:
Nice!

Just some extra tidbits I want to add:

- not everyone likes ssh agent forwarding, there is a fear that by enabling this, you ssh key can be re-used by the root user of the server your connecting to while you are connected, so you have to trust the server and its admins.

- regarding db dumps, there is also the problem with the `DEFINER` user in the dumps, if you transfer a dump from one server to another and the mysql user is different, it can cause problems. You can read more about this over here: https://github.com/platformsh/platformsh-cli/issues/643 (would be great if this could be handled automatically by the `magento-cloud.phar` utility, but for now we have to fix it manually, see below)

In our scripts, we execute something like this on our local machines when getting a database dump from a Cloud environment:
```
./tools/magento-cloud.phar db:dump -f "dbdump-develop-2018-05-19-09-29-54-906091.sql.gz" -d "./tools/dbdumps" -z -e "develop"
gzcat "./tools/dbdumps/dbdump-develop-2018-05-19-09-29-54-906091.sql.gz" | LANG=C LC_CTYPE=C LC_ALL=C sed -e 's/DEFINER[ ]*=[ ]*[^*]*\*/\*/' | gzip > "./tools/dbdumps/dbdump-develop-2018-05-19-09-29-54-906091.sql.gz.tmp"
mv "./tools/dbdumps/dbdump-develop-2018-05-19-09-29-54-906091.sql.gz.tmp" "./tools/dbdumps/dbdump-develop-2018-05-19-09-29-54-906091.sql.gz"
```

We're using `gzcat` because the default `zcat` from macOS does not work properly and is not compatible with GNU zcat, but there are probably solutions which are cross platform, just googling for a couple of seconds already gave me this possible solution: https://serverfault.com/a/704521

NickKost:
SSH agent forwarding is working, we can use github documentation for MacOs users - https://developer.github.com/v3/guides/using-ssh-agent-forwarding/, I asked @yyevgenii to check this and we will also add guidance for Windows users. 

zcat and db dump sections will probably remain the same until we do another test and create new pull request fo those changes to be added.

guides/v2.0/cloud/live/stage-prod-migrate.md
NickKost: vendor/bin/ece-tools db-dump might not have DEFINER replaced, this is being reviewed by @yyevgenii

JacobBrownAustin:
> "... there is a fear that by enabling this, your ssh key can be re-used by the root user of the server your connecting to while you are connected, so you have to trust the server and its admins."

We should recommend using the confirmation option of the ssh-agent.  I always tell people they should use this anyways.  On systems with X Windows (or other windowing systems), this is easy, and there are many options.  I'm not sure how doable it is on console-only UIs without something like a hardware key (my yubikey supports physical touch confirmation for example).

guides/v2.0/cloud/live/stage-prod-migrate.md
JacobBrownAustin: That is true - it does not.  We should probably add that as an option to `ece-tools db-dump` and may even want to make it as a default option because I see people running into it a lot recently.

NickKost:
@hguthrie please decline this pull request and merge new pull request, #1948, that one is for 2.1.

PR #1979, opened 2018-05-26
Removed broken image

Yogeshks:
## This PR is a:
Content fix
 

## Summary
 
## Additional information

2018 05 26-13-25-31-screenshot-devdocs


keharper:
I looked for an image with a similar name or path and couldn't find one. I also checked a snapshot of the devdocs website on archive.org, and it looks like the image was never there. Ordinarily, I'd prefer to restore the image, but in this case, I think it's OK to just delete the reference. Thanks @Yogeshks

PR #2004, opened 2018-05-30
Fixed 404 issue

gelanivishal:
## This PR is a:
- [x] Bug fix or improvement
 

## Summary
Fixed 404 issue in navigation link. Check Setup > Release Information
https://devdocs.magento.com/guides/v2.0/magento-release-information.html


jeff-matthews:
The **Setup** > **Release Information** link works, so I'm not sure how to reproduce the 404 that this Pr claims to fix. However, it does make me wonder why we're not linking to the release notes for the sake of simplicity. 

Assigning to you @jfrontain since you're probably more familiar with this area.

jfrontain:
Thanks, @gelanivishal!

PR #2013, opened 2018-05-31
Added spacing issue in command syntax.

NamrataChangani:
## This PR is a:
- [ ] Content fix or rewrite
 

## Summary
 Added spacing issue in command syntax.


shrielenee:
Hey @NamrataChangani - I will be rejecting this PR because these changes actually aren't needed. 

For the `tls-repo.md` file change - The image actually renders correctly in-line as seen in the [TLS required page](https://devdocs.magento.com/guides/v2.1/install-gde/system-requirements_repo-tls1-2.html).

For the `ulimit.md` file change - The code snippet is indented to auto-format it into code. It also renders correctly as seen in the [Set ulimit for the web server user page](https://devdocs.magento.com/guides/v2.1/comp-mgr/prereq/prereq_compman-ulimit.html).

PR #2121, opened 2018-06-06

neeta-wagento:
## This PR is a: Fix Typos
- [ ] New topic
- [ ] Content fix or rewrite
- [ ] Bug fix or improvement
 

## Summary
Fix Typos and grammer issues
 
### DIFF
--- a/guides/v2.2/coding-standards/technical-guidelines.md
+++ b/guides/v2.2/coding-standards/technical-guidelines.md
@@ -15,7 +15,7 @@ functional_areas:
 
 This document lists the fundamental coding and application design principles that guide Magento 2 developer team members.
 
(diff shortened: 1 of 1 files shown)


shrielenee:
Hi @neeta-wagento - Just a reminder to please fill out the provided template fully so we can quickly understand what changes you are proposing!

I'll be closing this PR because both of your suggested revisions are to text that is actually correct.

For the first change, "Magento core developers use this document as a reference during code reviews" is correct: devs use the doc as a reference when they are performing code reviews.

For the second change, the original "may be" is correct. "Maybe" is a different word and the meaning does not make sense here.

Thanks for taking the time to submit this!

PR #2242, opened 2018-06-13
MAGECLOUD-2150 editing locales

hguthrie:
Added the Change locale section to the Configuration management for store settings.

whatsnew
Now you can [change store locales]({{page.baseurl}}/cloud/live/sens-data-over.html#change-locales) when using SCD_ON_DEMAND without using the exporting and importing configuration process.

### DIFF
--- a/guides/v2.2/cloud/live/sens-data-over.md
+++ b/guides/v2.2/cloud/live/sens-data-over.md
@@ -198,7 +198,7 @@ If they differ, you can append content from the `config.local.php` file to your 
 You only need to complete this migration once. When you need to update the file, always update the `config.php` file.
 
 ## Change locales
-You can change your store locales without following a complex configuration import and export process, _if_ you have [SCD_ON_DEMAND]({{ page.baseurl }}/cloud/env/variables-intro.html#scd_on_demand) enabled for your Staging and Production environments. You simply update the locales using the Admin panel.
+You can change your store locales without following a complex configuration import and export process, _if_ you have [SCD_ON_DEMAND]({{ page.baseurl }}/cloud/env/variables-intro.html#scd_on_demand) enabled for your Staging and Production environments. You can update the locales using the Admin panel.
 
 {: .bs-callout .bs-callout-warning}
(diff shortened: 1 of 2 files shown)


guides/v2.2/cloud/live/sens-data-over.md
```diff
@@ -140,70 +140,88 @@ The **Pro plan** environment high-level overview of this process:
...
+2. Open the `config.php` file and delete the last line.
+3. Open the `config.local.php` file and copy the contents.
+4. Paste the contents into the `config.php` file, save, and complete adding it to Git.
 5. Deploy across your environments.
 
-You only need to complete this migration once. When you need to update the file, you will always update the new `config.php`.
+You only need to complete this migration once. When you need to update the file, always update the `config.php` file.
+
+## Change locales
+You can change your store locales without following a complex configuration import and export process, _if_ you have [SCD_ON_DEMAND]({{ page.baseurl }}/cloud/env/variables-intro.html#scd_on_demand) enabled for your Staging and Production environments. You can update the locales using the Admin panel.
```
NadiyaS: I would not mention environment types here, if SCD_ON_DEMAND is enabled (on any env) you can update locales using Admin panel.

hguthrie: Okay!, I will correct this in the release branch. Thanks Nadiia!

guides/v2.2/cloud/live/sens-data-over.md
```diff
@@ -140,70 +140,88 @@ The **Pro plan** environment high-level overview of this process:
...
+
+{: .bs-callout .bs-callout-warning}
+If you do **not** have `SCD_ON_DEMAND` enabled, then you need to enable the feature and generate a new `config.php` file. This process **overwrites** the store configuration; only do the following if the environments contain the same stores.
```
NadiyaS: It may confuse that we propose to enable SCD_ON_DEMAND and notify that need to do it only when environments contain the same stores (not clear why we mention another store). Maybe we need to say smth like, if you need to add locale to some env (e.g. to Staging or Production) but you do not want/need to enable SCD_ON_DEMAND there, you may add locale on another (Integration) env. [by enabling SCD_ON_DEMAND, adding necessary locale and update configuration file (config.php) with this new locale and share this file with target (Staging) env.] And then mention that doing this you need to be sure that these both environments should have the same stores.

PR #2330, opened 2018-06-22
Update order-intro.md

hitesh-wagento:
## This PR is a:
- [ ] New topic
- [ ] Content fix or rewrite
- [ ] Bug fix or improvement
 

## Summary
Fixed typo
 

keharper:
@hitesh-wagento The existing semicolons are more correct than the commas. I think a bulleted list would be better than the semicolons, but rewriting would require more time than I have right now. Closing.

PR #2336, opened 2018-06-22
Expand composer section

keithbentrup:
Composer is used to manage (not create) the module. If you're using composer, you'll need to make sure that the module is included in composer.json and composer update is run.

## This PR is a:
- [ ] New topic
- [x] Content fix or rewrite
- [x] Bug fix or improvement
 

## Summary
 
### DIFF
--- a/_videos/fundamentals/create-a-new-module.md
+++ b/_videos/fundamentals/create-a-new-module.md
@@ -30,7 +30,7 @@ Depending on how Magento 2 has been installed, core modules can either be locate
 
 If you build a module for a specific project, it is best to choose the app/code folder and commit to the project's repository.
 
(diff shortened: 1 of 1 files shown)


_videos/fundamentals/create-a-new-module.md
```diff
@@ -98,6 +98,13 @@ The `registration.php` is a standardized file that follows the same pattern for
 
 The only thing that varies is the module name, which in our case is `Learning_FirstUnit`.
 
+## Run the "composer update" command
+
+After committing your changes and including your new module in the "require" section of the `composer.json` file, dump the new required autoloader file with the following command:
```
jeff-matthews: @keithbentrup, what does _"dump the new required autoloader file"_ actually mean? We need to translate this into more plain language that uses less jargon.

PR #2344, opened 2018-06-23
Add reference to radio button and custom functions.

RakeshJesadiya:
Add reference to radio button tag with knockout - magento syntax and custom functions.


bdenham:
Hi @RakeshJesadiya, thanks for the PR! Great addition of the radio button. 

I'm curious about your other addition:
```
<if args="getCartParam('summary_count')"
```
Did you just want to show an example of using a function in place of a variable? Or is there something else you wanted to highlight?

RakeshJesadiya:
Yes, Its just example of a function, Nothing else.

PR #2367, opened 2018-06-26
Set  Dynamic Links

RakeshJesadiya:
Set Dynamic links and update wrong link.


hguthrie:
Thanks for fixing the missing link. I did remove the external links in effort to keep that particular list consistent.

PR #2375, opened 2018-06-27
Remove 404 script from plugins.md

convenient:
## This PR is a:
- [ ] New topic
- [ ] Content fix or rewrite
- [x] Bug fix or improvement

## Summary

I got a `File not found: https://gist.github.com/xcomSteveJohnson/c9a36d9ec887c4bbc34d.js` notification when browsing https://devdocs.magento.com/guides/v2.2/extension-dev-guide/plugins.html

This script no longer exists so should be removed

Also what was the purpose of this script? It seems highly suspicious to me, did we get some XSS code committed in? Have the devdocs been breached?!  ;)

If this is not a malicious act and it was intentional then including gist files from outside sources seem like a bad idea as they're out of control. What's the magento perspective on this?
 

## Additional information
N/A

### DIFF
--- a/guides/v2.2/extension-dev-guide/plugins.md
+++ b/guides/v2.2/extension-dev-guide/plugins.md
@@ -33,8 +33,6 @@ Plugins can not be used on following:
 
 The <code>di.xml</code> file in your {% glossarytooltip c1e4242b-1f1a-44c3-9d72-1d5b1435e142 %}module{% endglossarytooltip %} declares a plugin for a class object:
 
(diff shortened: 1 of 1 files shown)


jeff-matthews:
Hi @convenient,

Thanks for catching this. Rest assured, devdocs has not been breached. That gist was created by a former member of the documentation team a long time ago. I agree that we should not being using gists this way. In fact, @jcalcaben has been removing personal gists like this from devdocs. @jcalcaben , please remove this gist as well.

jcalcaben:
Thanks for the list @convenient !

jcalcaben:
This issue is fixed. Thanks for reporting it @convenient

PR #2363, opened 2018-06-26
Add Documentation for Asynchronous WebApi Endpoints

nuzil:
## This PR is a:
- [x] New topic

## Summary

This pull request is a first part of BULK API documentation.
Its covering Asynchronous WebApi Endpoints documentation

PR #2379, opened 2018-06-27
Add Documentation for Operation Status Endpoints

nuzil:
## This PR is a:
- [x] New topic

## Summary
 
Adding new topic with operation statuses endpoints description.
Description of 2 Endpoints included responses

### DIFF
--- a/guides/v2.3/rest/operation-status-endpoints.md
+++ b/guides/v2.3/rest/operation-status-endpoints.md
@@ -4,7 +4,6 @@ title: Status endpoints
 version: 2.3
 contributor_name: comwrap GmbH
 contributor_link: https://www.comwrap.com/
-github_link: rest/operation-status-endpoints.md
 functional_areas:
   - Integration
 ---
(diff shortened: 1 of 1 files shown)


nuzil:
@keharper can you please have a look at last line. I'm not sure how good to do it in documentation. I want to add this like a note, so developers know what is bulkUuid in endpoints. Also are there a way to create a relative links within documentation?

keharper:
We prefer to use a variable that corresponds to `https://devdocs.magento.com/guides/v2.x` then specify the rest of the path. In this case, you'd have 

[Magento Asynchronous API Requests]({{ page.baseurl }}/rest/asynchronous-web-endpoints.html)

keharper:
I missed the first part of your question. I would move the last line to line 13. We will also need to update the `asynchronous-web-endpoints` document to show a real call  where a UUID is returned.

nuzil:
"document to show a real call where a UUID is returned" - yes correct, mentioned that when wrote this part

keharper:
I'm closing this issue, since all the content is in [PR-2363](https://github.com/magento/devdocs/pull/2363/)

PR #2396, opened 2018-06-28
MAGEDOC-3040: RSS feed

jeff-matthews:
## This PR is a:
- [ ] New topic
- [ ] Content fix or rewrite
- [X] Bug fix or improvement
 

## Summary
 
When this pull request is merged, it will add an RSS feed to the devdocs site so that readers can subscribe to updates using the feed reader of their choice.
 

## Additional information

I don't think we need to worry about sorting the `<item>` nodes. Feed readers typically let users customize the sort order of articles anyway. When I previously implemented a feed on a Jekyll site, my reader would only show new topics that were published. I suggest we merge this, subscribe to the feed and test it.

To test, build the site locally, then navigate to http://localhost:4000/feed.xml in a browser. It'll take a while to load and format the file, so be patient. I've tried opening a tunnel from my machine to the internet using ngrok so I can test this using Feedly, but it's not working. Feedly doesn't recognize it, so I'm open to testing suggestions.

### DIFF
--- a/feed.xml
+++ b/feed.xml
@@ -1,0 +1,25 @@
+---
+layout: null
+---
+
+<?xml version="1.0" encoding="UTF-8"?>
+<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
+  <channel>
+    <title>{{ site.title | xml_escape }}</title>
+    <description>{{ site.description | xml_escape }}</description>
+    <link>{{ site.url }}</link>
+    <atom:link href="{{ "/feed.xml" | prepend: site.url }}" rel="self" type="application/rss+xml"/>
+    <pubDate>{{ site.time | date_to_rfc822 }}</pubDate>
+    <lastBuildDate>{{ site.time | date_to_rfc822 }}</lastBuildDate>
+    <generator>Jekyll v{{ jekyll.version }}</generator>
+    {% for page in site.html_pages %}
+      {% unless page.tag == 'omit' %}
+        <item>
+          <title>{{ page.title | xml_escape }}</title>
+          <description>{{ page.content | xml_escape }}</description>
+          <link>{{ page.url | prepend: site.url }}</link>
+        </item>
+      {% endunless %}
+    {% endfor %}
+  </channel>
+</rss>
--- a/feed.xml
+++ b/feed.xml
@@ -16,7 +16,7 @@ layout: null
       {% unless page.tag == 'omit' %}
         <item>
           <title>{{ page.title | xml_escape }}</title>
-          <description>{{ page.content | xml_escape }}</description>
+          <description>{{ page.content | xml_escape | strip_html | truncatewords: 75 }}</description>
           <link>{{ page.url | prepend: site.url }}</link>
         </item>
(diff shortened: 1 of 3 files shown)


jcalcaben:
for testing, I suggest forking the repo and creating a gh-pages branch with the built html. you'll need to remove the CNAME from the html build but you should be able to access it with the github.io URL.

jeff-matthews:
Oh yeah! Thanks for the suggestion @jcalcaben! Stand by for more info.

jeff-matthews:
I pushed this branch to my fork of devdocs: https://github.com/jeff-matthews/devdocs/tree/MAGEDOC-3040-rss-feed/

But I keep getting a 404 when I try to access to GH pages site: https://jeff-matthews.github.io/devdocs/_site/

Any recommendations?

jeff-matthews:
I had the files in a directory called `_site`. Promoting the files to root fixed the issue. Feedly still doesn't recognize the site though...

jeff-matthews:
I think I need to update the header element with this snippet so that readers recognize the site: 

`<link rel=”alternate” type=”application/rss+xml” title=”RSS” href=”url/to/rss/file” />`

jeff-matthews:
Yep, that line fixed it. Feedly now recognizes the feed coming from my personal GH pages site, but it's not showing any articles. That could be because Feedly needs time to process such a large feed. I'll give it a while and report back.

jeff-matthews:
Agh... actually, I forgot to change the `site.url` to reflect my forked GH pages domain. Fixing that now.

feed.xml
```diff
@@ -0,0 +1,25 @@
...
+    <link>{{ site.url }}</link>
+    <atom:link href="{{ "/feed.xml" | prepend: site.url }}" rel="self" type="application/rss+xml"/>
+    <pubDate>{{ site.time | date_to_rfc822 }}</pubDate>
+    <lastBuildDate>{{ site.time | date_to_rfc822 }}</lastBuildDate>
+    <generator>Jekyll v{{ jekyll.version }}</generator>
+    {% for page in site.html_pages %}
+      {% unless page.tag == 'omit' %}
+        <item>
+          <title>{{ page.title | xml_escape }}</title>
+          <description>{{ page.content | xml_escape }}</description>
```
mhauri: HI @jeff-matthews small suggestion: use `strip_html`, and `truncatewords: 75` this will improve the load time of the feed. `<description>{{ page.content | xml_escape | strip_html | truncatewords: 75 }}</description>`

jeff-matthews: Thanks @mhauri!

jeff-matthews:
The feed loads in the Feedly RSS reader:

rss

jeff-matthews:
To do:

Create a layout override to insert the RSS link in the header of each page:

> _includes/layout/header.html
```
<link rel="alternate" type="application/rss+xml" title="Magento DevDocs RSS" href="/feed.xml" />
```

jeff-matthews:
I added the RSS header link to the `_includes/layout/header-styles.html` file instead of creating an override.

Issue #2417, opened 2018-07-03
Feedback on page: /guides/v2.2/frontend-dev-guide/themes/js-bundling.html


kanduvisla:
## This issue is a:

- [ ] Wrong information in the documentation

## Description:

The documentation status that to enable JavaScript bundling, you need to have your site set in production mode and then navigate to "Stores > Configuration > Advanced > Developer". However, since Magento 2.2, the "Developer"-tab is no longer visible in the configuration.

### Steps to reproduce

Read the docs

## Expected result:

The documentation should be updated. Perhaps a note that JS bundling can only be enabled in the CLI in production mode:

    php -f bin/magento config:set dev/js/enable_js_bundling 1

bdenham:
Good catch, @kanduvisla. I will update this soon.

bhargavmehta:
@meker12  I want to take this on. 

A possible solution is to add one line in the doc representing the code for production mode or having one section for enabling in production mode.
Kindly suggest I will create a PR in a couple of hours after that.

Thank you

meker12:
@bhargavmehta We already have PR #2519 for this issue from @gelanivishal if you'd like to review and comment on those changes.

bdenham:
@kanduvisla and @bhargavmehta, I will make sure that the info from this issue gets merged with the changes in PR #2519 and will close it when that PR is merged. Thanks to both of you!

bdenham:
Fixed in PR #2519. Closing.

PR #2519, opened 2018-07-17
Improved JS bundling section for new version

gelanivishal:
## This PR is a:
- [ ] New topic
- [ ] Content fix or rewrite
- [X] Bug fix or improvement
 

## Summary
Updated Javascript bundling for new version https://github.com/magento/devdocs/issues/2417

### DIFF
--- a/guides/v2.2/frontend-dev-guide/themes/js-bundling.md
+++ b/guides/v2.2/frontend-dev-guide/themes/js-bundling.md
@@ -1,1 +1,135 @@
-../../../v2.1/frontend-dev-guide/themes/js-bundling.md
+---
+group: fedg
+subgroup: A_Themes
+title: JavaScript Bundling
+menu_title: JavaScript Bundling
+menu_order: 9000
+version: 2.2
+github_link: frontend-dev-guide/themes/js-bundling.md
+functional_areas:
+  - Frontend
+  - Theme
+---
+
+JavaScript bundling is an optimization technique you can use to reduce the number of server requests for JavaScript files.
+Bundling accomplishes this by merging multiple JavaScript files together into one file to reduce the number of page requests.
+
+## Enable JavaScript bundling
+
+Magento must be set to [production mode][production-mode] in order for bundling to work.
+
+To activate JavaScript bundling:
+
+- Enter the following command to make
+    ```bash
+    php -f bin/magento config:set dev/js/enable_js_bundling 1
+    ```
+    
+<div class="bs-callout bs-callout-info" markdown="1">
+Enable Javascript Bundling configuration is not longer available in Magento version 2.2 and next version. 
+</div>    
+
+To optimize bundling, set the following settings to `Yes`:
+
+* `Javascript Settings -> Merge JavaScript Files`
+* `Javascript Settings -> Minify Javascript Files`
+* `Static Files Settings -> Sign Static Files`
+
+## How bundling works in Magento
+
+By default, Magento bundles all JavaScript files and loads them synchronously on every page load.
+This means that each bundle is loaded one after the other and causes delays for JavaScript behaviors on a page.
+
+### Excluding files
+
+The `<exclude>` entry in a theme's `etc/view.xml` file tells Magento which files it should not bundle.
+JavaScript files excluded from bundling are loaded asynchronously by RequireJS as needed.
+
+Do not bundle JavaScript files used for testing or development because these will get loaded on every page.  
+
+The following code snippet from [Magento's Luma theme][luma-view-xml]{:target="_blank"} shows the types of files you should exclude in your theme.
+
+{% collapsible Show example %}
+
+{% highlight xml %}
+<vars module="Js_Bundle">
+    <var name="bundle_size">1MB</var>
+</vars>
+<exclude>
+    <item type="file">Lib::jquery/jquery.min.js</item>
+    <item type="file">Lib::jquery/jquery-ui-1.9.2.js</item>
+    <item type="file">Lib::jquery/jquery.ba-hashchange.min.js</item>
+    <item type="file">Lib::jquery/jquery.details.js</item>
+    <item type="file">Lib::jquery/jquery.details.min.js</item>
+    <item type="file">Lib::jquery/jquery.hoverIntent.js</item>
+    <item type="file">Lib::jquery/colorpicker/js/colorpicker.js</item>
+    <item type="file">Lib::requirejs/require.js</item>
+    <item type="file">Lib::requirejs/text.js</item>
+    <item type="file">Lib::date-format-normalizer.js</item>
+    <item type="file">Lib::legacy-build.min.js</item>
+    <item type="file">Lib::mage/captcha.js</item>
+    <item type="file">Lib::mage/dropdown_old.js</item>
+    <item type="file">Lib::mage/list.js</item>
+    <item type="file">Lib::mage/loader_old.js</item>
+    <item type="file">Lib::mage/webapi.js</item>
+    <item type="file">Lib::mage/zoom.js</item>
+    <item type="file">Lib::mage/translate-inline-vde.js</item>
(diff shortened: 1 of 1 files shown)


guides/v2.2/frontend-dev-guide/themes/js-bundling.md
```diff
@@ -1 +1,135 @@
...
+
+To activate JavaScript bundling:
+
+- Enter the following command to make
+    ```bash
+    php -f bin/magento config:set dev/js/enable_js_bundling 1
+    ```
+    
+<div class="bs-callout bs-callout-info" markdown="1">
+Enable Javascript Bundling configuration is not longer available in Magento version 2.2 and next version.
```
bassplayer7: Could we clarify that it's not available in production mode _in the admin panel_? While it is available in developer mode, it obviously doesn't do any good then as you noted.

guides/v2.2/frontend-dev-guide/themes/js-bundling.md
```diff
@@ -1 +1,135 @@
...
+
+To optimize bundling, set the following settings to `Yes`:
+
+* `Javascript Settings -> Merge JavaScript Files`
+* `Javascript Settings -> Minify Javascript Files`
+* `Static Files Settings -> Sign Static Files`
+
+## How bundling works in Magento
+
+By default, Magento bundles all JavaScript files and loads them synchronously on every page load.
```
bassplayer7: Isn't this only the case when bundling is enabled and not necessarily "by default"? By default individual modules are loaded asynchronously. I think it's **really** important to make it clear that bundling Javascript files moves most modules to be render-blocking in the head, though, so let's keep that part!

Something like:

> When bundling is enabled, Javascript files are loaded in synchronously in the head on every page load. They become render-blocking assets, unlike the default where every module is requested individually and asynchronously.

guides/v2.2/frontend-dev-guide/themes/js-bundling.md
```diff
@@ -1 +1,135 @@
...
+There are many ways to tune your theme's `etc/view.xml` file.  
+
+For example, Magento's Luma theme is configured to work well for all pages, but 
+by adding or removing items inside `exclude`, you can maximize browser performance for home, catalog, or product pages.
+
+Follow these steps to help you identify which JavaScript files to bundle for your theme:
+
+1. Create a blank page with the layouts you would like to tune.
+2. Compare the JavaScript files loaded in the pages with the JavaScript files in Magento.
+3. Use the results of that comparison to build your exclude list.
```
bassplayer7: This section seems a bit unclear to me. I'm wondering if it is re-stating some of the things mentioned earlier? Perhaps part of the issue is that it isn't clear here that files are included automatically, and only files that are manually excluded are not. I like your approach of creating a minimal page for testing!

guides/v2.2/frontend-dev-guide/themes/js-bundling.md
```diff
@@ -1 +1,135 @@
...
+    php -f bin/magento config:set dev/js/enable_js_bundling 1
+    ```
+    
+<div class="bs-callout bs-callout-info" markdown="1">
+Enable Javascript Bundling configuration is not longer available in Magento version 2.2 and next version. 
+</div>    
+
+To optimize bundling, set the following settings to `Yes`:
+
+* `Javascript Settings -> Merge JavaScript Files`
```
bassplayer7: I suggest either removing the note about merging or providing more information. The problem is that if merging AND bundling are enabled, it will merge the bundled files which negates the `bundle size`.

guides/v2.2/frontend-dev-guide/themes/js-bundling.md
```diff
@@ -1 +1,135 @@
...
+* `Javascript Settings -> Minify Javascript Files`
+* `Static Files Settings -> Sign Static Files`
+
+## How bundling works in Magento
+
+By default, Magento bundles all JavaScript files and loads them synchronously on every page load.
+This means that each bundle is loaded one after the other and causes delays for JavaScript behaviors on a page.
```
bassplayer7: It completely blocks rendering of the page, not just Javascript.

guides/v2.2/frontend-dev-guide/themes/js-bundling.md
```diff
@@ -1 +1,135 @@
...
+    
+<div class="bs-callout bs-callout-info" markdown="1">
+Enable Javascript Bundling configuration is not longer available in Magento version 2.2 and next version. 
+</div>    
+
+To optimize bundling, set the following settings to `Yes`:
+
+* `Javascript Settings -> Merge JavaScript Files`
+* `Javascript Settings -> Minify Javascript Files`
+* `Static Files Settings -> Sign Static Files`
```
bassplayer7: Sign Static Files doesn't optimize bundling, but does allow for cache busting. Perhaps that would be beneficial to clarify?

guides/v2.2/frontend-dev-guide/themes/js-bundling.md
gelanivishal: Please review it, Let me know if you have any changes.

guides/v2.2/frontend-dev-guide/themes/js-bundling.md
gelanivishal: Done.

guides/v2.2/frontend-dev-guide/themes/js-bundling.md
gelanivishal: Removed.

guides/v2.2/frontend-dev-guide/themes/js-bundling.md
hostep: Indeed, merging should be disabled, this is also mentioned over [here](https://devdocs.magento.com/guides/v2.3/performance-best-practices/configuration.html):

> We do not recommend using deprecated settings like merging JS files, as they were designed only for synchronously-loaded JS in the HEAD section of the page. Using this technique can cause bundling and requireJS logic to work incorrectly.

bassplayer7:
@gelanivishal, would you mind if I made some updates when I get a chance?

guides/v2.2/frontend-dev-guide/themes/js-bundling.md
meker12: Just a heads up that we just merged PR #3020 which updates the instructions to optimize bundling.

bhargavmehta:
Hello @gelanivishal  @bassplayer7 

I would also like to add one point. Magento dont allow to to modify `Stores -> Configuration -> Advanced -> Developer` in Production mode.
So I wished if we also review that ```Magento must be set to [production mode] in order for bundling to work.``` should be removed or not

Thank you

bassplayer7:
@bhargavmehta you have a good point. It's a bit confusing, though, because bundling does not work unless the store is in `production` mode, **but** it is not possible to enable bundling through the admin panel configuration when in production mode. As a result, the only practical way to use it is through the CLI. I think it is fair to leave but perhaps we should clarify.

I've been busy and though I'd get to making a couple edits earlier, but it hasn't happened yet obviously.

bassplayer7:
I just discovered I can't edit this. So I'll leave me review as my suggestions. Let me know if I should help with anything else here.

bdenham:
Hey @bassplayer7, thanks for jumping on this again. Do you mean you can't edit the file changes here through the GitHub UI?

bassplayer7:
@bdenham, yes, but more importantly: it didn't seem like I could edit without opening a different pull request.

bdenham:
@bassplayer7, I just made updates to the content of this PR, which includes a fix to the modified symlink into a real file. I'm hoping that will allow you to directly edit the file here in GH using the pencil icon.

bassplayer7:
Thanks @bdenham. It still doesn't seem to be letting me on there. Perhaps the best would be to merge as-is and then I can create another pull request?

PR #2564, opened 2018-07-20
Update ui-htmlcontent.md

likemusic:
> Processing and rendering is executed on the server side .

It's not completely true. Because content for component rendered on server-side and stored in html attribute. Therefor content is not rendered when page loaded and component not completely instantiated and rendered. It means that component is finally rendered on client side

## This PR is a:

- [ ] New topic
- [ ] Content update
- [x] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

When this pull request is merged, it will fix inaccuracy.

### DIFF
--- a/guides/v2.1/ui_comp_guide/components/ui-htmlcontent.md
+++ b/guides/v2.1/ui_comp_guide/components/ui-htmlcontent.md
@@ -9,7 +9,7 @@ github_link: ui_comp_guide/components/ui-htmlcontent.md
 
 ## Overview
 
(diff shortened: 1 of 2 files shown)


jeff-matthews:
@bdenham, this looks like a good edit, but deferring to you to make sure. It also appears to apply to 2.1, so the 2.1 file will need the same changes if you accept the edit.

bdenham:
Got it. Thanks!

omiroshnichenko:
I think before was meant that Block was rendered on server-side, but I agree that then, whole rendered block content, would be rendered on frontend side. Would be better to me to include both(server and frontend) rendering.

PR #2614, opened 2018-07-25
Fixed incorrect reference to ece-tools db:dump command

meker12:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [x] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

Fixed an incorrect reference to the ece-tools `db:dump` command in the _Magento Cloud Release Notes_.


hguthrie:
This might need a whats new if it is labeled Technical. Also, the Cleanup and Editorial shouldn't be used at the same time, and if it is Technical, I think it needs only the versions affected. I don't know. Too many conflicting label rules here. Check the labeling specs on the wiki.

PR #2683, opened 2018-08-06
Add a prompt to checkout the correct repo branch

ghostal:
When following the existing instructions in these docs, it isn't clear that there are different branches are intended for different installed versions.

## This PR is a:

- [ ] New topic
- [ ] Content update
- [x] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

Add a step to the instructions, prompting the user to select the correct branch of the sample data repository.

## Additional information

I'm not sure if what I've written here is exactly correct, in terms of which branch/tag you should use. The `develop` and `master` branches are talked about [here](https://github.com/magento/magento2-sample-data/tree/2.2.5#deploy-sample-data-from-github-repository) in the repo's README though, so I think I'm on the right track - but there's no mention of specific versions. I noticed that there is a tag for `2.2.5` and a branch for `2.2.5-develop`. I'm unsure which is more appropriate to use in a given situation. Happy to make additions/amendments if what I've written is incomplete/not got it quite right - I'd like to understand too! :smile:

### DIFF
--- a/_includes/install/sampledata/sample-data-clone.md
+++ b/_includes/install/sampledata/sample-data-clone.md
@@ -54,6 +54,17 @@ To clone the Magento sample data GitHub repository using the SSH protocol:
 Permission denied (publickey).
 fatal: The remote end hung up unexpectedly</pre>
     </div>
+7.  Ensure you checkout the branch of the sample data repository that corresponds with the branch you used from the main `magento2` repository.
+    
+    For example:
+    
+    If you used the `2.2-develop` branch of the Magento 2 repository, the Sample Data branch should be `2.2-develop`.
+    
+    If you used the `2.2.5` branch of the Magento 2 repository, the Sample Data branch should be `2.2.5`.
+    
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/devdocs?pullRequest=2683) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/devdocs?pullRequest=2683) before we can accept your contribution.<br/>

osrecio:
Hello @ghostal ,

You should specify that the version used should be the same as your branch cloned in Magento Repo.

For example:

If you cloned Magento 2 in 2.2-develop. The Sample Data should be 2.2-develop.
If you cloned Magento 2 in 2.2.5. The Sample Data should be 2.2.5

ghostal:
Amended the commit according to your request. Reworded it a little, just so as to avoid conflating the different git terms `clone` and `checkout`.

osrecio:
Thanks @ghostal . Now is correct. Seems very clear for me.

Issue #2664, opened 2018-08-01
Feedback on page: /guides/v2.2/install-gde/system-requirements-tls-parent.html


skotasthane:
Should this page be empty? or removed?

jeff-matthews:
@skotasthane, thanks for raising this issue. We definitely shouldn't be publishing blank pages. I think the page should redirect to https://devdocs.magento.com/guides/v2.2/install-gde/system-requirements_repo-tls1-2.html.

shrielenee:
Hey @skotasthane - I'm working to fix this up! Can you tell me how you came across this blank page? Was it from a site search, Google search, or from the nav? Thanks!

shrielenee:
I've added a redirect to v2.0 and v2.1 (which 2.2 and 2.3 are symlinked to) to direct visitors to the correct, non-blank page: /install-gde/system-requirements_repo-tls1-2.html.

Because of symlinking, the v2.2 and v2.3 versions redirect to the v2.3 version of https://devdocs.magento.com/guides/v2.2/install-gde/system-requirements_repo-tls1-2.html. We are making the decision not to break the symlinks because the content in the page is the same across all versions currently. Breaking the symlinks would require any future content updates to be made in all applicable versions. 

I've also ensured there are no links to the blank page within the content of devdocs pages. That's why I'm curious where you found the link to this blank one!

skotasthane:
I was searching for 'tls' in magento dev docs. so, you can say 'site search' is how I came across this

jeff-matthews:
As soon as you delete the empty file Shrie, we'll reindex the site so that it doesn't show up in our site search. We can do that Monday.

shrielenee:
Thanks for bringing this to our attention @skotasthane. I have an open PR to remedy this. 🥇

PR #2688, opened 2018-08-06
GH-2664 tls redirect

shrielenee:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [x] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

When this pull request is merged, it will add redirects from all versions of `/install-gde/system-requirements-tls-parent.html` to `/install-gde/system-requirements_repo-tls1-2.html`. It will also remove the blank files (`/install-gde/system-requirements-tls-parent.html`) for all versions.

This fixes GH issue #2664


guides/v2.1/install-gde/system-requirements_repo-tls1-2.md
```diff
@@ -9,6 +9,10 @@ level3_menu_node: level3child
 level3_subgroup: tls
 version: 2.1
 github_link: install-gde/system-requirements_repo-tls1-2.md
+redirect_from:
```
jeff-matthews: Adding redirects for multiple versions won't work for symlinked files (e.g., 2.2 and 2.3). If you want to redirect, you must break those symlinks and create unique files for 2.2 and 2.3. Since this is a redirect from an empty file, I don't think it's really necessary that we do that. I wouldn't want this example to cause confusion for others going forward however, so if you aren't going to create proper redirects, please remove them.

jeff-matthews:
Site reindexed; topic not appearing in site search

PR #2712, opened 2018-08-09
Update Cloud plans page

hguthrie:
- Removed parenthesis from Cloud variable
- BI Essentials added to Starter
- AWS VPN section will be removed


hguthrie:
There's still some debate whether we need this topic at all, but I don't want to remove it until there is something comparable in place.

PR #2748, opened 2018-08-15
replace pro with essentials

hguthrie:
BUG fix per internal request


hguthrie:
I want to SCRAP this topic. But can't do it just yet.

PR #2794, opened 2018-08-23
404 > Remove abbreviation

rafaelstz:
## This PR is a:

- [ ] New topic
- [x] Content update
- [ ] Content fix or rewrite
- [x] Bug fix or improvement

## Summary

I think that it's relevant to remove this abbreviation, as we do in other internal pages.

### DIFF
--- a/404.md
+++ b/404.md
@@ -13,7 +13,7 @@ permalink: /404.html
 The page you are looking for could not be found or the documentation does not exist.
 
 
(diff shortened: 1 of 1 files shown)


shrielenee:
Hey @rafaelstz - thanks for bringing this to our attention. Since "repo" is a common term in not only our doc set, but many other developer doc sets, and we use it frequently, I'm going to leave this as-is. Changing it to "repository" does not make it significantly clearer, as our developer audience is familiar with the shorter term.

Keep those contributions coming! 💯

PR #2795, opened 2018-08-23
Removed recursive link

meker12:
## This PR is a:

- [x] Content fix or rewrite

## Summary

Removed `Related topic` link that linked to same page.


dshevtsov:
@meker12 please update the corresponding link definition instead of removing the link.

meker12:
The related topic link added is linking to the topic immediately preceding the topic in the TOC --since there are only 4 topics in the contents, it seems like adding unnecessary maintenance to include the "Related topic link" .

PR #2827, opened 2018-08-28
Add Url Field

osrecio:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [X] Bug fix or improvement

## Summary

In order to know quickly that url will modify the PR, the field is added in "additional information".


osrecio:
Good Point @mhauri . I will add the change

osrecio:
Hi @mhauri ,

I deleted the checkboxes from your suggestion. I think is not necessary, only with the list should be enough. What do you think?

mhauri:
@osrecio sure, the list should be enough.

PR #2853, opened 2018-09-02
Link to advanced-setup went to a markdown file instead of a html file…

hostep:
…, fixed in this commit.

## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [x] Bug fix or improvement

## Summary

When this pull request is merged, it will fix a link which pointed to a markdown file instead of a html file.

You might want to search the entire codebase, there are a couple more occurrences which are probably incorrect. [One example](https://devdocs.magento.com/guides/v2.3/rest/operation-status-search.html) (link to 'Bulk API request' is incorrect as well). Don't have the time to check them all right now.
Regex you can use to search through all source code: `\{\{\s?page.baseurl\s?\}\}.+\.md` (might flag false positives) 

## Additional information

List all affected URL's 

- https://devdocs.magento.com/guides/v2.2/performance-best-practices/hardware.html

### DIFF
--- a/guides/v2.2/performance-best-practices/hardware.md
+++ b/guides/v2.2/performance-best-practices/hardware.md
@@ -24,7 +24,7 @@ Magento 2 has differing PHP memory requirements, based on how your system is dep
 Scenarios and expected PHP memory requirements:
 * Webnode serving only storefront pages: 256 MB
 * Webnode serving admin pages with a large catalog: 1 GB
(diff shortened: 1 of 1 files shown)


keharper:
Thanks for the correction and the regex, @hostep. I'll update the other instances in a separate PR.

PR #2858, opened 2018-09-03
Fixes incorrect rendering of PHP version tables in system requirements.

hostep:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [x] Bug fix or improvement

## Summary

When this pull request is merged, it will start rendering all PHP version tables correctly again.

## Additional information

List all affected URL's 

- https://devdocs.magento.com/guides/v2.3/install-gde/prereq/php-centos-ubuntu.html
- https://devdocs.magento.com/guides/v2.2/install-gde/system-requirements-tech.html
- https://devdocs.magento.com/guides/v2.2/install-gde/prereq/php-ubuntu.html
- https://devdocs.magento.com/guides/v2.2/install-gde/prereq/php-centos.html
- https://devdocs.magento.com/guides/v2.1/install-gde/prereq/php-ubuntu.html
- https://devdocs.magento.com/guides/v2.1/install-gde/prereq/php-centos.html

As comparison, here is a table which does get rendered correctly:
https://devdocs.magento.com/guides/v2.0/install-gde/prereq/php-centos.html

The actual fix here is to make sure a new line is added before the include of the partial templates.

I've also cleaned up the partial templates themselves a bit so they are more correct (end with a `|`, and have spaces surrounding the `|` at the correct places). Even though this doesn't really fix anything, I think it's better to follow the markdown "standards".


shrielenee:
These darn tables! Thanks for catching this. 💯

PR #2902, opened 2018-09-07
Add missing images from gh-pages

jcalcaben:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [X] Bug fix or improvement

## Summary

When this pull request is merged, it will copy over missing `common/images` image files found in the `gh-pages` branch but not in the source.


jcalcaben:
The only reasonable scenario I can think of is that someone may decide an image is no longer needed in a topic and deletes it in the source. Since the publish process only creates and update files, the image remains in `gh-pages`. Some of these images are used in multiple places, so the image still resolves even though it doesn't exist in the source.

PR #3002, opened 2018-10-01
 ask_magento twitter account no longer available

jonashrem:
The twitter account ask_magento is no longer available (see https://twitter.com/ask_magento)
> This handle is no longer monitored. Find us over at magento!

So, the link should be removed

## This PR is a:

- [ ] New topic
- [x] Content update
- [ ] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

removed old twitter account from help page.

### DIFF
--- a/_includes/install/get-help.md
+++ b/_includes/install/get-help.md
@@ -4,7 +4,7 @@ In the event you need additional help, consult the following resources.
 
 |Magento edition|How to get help|
 |--- |--- |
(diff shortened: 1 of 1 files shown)


shrielenee:
I think a more appropriate change would be to change the Twitter handle to the one that is used now, @magento. I will take care of that change. Thanks for bringing this to our attention!

PR #3030, opened 2018-10-03
Updated disable-module-output.md with DB options (#2959)

### LINKED ISSUE #2959: Feedback on page: /guides/v2.2/config-guide/config/disable-module-output.html
# General issue
No references to DB tables where module could be also disabled.

## Description:
It would be great to add info that event if there is no any disabled modules in mentioned in article files (`config.xml` in Backend module and `<Magento_install_dir>/app/etc/config.php`) it's still possible that module is disabled by value in database in table `core_config_data` and it can be found byvalue like `advanced/modules_disable_output/VendoeName_ModuelName` in `path` column.

michizhou:
## This PR is a:

- [ ] New topic
- [x] Content update
- [ ] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

Add information about how module output can be disabled through database tables.

## Additional information
Added brief note about how even if there is no any disabled modules in mentioned in article files it's still possible the module could be disabled by value in a database table and how it can be found by value with advanced/modules_disable_output/VendorName_ModuleName in path column. Closes #2959 

List all affected URL's 

- https://devdocs.magento.com/guides/v2.2/config-guide/config/disable-module-output.html

### DIFF
--- a/guides/v2.2/config-guide/config/disable-module-output.md
+++ b/guides/v2.2/config-guide/config/disable-module-output.md
@@ -79,3 +79,6 @@ Archive the original `<Magento_install_dir>/app/etc/config.php` file. Then add t
 Here, the `array` beneath `modules_disable_output` contains a list of modules. A value of `1` disables output for that module.
 
 As a sample result of this configuration, customers can no longer sign up to receive newsletters.
+
+{:.bs-callout .bs-callout-info}
+Even if no disabled modules are specified in the above options, the module can be disabled by value in database in table `core_config_data`. Such a module can be found by value with `advanced/modules_disable_output/VendorName_ModuleName` in the `path` column.


jeff-matthews:
Per @keharper, we need a Magento architect to review. Here are Kevin's comments:

> I was not able to confirm this PR to my satisfaction, and I ran out of time.
>
> I can see that there's a DB column named `core_config_data.path`. There is another column in the same table called `value`. I suspect that column should have a specific value (0 or 1) when a module is disabled.
>
> My gut tells me the wording should be very different than what was proposed. Something like
>
> "A module can also be disabled in the database if the `path` column contains a value similar to >`advanced/modules_disable_output/<VendorName_ModuleName>` and the `value` column contains a value of <whatever>. Magento does not recommend editing the database directly."
>
> Because of that last sentence, I don't think I can just push this back. We need an architect to decide on this. Maybe we don't even post this DB info and it becomes a KB item.

shrielenee:
Will ping the architects tagged here to see what the status is.

keharper:
@michizhou I've had an architect and product manager review your contribution, but we don't want to encourage manually editing the database. The discussions made us realize there needs to be a better way to manage module output, so we've opened a PR to create a command for that purpose. 

https://github.com/magento/community-features/issues/96

michizhou:
@keharper I completely understand your concerns. I'm sorry my contribution didn't quite meet your expectations, but I am grateful to have made it anyways! Wish you guys the best of luck with creating a new CLI command that can better manage module output.

PR #3031, opened 2018-10-03
removed archives-php-issue.md for magento 2.3

jonashrem:
The warning box in {% include install/archives-php-issue.md %} says

> Magento 2.2 archives are compatible with PHP 7.0 only. If you’re using PHP 7.1, download Magento 2.2 with Composer instead to avoid compatibility issues during installation.

which only applies to magento 2.2.

As magento 2.3 dropped php 7.0 support it might even be confusing, so I suggest to remove this warning for magento 2.3

## This PR is a:

- [ ] New topic
- [ ] Content update
- [x] Content fix or rewrite
- [ ] Bug fix or improvement


shrielenee:
Great catch @jonashrem - thanks for this!

PR #3033, opened 2018-10-03
updated php requirements for magento 2.3

jonashrem:
this page has been symlinked to the 2.1 version which supports php 5.6 and 7.0 but for magento 2.3 php 7.1 or 7.2 are required, so I suggest using a copy of that page with update requirements

## This PR is a:

- [ ] New topic
- [ ] Content update
- [x] Content fix or rewrite
- [ ] Bug fix or improvement


jeff-matthews:
Thanks @jonashrem!

PR #3034, opened 2018-10-04
update composer.json example

jonashrem:
The example of the composer.json had php 5.5 to 7.0 es supported php versions.

As non of those versions are supported by magento 2.3 it should be updates to php 7.1 and php 7.2

## This PR is a:

- [ ] New topic
- [x] Content update
- [ ] Content fix or rewrite
- [ ] Bug fix or improvement

### DIFF
--- a/guides/v2.3/frontend-dev-guide/themes/theme-create.md
+++ b/guides/v2.3/frontend-dev-guide/themes/theme-create.md
@@ -96,7 +96,7 @@ Example of a theme `composer.json` file:
     "name": "magento/theme-frontend-luma",
     "description": "N/A",
     "require": {
-        "php": "~5.5.0|~5.6.0|~7.0.0",
+        "php": "~7.1.3|~7.2.0",
         "magento/theme-frontend-blank": "100.0.*",
         "magento/framework": "100.0.*"
     },


keharper:
The code update needs to go further than updating the PHP version (although that's the most important part). 

I would investigate linking to the app/design/frontend/Magento/luma/composer.json file directly.

bdenham:
@jonashrem I unmerged the changes in this PR because more needs to be done as @keharper mentions: It would be better overall to link to the code sample in the code repo directly at app/design/frontend/Magento/luma/composer.json. Copying and pasting the content is another solution. Let us know if you need help doing this. Thanks!

jonashrem:
@bdenham you mean that file? https://raw.githubusercontent.com/magento/magento2/2.3-develop/app/design/frontend/Magento/luma/composer.json

As jekyll only supports includes from local files ( https://jekyllrb.com/docs/includes/ ) I see that it can either be copied or included as an iframe. Do you see any other possibilities?

If not, I think copying the file would be the better solution here.

bdenham:
Yes, I think copying it would be fine. Go for it!

jonashrem:
here you go #3042

PR #3035, opened 2018-10-04
fix outdated link

jonashrem:
The pages 

*	({{page.baseurl}}/install-gde/prereq/php-ubuntu.html)
*	({{page.baseurl}}/install-gde/prereq/php-centos.html)

are no longer existing for magento 2.3

Instead there is a new page called install-gde/prereq/php-centos-ubuntu.html, so the link should be changed.

Also the php versions changed

## This PR is a:

- [ ] New topic
- [x] Content update
- [ ] Content fix or rewrite
- [ ] Bug fix or improvement

### DIFF
--- a/guides/v2.3/install-gde/prereq/install-rabbitmq.md
+++ b/guides/v2.3/install-gde/prereq/install-rabbitmq.md
@@ -135,8 +135,7 @@ After you have connected Magento and RabbitMQ, you must start the message queue 
 
 *	[Installing optional software]({{page.baseurl}}/install-gde/prereq/optional.html)
 *	[Apache]({{page.baseurl}}/install-gde/prereq/apache.html)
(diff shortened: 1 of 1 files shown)


keharper:
@jonashrem brings up a good point: two files were deleted and no redirects were put in place. We need to add the following to the frontmatter of the `php-centos-ubuntu.md` file:

```
redirect_from:
  - /guides/v2.3//install-gde/prereq/php-ubuntu.html
  - /guides/v2.3//install-gde/prereq/php-centos.html
```

@jonashrem would you be willing to create a PR to do this? If not, I'll create an issue.

meker12:
> @jonashrem brings up a good point: two files were deleted and no redirects were put in place. We need to add the following to the frontmatter of the `php-centos-ubuntu.md` file:
> 
> ```
> redirect_from:
>   - /guides/v2.3//install-gde/prereq/php-ubuntu.html
>   - /guides/v2.3//install-gde/prereq/php-centos.html
> ```

I thought about doing that, but I wasn't sure if the files existed in the version 2.3 documentation in the first place.  I should have checked.

PR #3038, opened 2018-10-04
Fix symlink issue

jeff-matthews:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [X] Bug fix or improvement

## Summary

When this pull request is merged, it will fix a symlink issue introduced by #3033 that's breaking the build.


jonashrem:
oh, did I break the symlink with the github web interface?

jeff-matthews:
@jonashrem, yes. it appears that removing a symlink using the GitHub interface doesn't remove it properly. Thanks for sharing some additional information though. We were a bit stumped about how this happened. Now we know that it has something to do with removing a symlink via the GitHub interface.

dshevtsov:
It seems that the link was edited instead of being removed and replaced by a regular file. So the type of the `change-docroot-to-pub.md` was still symlink, and the system expected to use its content as a reference.

I see that GitHub allows you to edit symlinks as a regular files, and you don't even know that it's a symlink. That is the issue in this case.
You can remove files, but you cannot add them using GH web UI.

PR #3045, opened 2018-10-05
MAGEDOC-3173: 2.3 product availability

jeff-matthews:
## This PR is a:

- [x] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

When this pull request is merged, it will provide detailed product availability info for 2.3.

## Additional information

List all affected URL's 

- https://devdocs.magento.com/availability.html


dshevtsov:
@jeff-matthews
_whatsnew_ should be added at the very end of the PR description. And it requires labels that indicate affected versions.
I'll update this.

dshevtsov:
shouldn't it be moved to the _guides/2.3_ ?

jeff-matthews:
@dshevtsov, no, it shouldn't be moved. Although the current iteration of this topic is 2.3 specific, that may not always be the case. I don't want to introduce redirects to solve that hypothetical and the original request was for a simple URL.

PR #3054, opened 2018-10-08
Updated custom layout page with block replacement examples (#1910)

### LINKED ISSUE #1910: Add example for the layout_markup_replace_elements contents
## This issue is a:
[ ] Bug on the website
[ ] New topic request
[x ] Topic clarification request
[ ] New DevDocs feature request
[ ] Other
 

## Description:
The **Replace elements** section lacks information. An example would be appreciated.

 

## Expected result:
 Add an example of replacing an element. How to remove it and add a new one to replace the old one.
see : https://devdocs.magento.com/guides/v2.2/frontend-dev-guide/layouts/xml-manage.html#layout_markup_replace_elements

## Possible solutions:
 Inform of possible options to change the class of an existing block. Reference the parent block or container and then remove the old block and add a new one with the same name...

## Additional information:
I ask this because before 2.2 we could change the class from the layout xml for a given block with the `referenceBlock` tag, but now it does not work anymore. Hence, we need to use di.xml for preference definitions, but this complicates readability because we now need to check the `layout` and the `di` file to know what class the block has. And the size of the `di.xml` file just increases for no apparent good reason.

So I would like to know, how to remove and add a new block **with the same name** as the one removed so that other blocks referencing the block for positioning still display properly.

michizhou:
## This PR is a:

- [ ] New topic
- [x] Content update
- [ ] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

Update the documentation for Magento layout customization by including an crucial section about various options to change the class of an existing block element and replace old block definitions.

## Additional information
Included comprehensive example cases of replacing existing block elements via the class reference and parent block/container reference methods. Also fixed documentation errors. Closes #1910. 

List all affected URL's 

- https://devdocs.magento.com/guides/v2.2/frontend-dev-guide/layouts/xml-manage.html#layout_markup_replace_elements


michizhou:
Hey @bdenham the reason I added the square brackets around the nodes in the documentation is due to the fact that some of the nodes of the original page documentation had this formatting, so I falsely assumed that all of the nodes in the documentation should have followed the same formatting. I deeply apologize for my mistake, and have removed these brackets from the nodes in the documentation.

guides/v2.2/frontend-dev-guide/layouts/xml-manage.md
```diff
@@ -383,7 +383,85 @@ In this file, reference the element having added the `remove` attribute:
...
+
+The following block, which is our old block, is declared in `app/design/frontend/catalog/product_view_attributes/layout/local.xml`:
+
+```xml
+<block type="catalog/product_view_attributes" name="product.attributes" as="additional" template="catalog/product/view/attributes.phtml">
+    <action method="addToParentGroup"><group>detailed_info</group></action>
+    <action method="setTitle" translate="value"><value>Additional Information</value></action>
+</block>
+```
+Let us say that we want to replace the block inside our [`<reference>`] container to replaced with the following custom block type below:
```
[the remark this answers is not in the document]
michizhou: Thanks for the advice! I have already updated the replace elements section accordingly.

michizhou:
Hey @bdenham I just updated the custom layouts page with your recommended changes. Let me know if you need me to make any more modifications. Thanks!

bdenham:
Thanks, @michizhou! I'm still reviewing and I'm also having @Vinai take a look, so look for more suggestions soon.

michizhou:
@bdenham I just replaced the two paragraphs you referenced in your last review with the simplified description text you suggested. Hope this helps!

guides/v2.2/frontend-dev-guide/layouts/xml-manage.md
```diff
@@ -383,7 +382,83 @@ In this file, reference the element having added the `remove` attribute:
...
+Second, it then adds our own custom block to the content section of the page within the `<reference>` container.
+
+As a final note, it is important that your block be defined on the backend side. If it is not defined, Magento will not understand which block you are referencing to it. This will mean that Magento will not properly render the content that is located inside your block.
+
+For instance, in order for the new block in the above case to be processed properly, Magento needs `Namespace_Modulename_Block_Groupview` to be defined at `app/code/local/Namespace/Modulename/Block/Groupview.php` (where `Namespace_Modulename` stands for your custom module name).
+
+Some block definitions do not contain an explicit name attribute, which means you cannot override the block from a layout XML file. For these blocks, you can use the class preference method to override the template.
+
+**Example 2**: override the block containing the `cart/item/default.phtml` template found in `Magento/Checkout/view/frontend/layout/checkout_cart_item_renderers.xml`.
```
bassplayer7: While this is technically possible, I don't think this is a good approach. Now, we can add functionality through [ViewModels](https://magento.stackexchange.com/questions/227725/magento-2-how-to-add-a-viewmodel-to-a-core-block) (this would be a good thing for a DevDocs topic) and we can modify functionality through [plugins (interceptors)](https://devdocs.magento.com/guides/v2.2/extension-dev-guide/plugins.html). This relegates overriding blocks to rare instances where its necessary to change something in a private/protected method, or something like that. There is a [topic that currently addresses the `<preference/>`](https://devdocs.magento.com/guides/v2.2/extension-dev-guide/build/di-xml-file.html#abstraction-implementation-mappings) and it can be used for any block.

guides/v2.2/frontend-dev-guide/layouts/xml-manage.md
```diff
@@ -383,7 +382,83 @@ In this file, reference the element having added the `remove` attribute:
...
+
+Let us say that we want to override the block containing the `cart/item/default.phtml` template found in `Magento/Checkout/view/frontend/layout/checkout_cart_item_renderers.xml`:
+
+```xml
+<block class="Magento\Checkout\Block\Cart\Item\Renderer" as="virtual" template="cart/item/default.phtml">
+```
+
+The class attribute on the block sets the scope for the template path, which means that all it takes to change the template's module scope is to change the block's class with a class preference in your `di.xml` file:
```
bassplayer7: While the default scope for the template is based on the block class, the better way to handle it is to include a namespace in the template name like the core does in the `checkout_cart_item_renderers.xml`: `Magento_Checkout::cart/item/default.phtml`. This prevents problems when overriding the template in a theme.

guides/v2.2/frontend-dev-guide/layouts/xml-manage.md
```diff
@@ -383,7 +382,83 @@ In this file, reference the element having added the `remove` attribute:
...
+You can replace this block in `layout.xml` with a custom block that uses the same name, as follows:
+
+```xml
+<block type="attributesasgroup/groupview" name="product.attributes" as="additional" template="webguys/attributesasgroup/groupview.phtml">
+    <action method="addToParentGroup"><group>detailed_info</group></action>
+    <action method="setTitle" translate="value"><value>Additional Information</value></action>
+</block>
+```
+
+Magento processes the layout updates made within the special `local.xml` file last, after applying all other layout updates from other files. This means that you can include a `layout.xml` inside your theme's layout directory to ensure that your layout updates will replace an existing block type with the new block that defines your changes. For example:
```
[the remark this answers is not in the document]
michizhou: @bassplayer7 Should I just get rid of this example altogether then? Or can I just edit this?

magento-cicd2:
Can one of the admins verify this patch?

michizhou:
@bassplayer7 So sorry for the delay in response from my end, but I just finished incorporating your recommended changes into the page contents, especially in regards to providing documentation for better block replacement methods and examples. Please let me know if more changes must be made.

bdenham:
Hey @bassplayer7, what do you think?

bassplayer7:
Good changes! I'll post a review...

guides/v2.2/frontend-dev-guide/layouts/xml-manage.md
```diff
@@ -383,7 +382,95 @@ In this file, reference the element having added the `remove` attribute:
 
 ## Replace elements {#layout_markup_replace_elements}
 
-To replace an element, [remove it] and add a new one.
+Elements can be replaced with custom blocks by one of the following methods below:
```
bassplayer7: I'm going to suggest something a little radical but think it is true to the position of Magento 2 code: remove "Replace Elements" section altogether. As far as I know it is not possible to do a true "replacement" because you would have to use a different name anyway – which, would not be truly replacing but instead removing one and adding another. I think it would be acceptable to have a _brief_ section about replacing the block class, but that is beyond the scope of layout XML anyway, because all that happens in `di.xml`. Example #1 here is still Magento 1, mostly, and I suggest removing it altogether. I haven't added any comments to Example #1, because I don't think it is salvageable.

guides/v2.2/frontend-dev-guide/layouts/xml-manage.md
```diff
@@ -383,7 +382,95 @@ In this file, reference the element having added the `remove` attribute:
...
+</block>
+```
+
+As a final note, it is important that your block be defined on the backend side. If it is not defined, Magento will not understand which block you are referencing to it. This will mean that Magento will not properly render the content that is located inside your block.
+
+For instance, in order for the new block in the above case to be processed properly, Magento needs `Namespace_Modulename_Block_Groupview` to be defined at `app/code/local/Namespace/Modulename/Block/Groupview.php` (where `Namespace_Modulename` stands for your custom module name).
+
+Some block definitions do not contain an explicit name attribute, which means you cannot override the block from a layout XML file. For these blocks, you can add functionality through ViewModels or modify functionality through plugins (interceptors).
+
+**Example 2**: add functionality with a ViewModel inserted into the `cart/item/default.phtml` template found in `Magento/Checkout/view/frontend/layout/checkout_cart_item_renderers.xml`.
```
bassplayer7: I suggest removing this from an "Example" point and creating a new section: `## Add functionality to existing elements {#layout_markup_add_to_elements}`. This section is a wonderful additional to this document!

I think this section is important enough that it could do well in its own topic.

guides/v2.2/frontend-dev-guide/layouts/xml-manage.md
```diff
@@ -383,7 +382,95 @@ In this file, reference the element having added the `remove` attribute:
...
+<body>
+    <referenceBlock name="Magento/Checkout/view/frontend/layout/checkout_cart_item_renderers.xml" template="Magento_Checkout::cart/item/default.phtml" />
+        <arguments>
+           <argument name="viewModel" xsi:type="object">Vendor\CustomModule\ViewModel\Class</argument>
+        </arguments>
+    </referenceBlock>
+</body>
+```
+
+Even though the scope for the template path is based on the block class, changing the block's class with a class preference in your `di.xml` file could lead to problems when overriding your template in a theme. To prevent this, the best way to change the template's module scope is to include a namespace in the template name, similar to how the core does in `checkout_cart_item_renderers.xml` (i.e. `Magento_Checkout::cart/item/default.phtml`).
```
bassplayer7: This seems unrelated since `template` is not necessary. Good note, and perhaps could be useful somewhere else, but seems distracting to me here.

guides/v2.2/frontend-dev-guide/layouts/xml-manage.md
```diff
@@ -383,7 +382,95 @@ In this file, reference the element having added the `remove` attribute:
...
+class Class implements \Magento\Framework\View\Element\Block\ArgumentInterface
+{
+    public function __construct()
+    {
+
+    }
+}
+```
+
+To substitute or extend the behavior of original, public methods for any class or interface, we can make use of plugins, or interceptors, which are classes that modify the behavior of public class functions by intercepting a function call and running code before, after, or around that function call.
```
bassplayer7: This should be a new section - even if it just has a bold heading.

guides/v2.2/frontend-dev-guide/layouts/xml-manage.md
```diff
@@ -383,7 +382,95 @@ In this file, reference the element having added the `remove` attribute:
...
+
+```xml
+<config>
+    <type name="{ObservedType}">
+      <plugin name="{pluginName}" type="{PluginClassName}" sortOrder="1" disabled="false" />
+    </type>
+</config>
+```
+
+You must specify the following elements: type name (a class or interface which the plugin observes), plugin name (an arbitrary plugin name that identifies a plugin that is also used to merge the configurations for the plugin), and plugin type (name of a plugin’s class or its virtual type; use the following naming convention when you specify this element: \Vendor\Module\Plugin\<ClassName>).
```
bassplayer7: This seems to be duplicating content that should be on the actual Plugin / Interceptor topic.

guides/v2.2/frontend-dev-guide/layouts/xml-manage.md
```diff
@@ -383,7 +382,95 @@ In this file, reference the element having added the `remove` attribute:
...
+
+Now we have to verify that the block actually exists even if you don't need to do anything with it. To do this we can create a skeleton block, which serves to change the scope of the old block definition:
```
bassplayer7: I don't understand how this section (last 3 paragraphs) follows from the discussion of ViewModels and Plugins?

futratladan:
https://instagram.com/futratladan02

magento-cicd2:

shrielenee:
Hey there @michizhou --- it looks like this PR potentially fixes an open issue, https://github.com/magento/devdocs/issues/1910, but there has been no revision here based on feedback given. Can you revise and resubmit for approval?

michizhou:
@shrielenee My apologies for the lackluster response on my end. I have been busy with other things for the past few weeks, but I'll be sure to revise and resubmit by the end of this week.

guides/v2.2/frontend-dev-guide/layouts/xml-manage.md
michizhou: My apologies, this section was rather incomplete when it came to using Plugins to modify the functionality of existing blocks. I have just updated the section with new examples.

michizhou:
@bassplayer7 I deeply apologize for the months-long delay and lack of revision on my end, but I just finished putting in your latest suggested changes into the layouts page. I have taken extra care in order to provide adequately detailed documentation for adding and modifying functionalities of blocks with ViewModels and Plugins along with examples. Please let me know if any other changes are needed.

guides/v2.2/frontend-dev-guide/layouts/xml-manage.md
```diff
@@ -345,56 +344,143 @@ This would make the product page look like following:
...
 
-Elements are removed using the `remove` attribute for the `<referenceBlock>` and `<referenceContainer>`. 
+Let us say that we want to add functionality to a core template with custom logic using a ViewModel in the `cart/item/default.phtml` template found in `Magento/Checkout/view/frontend/layout/checkout_cart_item_renderers.xml`:
 
-**Example**: remove the Compare Products {% glossarytooltip 31751771-8163-434b-88bc-c5f94d859fc3 %}sidebar{% endglossarytooltip %} block from all store pages. 
+```xml
+<?xml version="1.0"?>
+<page xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:View/Layout/etc/page_configuration.xsd">
+<body>
+    <referenceBlock cart.renderer="Magento/Checkout/view/frontend/layout/checkout_cart_item_renderers.xml">
```
bassplayer7: This is not valid XML: `cart.renderer="Magento/Checkout/view/frontend/layout/checkout_cart_item_renderers.xml"`. Consider replacing with: `name="checkout.cart.item.renderers.default"`

guides/v2.2/frontend-dev-guide/layouts/xml-manage.md
```diff
@@ -345,56 +344,143 @@ This would make the product page look like following:
...
-...
-        <referenceContainer name="sidebar.additional">
-            <block class="Magento\Catalog\Block\Product\Compare\Sidebar" name="catalog.compare.sidebar" template="product/compare/sidebar.phtml"/>
-        </referenceContainer>
-...
-    </body>
-</page>
+<config>
+    <type name="{ObservedType}">
+      <plugin name="{pluginName}" type="{PluginClassName}" sortOrder="1" disabled="false" />
```
bassplayer7: I suggest removing `sortOrder` and `disabled` attributes. They should not be encouraged or implied as necessary.

guides/v2.2/frontend-dev-guide/layouts/xml-manage.md
```diff
@@ -345,56 +344,143 @@ This would make the product page look like following:
...
-<page xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:View/Layout/etc/page_configuration.xsd">
-    <body>
-...
-        <referenceBlock name="catalog.compare.sidebar" remove="true" />
-...
-    </body>
-</page>
+<config>
+    <type name="Magento\Checkout\Model\Cart">
+        <plugin name="MagentoCart" type="Magento\Test\Model\Cart" sortOrder="1" />
```
bassplayer7: I propose removing this `sortOrder` as well. Further, would there be an alternate namespace we could use. In Magento 2, `Test` means something specific, and it would be nice to avoid that. What about: `Company\Sample`? Please change all subsequent uses as well.

michizhou:
@bassplayer7 I just applied the minor corrections you suggested to the layouts page. Let me know if that works for you. Thanks again for your input!

bassplayer7:
@michizhou, looks great.

bdenham:
running tests

bdenham:
Thanks @michizhou for a great addition to the docs!

PR #3057, opened 2018-10-08
Generate all navigations from _data/*yml using the same logic

belbiy:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [X] Bug fix or improvement

## Summary

When this pull request is merged, it will generate all navigations from _data/*yml using the same logic.
The logic is stored in _includes/layout/nav-item.html and used in main navigation, sidebar and footer.
`include_versions`, `exclude_versions`, and `versionless` work the same way for all the places.
Added the _data/footer-links.yml file, where footer links are stored.

## Additional information

Affects all pages.

### DIFF
--- a/_data/footer-links.yml
+++ b/_data/footer-links.yml
@@ -1,0 +1,21 @@
+- label: Become a Contributor
+  url: https://github.com/magento/devdocs/blob/master/.github/CONTRIBUTING.md
+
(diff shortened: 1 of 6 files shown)


dshevtsov:
the branch contains less broken links then master which is good

PR #3059, opened 2018-10-08
Weekly digest

dshevtsov:
## This PR is a:

- Content update

## Summary

When this pull request is merged, it will update What's new on DevDocs.

### DIFF
--- a/guides/v2.1/magento-devdocs-whatsnew.md
+++ b/guides/v2.1/magento-devdocs-whatsnew.md
@@ -9,9 +9,14 @@ We exclude from this list proofreading, spelling checks, and all minor updates.
 
 ## October 2018
 
-Description  | Versions applied to  | New or Updated | Date
--------------|--------------|----------------------|--------
-Added [Two-Factor Authentication](https://devdocs.magento.com/guides/v2.2/security/two-factor-authentication.html) and [Google reCAPTCHA](https://devdocs.magento.com/guides/v2.2/security/google-recaptcha.html) installation and troubleshooting information for [Security settings](https://devdocs.magento.com/guides/v2.2/config-guide/secy/secy.html).|2.1.x, 2.2.x|New topic|October  1
+Description                                                                                                                                                                                                                                                                                                                                                                                                                           | Versions applied to | New or Updated    | Date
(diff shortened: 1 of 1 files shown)


guides/v2.1/magento-devdocs-whatsnew.md
```diff
@@ -9,9 +9,14 @@ We exclude from this list proofreading, spelling checks, and all minor updates.
...
+Added [Multi-Repo Docs](https://devdocs.magento.com/community/resources/multi-repo-docs.html) features and overview                                                                                                                                                                                                                                                                                                                   | 2.x                 | New topic         | October  7
+MFTF 2.3.7 release:<br/>Updated the [pressKey](https://devdocs.magento.com/mftf/2.3/test/actions.html#presskey) action.<br/>Updated the [action group argument](https://devdocs.magento.com/mftf/2.3/test/action-groups.html#data-type-usage) syntax for a value based on data entity resolution.<br/>Added the [DEFAULT_TIMEZONE](https://devdocs.magento.com/mftf/2.3/configuration.html#default_timezone) configuration parameter. | 2.3.x               | Technical changes | October  5
```
meker12: Putting the updated items in a bulleted list would be more effective than line breaks.  I'm pretty sure you can do that--we have it in the [Cloud Architecture](https://devdocs.magento.com/guides/v2.2/cloud/architecture/cloud-architecture.html) article.  If bullets won't work, maybe dashes.

dshevtsov: It must be done in the corresponding PR, because the process is automated and the table contains only imports from the corresponding PRs.
The [table you mentioned](https://raw.githubusercontent.com/magento/devdocs/master/guides/v2.1/cloud/architecture/cloud-architecture.md) is in HTML. We're avoiding HTML tables for the reason of better readability and maintenance of the topic.
It is still possible to add simple `-` or `*`, though they won't be converted from Markdown to  HTML on the resulting webpage.

dshevtsov:
I have an intention to refactor this list and to separate it into files at least by year. Just don't have a gap for it in my backlog.
If you have any ideas, you're welcome to create a PR. We can discuss it then.

dshevtsov:
I don't think that we need to maintain old links in this topic

PR #3099, opened 2018-10-15
[task] introduce new mage2bloburl variable

mhauri:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [x] Bug fix or improvement

## Summary

When this pull request is merged, it will introduce a new variable `mage2bloburl` which is a link to `https://github.com/magento/magento2/blob/` by this we can replace all hardcoded `https://github.com/magento/magento2/blob/` links.

## Additional information

List all affected URLs 

- ...
- ...

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -33,6 +33,7 @@ versions:
 
 github_link: true
 feedback_link: true
+mage2bloburl: "https://github.com/magento/magento2/blob/"
 
 # 2.0.0 variables
 gdeurl: "/guides/v2.0/"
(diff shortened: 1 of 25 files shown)


dshevtsov:
@mhauri thank you for the suggested change.
As w're introducing a new variable, makes sense to replace all occasions of `https://github.com/magento/magento2/blob/` with the variable throughout devdocs topics.
Would you like to add a commit with the suggested change for all devdocs topics?

mhauri:
@dshevtsov absolutely, I'll do it in a minute.

PR #3113, opened 2018-10-16
Magento cloud magecloud 2838

hguthrie:
Edited the PR submitted by Cloud team.

whatsnew
Updated the [Launch Docker configuration](https://devdocs.magento.com/guides/v2.2/cloud/docker/docker-config.html) to include the `docker:build --php` option to specify the version of PHP compatible with your Magento Commerce version.

### DIFF
--- a/guides/v2.1/cloud/docker/docker-config.md
+++ b/guides/v2.1/cloud/docker/docker-config.md
@@ -13,19 +13,22 @@ functional_areas:
 
 The `{{site.data.var.ct}}` package v2002.0.13 or later deploys to a read-only file system in the Docker container, which mirrors the read-only file system deployed in the Production environment. You can use the `docker:build` command in the `{{site.data.var.ct}}` package to generate the Docker compose configuration and deploy {{site.data.var.ece}} in a Docker container. To specify a particular version, use the following options:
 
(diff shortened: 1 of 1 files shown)


guides/v2.1/cloud/docker/docker-config.md
```diff
@@ -30,16 +41,16 @@ This version also provides a ` docker:config:convert` command to convert PHP con
...
 
 1.  In your local environment, start the Docker configuration generator.
 
     ```bash
-    vendor/bin/ece-tools docker:build
+    ./vendor/bin/ece-tools docker:build
     ```
 
+    {: .bs-callout .bs-callout-info}
+    You can use the `--php` option to specify the version of PHP compatible with your {{site.data.var.ee}} version.
```
meker12: I think this note is fine, but I think you would only need to do this if you have more than one version of PHP installed, yes?   --   See the info on this page: https://devdocs.magento.com/guides/v2.2/install-gde/trouble/readiness/tshoot_rc_php.html

hguthrie: For edification, this allows you to select the PHP image, so it is not dependent on local version.

PR #3129, opened 2018-10-18
Cloud: some small fixes in the go live checklist document.

hostep:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [x] Bug fix or improvement

## Summary

1) change `master` to `production` branch
2) fix an external link

I'm not 100% sure about the first one, this might be different for people using an older cloud project, over there it might still be the `master` branch, but I'm not sure, maybe best double check and if this is wrong, then reword this a bit somehow.

## Additional information

List all affected URLs 

- https://devdocs.magento.com/guides/v2.2/cloud/live/go-live-checklist.html

### DIFF
--- a/guides/v2.2/cloud/live/go-live-checklist.md
+++ b/guides/v2.2/cloud/live/go-live-checklist.md
@@ -8,7 +8,7 @@ functional_areas:
 
 ## Pre-Launch Checklist {#checklist}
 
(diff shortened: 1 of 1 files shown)


hguthrie:
That's funny, my first instinct is to think about merging staging to production. I will double check this so the language is clear and get back to you.

PR #3138, opened 2018-10-19
Fixed Cloud terminal code blocks

hguthrie:
This is a minor fix to some Cloud guide code blocks.
These snippets of code represent system responses and do not need a copy command. The `no-copy` attribute for some code blocks did not have a line return, as expected. Some code blocks with the "terminal" indicator were corrected, some received a new `no-copy` attribute, and some were not touched.


dshevtsov:
I was wrong, sorry. Looks good!

PR #3149, opened 2018-10-22
MAGECLOUD-2874 Double slashes in links

hguthrie:
Addressed a BUG in the Cloud ece-tools release notes. Two links included an extra forward slash,

```
}//cloud
```

so I removed one slash from each.


jeff-matthews:
I keep getting a 404 whenever I try to review the changed files. This might be related to GH's outage this weekend?

PR #3158, opened 2018-10-23
MAGECLOUD-2395 Add health notifications to integrations

hguthrie:
Added Health notifications to the Integrations for Cloud.

- Updated the Integrations landing page to include more than Git-hosting
- Added the new Health notifications topic
- Moved Integrations further down the navigation so it appears after local development setup
- Initially, I thought this would go in the configure notifications sections, but I realized this was not the case. There are a few minor changes, such as fixing a reference to the sample env YAML file, which doesn't exist.

These changes affect:
https://devdocs.magento.com/guides/v2.2/cloud/integrations/cloud-integrations.html
https://devdocs.magento.com/guides/v2.2/cloud/integrations/health-notifications.html

whatsnew
Added [Health notifications](https://devdocs.magento.com/guides/v2.2/cloud/integrations/health-notifications.html) to [Integrations](https://devdocs.magento.com/guides/v2.2/cloud/integrations/cloud-integrations.html) for Cloud.

### DIFF
--- a/guides/v2.1/cloud/env/log-handlers.md
+++ b/guides/v2.1/cloud/env/log-handlers.md
@@ -9,7 +9,7 @@ functional_areas:
 
 You can configure logging handlers to send messages to a remote logging server. Logging handler pushes build and deploy logs to other systems, similarly to the way you can push logs to Slack and email. Simply enable a _syslog_ handler, which is ideal for logging messages related to hardware, or a Graylog Extended Log Format (GELF) handler, which is ideal for logging messages from software applications.
 
-The following example configures both of these handlers by adding the configuration to the `.magento.env.yaml` file. For the minimum logging level (`min_level`), see [Set up notifications—Log levels]({{ page.baseurl }}/cloud/env/setup-notifications.html#log-levels).
+The following example configures both of these handlers by adding the configuration to the `.magento.env.yaml` file. For the minimum logging level (`min_level`), see [Log levels](#log-levels).
 
 ```yaml
 log:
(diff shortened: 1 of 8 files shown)


guides/v2.1/cloud/env/variables-intro.md
```diff
@@ -64,7 +66,9 @@ return array(
    ...
 );
 ```
-{% include note.html type="info" content="JS bundling and JS/CSS merging do not work with SCD on demand." %}
+
+{: .bs-callout .bs-callout-info}
+JS bundling and JS/CSS merging do not work with SCD on demand.
```
mveeramneni: we might have to double check this warning

hguthrie: Yeah, I don't see it in the DIST packaged in ece-tools. Will remove.

guides/v2.1/cloud/integrations/health-notifications.md
```diff
@@ -0,0 +1,43 @@
...
+
+#### To register health notifications in a Slack channel:
+
+```bash
+magento-cloud integration:add --type health.slack --token SLACK_BOT_TOKEN --channel '#slack-channel-name'
+```
+
+## PagerDuty notifications
+
+PagerDuty is an external service that can notify on-call team members of important issues. Before you can receive health notifications in PagerDuty, you must create a [PagerDuty integration](https://v2.developer.pagerduty.com/v2/docs/integrating) that uses the Events API version 2. Use the Integration Key as known as the _routing key_ to register your integration.
```
meker12: extra "as"

guides/v2.1/cloud/env/log-handlers.md
```diff
@@ -47,3 +47,16 @@ log:
...
 ```
+
+### Log levels
+
+Log levels determine the level of detail your notification messages contain. You can choose from the following options:
+
+-   **debug**—Detailed debug information.
+-   **info**—Interesting events. For example, a user logs in, SQL logs, etc.
+-   **notice**—Normal but significant events.
+-   **warning**—Exceptional occurrences that are not errors. For example, use of deprecated APIs, poor use of an API, undesirable things that are not necessarily wrong.
```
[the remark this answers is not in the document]
hguthrie: It's so romantic; don't you think?

hguthrie:
Done; thanks for the review.

Issue #3168, opened 2018-10-24
Feedback on page: /guides/v2.2/frontend-dev-guide/layouts/xml-instructions.html


Karlasa:
# General issue

## Description:

Under Container section

Attribute | Description | Values | Required?
-- | -- | -- | --
name | A name that can be used to address the container in which this attribute is assigned. The name must be unique per generated page. | A-Z, a-z, 0-9, underscore (_), period (.), dash (-). Should start with a letter. Case-sensitive. | yes

Required section should be "No"

meker12:
Fixed in PR #3169

PR #3169, opened 2018-10-24
change container name attribute required value - fix issue #3168

Karlasa:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [x] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary
fix for issue #3168
change container name attribute requirement from Yes to No

## Additional information

List all affected URLs 

- ...
- ...

### DIFF
--- a/guides/v2.2/frontend-dev-guide/layouts/xml-instructions.md
+++ b/guides/v2.2/frontend-dev-guide/layouts/xml-instructions.md
@@ -118,7 +118,7 @@ A container renders child elements during view output generation. It can be empt
          <td><code>name</code></td>
          <td>A name that can be used to address the container in which this attribute is assigned. The name must be unique per generated page.</td>
          <td>A-Z, a-z, 0-9, underscore (_), period (.), dash (-). Should start with a letter. Case-sensitive.</td>
-         <td>yes</td>
+         <td>no</td>
       </tr>
       <tr class="odd">
          <td><code>label</code></td>


jeff-matthews:
@Karlasa, have you verified that `name` is not required for 2.2 or are you just trying to make the 2.2 file match the 2.1 file?

https://github.com/magento/devdocs/blob/master/guides/v2.1/frontend-dev-guide/layouts/xml-instructions.md

Karlasa:
@jeff-matthews  i think it changed from required to not required in the same update as block name one 
Notification text in top of block attributes 

> The class attribute is no longer required in versions 2.2.1 and above as it will default to Magento\Framework\View\Element\Template. In versions lower than 2.2.1, the class attribute is still required.

Also tested and it works well without name

jeff-matthews:
Thanks for confirming @Karlasa.

@osrecio or @mhauri, if either of you can please verify, I'll merge this PR.

Karlasa:
@mhauri  Based on the framework:View/Layout/etc/page_configuration.xsd schema it also says the same about **block** and yet it's optional. It's not about having it set or not it's about defining it. If you don't define it it will be still exist but it will be random name. 

Of course it's best practice to have it, but still it's not required ;)

lorikrell:
Definitely worth adding a note. @Karlasa I can help with the note if you like!

lorikrell:
Adding the note in another PR, will link it to this one. Thanks!

PR #3223, opened 2018-11-01
Fixing the broken "Example 2" collspsible element

zanilee:
The "Example 2" is rendered as "Example 1" list item and has "</ul> </div> </div>" tags displayed after.

## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [x] Bug fix or improvement

## Summary

## Additional information

List all affected URLs 

- https://devdocs.magento.com/guides/v2.1/javascript-dev-guide/javascript/js-resources.html
- versions 2.2 and 2.3 are symlinked

### DIFF
--- a/guides/v2.1/javascript-dev-guide/javascript/js-resources.md
+++ b/guides/v2.1/javascript-dev-guide/javascript/js-resources.md
@@ -38,8 +38,6 @@ We recommend specifying JavaScript resources in the templates rather than in the
 
 JS resources are accessed using relative paths.
 
-Examples:
-
 **Example 1**
 
 - File actual location: `app/code/Magento/ConfigurableProduct/view/frontend/web/js/configurable.js`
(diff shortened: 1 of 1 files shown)


shrielenee:
Hey @zanilee - Thanks for this! I think to be correct, you should first remove the "Examples:" text, then make all the examples not collapsible (remove the collapsible elements from Examples 2 and 3). That way, they all appear one after another and are equally visible.

Can you make this revision and push back up for review? 🏆

zanilee:
Hi, @shrielenee !

The collapsible elements were in the article before my fix, so I didn't touch them. Just did some adjustment for them not to be displayed broken. I'm not sure if we should remove them at all

shrielenee:
I totally understand @zanilee!

Since we're already editing it, however, I think we should remove the collapsible elements, etc. based on my previous response. If you can fix that up we'll get it approved and merged.

zanilee:
Hi, @shrielenee !

I corrected the PR according to your comments.

PR #3240, opened 2018-11-06
Redoc updates

belbiy:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [X] Bug fix or improvement

## Summary

1. Move all Redoc logic into `_layouts/redoc.html`.
2. Each Redoc page will have only the `specUrl` property.
3. Update Redoc to latest version
4. Add basic theming to Redoc to improve readability.

This is how it will look like: 
screen shot 2018-11-06 at 12 34 44 pm

## Additional information

List all affected URLs 

- redoc/2.3/*

### DIFF
--- a/_layouts/redoc.html
+++ b/_layouts/redoc.html
@@ -1,7 +1,35 @@
 {% include layout/header.html %}
 
 <div class="content-container">
-    {{ content }}
+    <div id="redoc-container"></div>
 </div>
+
+<script src="https://cdn.jsdelivr.net/npm/redoc@2.0.0-alpha.41/bundles/redoc.standalone.js"> </script>
+<script>
+    Redoc.init('{{ page.specUrl }}', {
+        scrollYOffset: '.site-nav',
+        hideHostname: true,
+        sortPropsAlphabetically: true,
+        pathInMiddlePanel: true,
+        hideDownloadButton: true,
+        showExtensions: true,
+        nativeScrollbars: true,
+        lazyRendering: true,
+        theme: {
+            colors: {
+                primary: {
+                    main: '#2b2b2c',
+                }
+            },
+            typography: {
+                fontFamily: '',
+                headings: {
+                    fontFamily: '',
+                }
+            },
+        }
+    }, document.getElementById('redoc-container'))
+
+</script>
 
 {% include layout/footer.html %}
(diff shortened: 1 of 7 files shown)


_layouts/redoc.html
```diff
@@ -1,7 +1,35 @@
 {% include layout/header.html %}
 
 <div class="content-container">
-    {{ content }}
+    <div id="redoc-container"></div>
 </div>
 
+<script src="https://cdn.jsdelivr.net/npm/redoc@2.0.0-alpha.41/bundles/redoc.standalone.js"> </script>
```
jeff-matthews: @jcalcaben, will need to confirm, but I don't think we can use v2 of redoc without losing sample requests/responses.

belbiy: Gotcha, I was assuming the new version supports that, but now see that's not entirely true.

jcalcaben:
This looks good to me.
Thanks for finding a solution that lets us use the most recent ReDoc version.
Adding @keharper as a reviewer as he may have some feedback too.

keharper:
I will approve this PR, but I'd like to discuss the presentation as a whole. Looking at the changes in the `_layouts/redoc.html` file, it looks like @belbiy made the main color. darker and perhaps changed the default font. My comments include things he changed and things that probably changed with the version upgrade.

This is the response pane for GET /V1/attributeMetadata/customer as it currently exists on devdocs

current-redoc

And this is the pane for the same endpoint with the changes made in the PR.

new-styling-redoc

Things I like on the existing version:
* Larger text
* White text  is distinguished from string data types

Updates that I like
* Response codes are cleaner, with green for a successful request and red for unsuccessful requests

Things I don't like in either version
* The blue used for integer data types is hard to see
* The red used for boolean data types also fades into the background
* The "headings" in the left navigation (see the screen capture at the top of this PR) don't stand out.

belbiy:
Thank you for the feedback, @keharper ! I'll investigate on how to improve the contrast with the tools that ReDoc provides to customize. I'm sure i can resolve that.

belbiy:
Unfortunately, the theming mechanism is limited in the current version of Redoc. This is what we can change so far: https://github.com/Rebilly/ReDoc/blob/master/src/theme.ts

I have made some changes to the color scheme to increase the contrast and I reset the font to default devdocs font declaration.

We just have to wait for theming improvements from them, because it's the only reliable way of making the visual changes.

jeff-matthews:
Thanks for the updates @belbiy!

PR #3270, opened 2018-11-07
Implement link checking as a Jekyll hook

dshevtsov:
## This PR is an:

Improvement

## Summary

When this pull request is merged, it will enable a Jekyll hook with link checking functionality when the `jekyll serve` command is run and after each regeneration.

```
bundle exec jekyll serve
```

The report is shown in console without breaking the Jekyll server.

html-proofer options are moved to the `_config.checks.yml`. The `exclude` list is added to the `url_ignore` list. It makes possible to use `_config.local.yml` for selective site generation.

Both implementations rake and hook now read options from the `_config.checks.yml`.

Fixed the issue with `baseurl` on staging environment using the `url_swap` option in html-proofer.


configs/_config.checks.yml
```diff
@@ -0,0 +1,67 @@
+html_proofer:
```
jeff-matthews: Is there a reason why we can't keep this config at the root of the project like all other config YAMLs?

configs/_config.checks.yml
```diff
@@ -0,0 +1,67 @@
...
+  file_ignore:
```
jeff-matthews: Before merging, you'll need to remove all entries from the list to match the `master` branch. We merged several PRs today to increase test coverage and I want to be sure we continue checking those.

configs/_config.checks.yml
dshevtsov: No, we can still keep it there
This was a part of my suggestion to move all the special configs (`.prod.yml`, `.local.yml`, etc) into the `config` directory; and keep only `_config.yml` at the root.
Contributors won't need these configs in most cases, so no need to expose them at the root level.

configs/_config.checks.yml
dshevtsov: sure

dshevtsov:
@jeff-matthews updated
please check

_config.stage.yml
```diff
@@ -0,0 +1 @@
+url: "https://devdocs.magedevteam.com"
```
jeff-matthews: Does this file fix link checking on the Jenkins staging build? Seems a bit out of scope for this PR, but if it works, great.

guides/v2.1/mrg/ee/MsrpStaging.md
```diff
@@ -23,4 +23,4 @@ You can find the list of modules that have dependencies on the Magento_MsrpStagi
 
 ## Additional information
 
-For more Magento 2 developer documentation, see [Magento 2 Developer Documentation]({{ site.baseurl }}). Also, there you can track [backward incompatible changes made in a {{site.data.var.ee}} mainline after the Magento 2.0 release]({{ site.baseurl }}/guides/v2.0/release-notes/backward-incompatible-changes/commerce.html).
+For more Magento 2 developer documentation, see [Magento 2 Developer Documentation]({{ site.baseurl }}/). Also, there you can track [backward incompatible changes made in a {{site.data.var.ee}} mainline after the Magento 2.0 release]({{ site.baseurl }}/guides/v2.0/release-notes/backward-incompatible-changes/commerce.html).
```
jeff-matthews: I'm glad you fixed these links, but it's not a good idea to mix too many different types of commits in a single PR. It's difficult to understand the PRs purpose after a while.

rakelib/proofer.rb
```diff
@@ -1,77 +1,22 @@
...
-      typhoeus: { followlocation: true, connecttimeout: 10, timeout: 30 },
-      hydra: { max_concurrency: 50 },
-      # cache: { :timeframe => '30d' }
-    }
-  end
-
-  # Convert array of strings to array of regular expressions
-  def array_to_re(array)
-    array.map { |item| /#{item}/ }
+    config = YAML.load_file('_config.checks.yml')
```
jeff-matthews: Can you add comments to describe what's going on here?

_config.stage.yml
dshevtsov: For the fix, we need a `baseurl` without `url`.  That's why I moved the URL in the separate file. We cannot do it in the CLI like with `--baseurl`.
But seems that staging works ok with `baseurl` only, so we don't need this anymore.

guides/v2.1/mrg/ee/MsrpStaging.md
dshevtsov: I understand this and totally agree.
But I had to ensure that errors reported by the html-proofer on staging were caused by broken links, not bugs in my script.

rakelib/proofer.rb
dshevtsov: I added a comment to the method.
Should I add comment to this line, or to each line in this method?

guides/v2.1/mrg/ee/MsrpStaging.md
dshevtsov: I can revert the commit with broken links fix in this PR

_config.stage.yml
dshevtsov: With no URL, redirects use URL from the `_config.yml`.
So, we will still need `_config.stage.yml` to override the main config.

guides/v2.1/mrg/ee/MsrpStaging.md
jeff-matthews: No, don't worry about reverting.

PR #3273, opened 2018-11-08
MAGECLOUD-2894 Move global variables

hguthrie:
This PR moves the Global environment variables from the Environment variable landing (intro) page into a separate topic, just like the other variable types.


guides/v2.1/cloud/env/variables-intro.md
```diff
@@ -9,6 +9,7 @@ functional_areas:
 {{site.data.var.ece}} enables you to assign environment variables to override configuration options:
 
 -   [ADMIN]({{ page.baseurl }}/cloud/env/environment-vars_magento.html)—variables override project ADMIN variables
```
meker12: variables that override ...

For parallelism, add "that" to all the descriptions --  possible just remove the em-dash and the "that"

guides/v2.1/cloud/env/variables-global.md
```diff
@@ -0,0 +1,72 @@
...
+stage:
+  global:
+    GLOBAL_VARIABLE_NAME: value
+```
+### `MIN_LOGGING_LEVEL`
+
+-  **Default**—_Not set_
+-  **Version**—Magento 2.1.4 and later
+
+Overrides the minimum logging level for all output streams without making changes to the code. This helps to improve troubleshooting problems with deployment. For example, if your deployment fails, you can use this variable to increase the logging granularity globally. See [Set up notifications—Log levels]({{ page.baseurl }}/cloud/env/setup-notifications.html#log-levels). The `min_level` value in Logging handlers overwrites this setting.
```
meker12: "This helps to improve troubleshooting problems with deployment."  -- sounds a bit odd; makes it easier to troubleshoot .... or helps when troubleshooting?

guides/v2.1/cloud/env/variables-global.md
```diff
@@ -0,0 +1,72 @@
...
+stage:
+    global:
+        MIN_LOGGING_LEVEL: debug
+```
+### `SCD_ON_DEMAND`
+
+-  **Default**—_Not set_
+-  **Version**—Magento 2.1.4 and later
+
+Enable generation of static content when requested by a user. This is ideal for the development and testing workflow, because it decreases the deployment time.
```
meker12: This is ideal ...  seems vague --   Maybe:  During development and testing, you can set this variable to true to decrease deployment time.      --  it makes me wonder about what would happen if I left it set in production workflow.

guides/v2.1/cloud/env/variables-global.md
```diff
@@ -0,0 +1,72 @@
...
+
+Pre-loading the cache using the [`post_deploy` hook]({{ page.baseurl }}/cloud/project/project-conf-files_magento-app.html#hooks) reduces site downtime. The cache warming is available only for Pro projects that contain Staging and Production environments in the Project Web UI and for Starter projects. Add the `SCD_ON_DEMAND` environment variable to the `global` stage in the `.magento.env.yaml` file:
```
meker12: I'm not sure of the connection between the first paragraph and the second.  I know the SCD concepts are complex. Not sure if there is a way to connect the concepts better -- pre-loading the cache  relationship to SCD_ON_DEMAND.

guides/v2.1/cloud/env/variables-global.md
```diff
@@ -0,0 +1,72 @@
...
+
+```yaml
+stage:
+  global:
+    SCD_ON_DEMAND: true
+```
+
+The `SCD_ON_DEMAND` variable skips the SCD and the `STATIC_CONTENT_SYMLINK` in both phases (build and deploy), clears the `pub/static` and `var/view_preprocessed` folders, and writes the following to the `app/etc/env.php` file:
```
meker12: make STATIC_CONTENT_SYMLINK  a link to https://devdocs.magento.com/guides/v2.2/cloud/env/variables-deploy.html#static_content_symlink ?

General observation -- I know SCD is a known acronym but I always find it hard to understand the descriptions when we use too many acronyms.  Would it make sense sometimes to spell it out for clarity -- especially when explaining how a process works .

The `SCD_ON_DEMAND` variable skips deploying static content and generating static content symlinks (``STATIC_CONTENT_SYMLINK`) in both the build and deploy phase, clears the ....

Also-- I find myself wondering if the skipping part is related to the "SKIP_SCD" variable.

I know that these mysteries will be cleared up when you have the SCD doc updates ready, but these are current thoughts on current content that you might be able to tweak in the interim.

hguthrie: Added first occurrence.
I know SCD concepts are enigmatic, somewhat. Clarification is on the horizon.

PR #3281, opened 2018-11-08
fixes wrong escaping

riconeitzel:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [x] Bug fix or improvement

## Summary

When this pull request is merged, it will fix wrong escaping

## Additional information

List all affected URLs 

- https://devdocs.magento.com/guides/v2.2/javascript-dev-guide/javascript/js_init.html
- https://devdocs.magento.com/guides/v2.1/javascript-dev-guide/javascript/js_init.html

### DIFF
--- a/guides/v2.1/javascript-dev-guide/javascript/js_init.md
+++ b/guides/v2.1/javascript-dev-guide/javascript/js_init.md
@@ -55,9 +55,9 @@ Depending on the type of the inserted JS component, processing is performed as f
 return function(config, element) { ... };
 ```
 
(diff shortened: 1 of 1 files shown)


meker12:
@mwr and @riconeitzel  I made some additional edits and corrections to this article.  Have a question about content in line #63:

```If none of the previous cases is true, the component is executed with no further processing. Such a component does not require either config or element. The recommended way to declare such components is using the <script> tag.```

The `[using the <script> tag]` has link encoding in the original file, but there was no anchor reference.

I was not sure if the reference target should be the following section `Declarative notation using the <script type="text/x-magento-init" /> tag` or if it should be linking to the imperative notation section, which uses the `<script>` tag.

PR #3282, opened 2018-11-08
fixes wrongly usage of define function in example

riconeitzel:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [x] Bug fix or improvement

## Summary

When this pull request is merged, it will fix a code example that wrongly shows 'define' instead of 'require'

## Additional information

List all affected URLs 

- https://devdocs.magento.com/guides/v2.2/javascript-dev-guide/javascript/js_init.html
- https://devdocs.magento.com/guides/v2.1/javascript-dev-guide/javascript/js_init.html

### DIFF
--- a/guides/v2.1/javascript-dev-guide/javascript/js_init.md
+++ b/guides/v2.1/javascript-dev-guide/javascript/js_init.md
@@ -154,7 +154,7 @@ In a similar way, you can initialize any JS component that a returns callback fu
 For example:
 
 ```javascript
-define ([
+require ([
     'jquery',
     'mage/gallery/gallery'
 ], function ($, Gallery) {


magento-cicd2:
Can one of the admins verify this patch?

riconeitzel:
From the [require.js source code](url
http://requirejs.org/docs/release/2.0.1/comments/require.js) (line 1902):

```
/**
 * The function that handles definitions of modules. Differs from
 * require() in that a string for the module should be the first argument,
 * and the function to execute after dependencies are loaded should
 * return a value to define the module corresponding to the first argument's
 * name.
 */
```

So it either lacks that first string argument and is an invalid code sample or It is a require call and not a define

VladimirZaets:
Hi @riconeitzel, thanks for collaboration.
The `require` function is used for modules load.  For module declaration we should use `define` function and declare all dependencies that are needed for the module. The dependencies will be loaded when module is called.

> So it either lacks that first string argument and is an invalid code sample or It is a require call and not a define

The first string argument is module name and this argument is optional. 

You can find more information and examples about `define` in the [official documentation] of requirejs library (https://requirejs.org/docs/api.html#define)

PR #3296, opened 2018-11-14
Unify usage of magento cli code blocks

Vinai:
Unify usage of the bin/magento command in bash code blocks.

## This PR is a:

- [ ] New topic
- [ ] Content update
- [x] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

Some commands where missing the `magento` command and only specified arguments, and because only the first line of a bash code block is prefixed with a $ automatically, the second line needs to include the `$` of the prompt explicitly.

## Additional information

Affected page:
https://devdocs.magento.com/guides/v2.3/extension-dev-guide/declarative-schema/migration-commands.html


keharper:
Thanks @Vinai. We've had some internal discussions lately about how to refer to the `magento` command, and `bin/magento` was the preferred way. I updated the command examples to include the `bin/`.

PR #3354, opened 2018-11-26
Update language around async code

DrewML:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [X] Bug fix or improvement

## Summary

When this pull request is merged, it will clarify language around asynchronous operations.

It's clear that `10.5.3` was added with the intention of enforcing async ajax requests. However, the current language enforces everyone to model _any_ async code with an HTTP request, which doesn't make any sense.

As an example, `setTimeout` from the W3C spec can be used for timing and scheduling of async operations, but has nothing to do with external requests.


magento-cicd2:

jeff-matthews:
run tests

PR #3369, opened 2018-11-28
Declarative schema: reference ID changes

keharper:
## This PR is a:

- [ ] New topic
- [x] Content update
- [ ] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

This pull request replaces the `name` parameter, formerly used in constraint and index definitions, with `referenceId`

## Additional information

List all affected URLs 
https://devdocs.magento.com/guides/v2.3/extension-dev-guide/declarative-schema/migration-commands.html

https://devdocs.magento.com/guides/v2.3/extension-dev-guide/declarative-schema/db-schema.html


keharper:
run tests

guides/v2.3/extension-dev-guide/declarative-schema/db-schema.md
```diff
@@ -143,18 +159,19 @@ Example:
...
-The `constraint` subnode is represented by 3 different types:
+The `constraint` subnode always contains the following parameters:
 
-* `primary`
-* `unique`
-* `foreign`
+Attribute | Description
+--- | ---
+`type` | One of `primary`, `unique`, or `foreign`
+`referenceId` |  In most instances, a system-generated reference ID that represents an entity in the schema. System-generated IDs cannot be changed. However, you can override `referenceId` values.
```
fascinosum: This is uncorrect. We have to change it to something like that:
This is a custom identifier that is used only for relation mapping in the scope of Declarative Schema declarations - db_schema.xml files. The real entity in the database will have a system-generated name. The most convenient way - to set the value of this attribute the same as the system-generated name. You can get thesenames from db_schema_whitelist.json as the result of execution `generate-whitelist` command [https://devdocs.magedevteam.com/kh_schema-updates/guides/v2.3/extension-dev-guide/declarative-schema/migration-commands.html#create-a-schema-whitelist]

guides/v2.3/extension-dev-guide/declarative-schema/db-schema.md
```diff
@@ -163,69 +180,141 @@ The `foreign` constraint is similar to foreign keys in SQL. This type of constra
...
 ```
+
 #### `index` subnode
 
-The `index` subnode has the same structure as internal constraints but contains different logic. While constraints are used for defining limitations, indexes are used for speeding up DQL operations.
+The `index` subnode has the same structure as internal constraints but contains different logic. While constraints are used for defining limitations, indexes are used for speeding up DQL operations. The following attributes define an index:
+
+Attribute | Description
+--- | ---
+`referenceId` | In most instances, a system-generated reference ID that represents an entity in the schema. System-generated IDs cannot be changed. However, you can override `referenceId` values
```
fascinosum: The same as for constraints

guides/v2.3/extension-dev-guide/declarative-schema/db-schema.md
```diff
@@ -163,69 +180,141 @@ The `foreign` constraint is similar to foreign keys in SQL. This type of constra
...
 The following example changes the `type` of the `title` column from `varchar` to  `tinytext`.
 
-![Change column type]({{ page.baseurl }}/extension-dev-guide/declarative-schema/images/change-column-type.png)
+```diff
+<schema xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                 xsi:noNamespaceSchemaLocation="urn:magento:framework:Setup/Declaration/Schema/etc/schema.xsd">
+    <table name="declarative_table">
+        <column xsi:type="int" name="id_column" padding="10" unsigned="true" nullable="false" comment="Entity Id"/>
+        <column xsi:type="int" name="severity" padding="10" unsigned="true" nullable="false" comment="Severity code"/>
++       <column xsi:type="tinytext" name="title" nullable="false" length="255" comment="Title"/>
```
fascinosum: -       <column xsi:type="varchar" name="title" nullable="false" length="255" comment="Title"/>
+       <column xsi:type="tinytext" name="title" nullable="false" length="255" comment="Title"/>

guides/v2.3/extension-dev-guide/declarative-schema/db-schema.md
```diff
@@ -245,20 +334,107 @@ onCreate="migrateDataFromAnotherTable(catalog_category_entity,entity_id)"
...
+-       <constraint xsi:type="foreign" referenceId="FL_ALLOWED_SEVERITIES" table="declarative_table" 
+-               column="severity" referenceTable="severities" referenceColumn="severity_identifier" 
+-               HonDelete="CASCADE"/>
+    </table>
+</schema>
+```
+
+### Recreate a foreign key
+
+In this example, Module A defines a new table with primary key `id_column`. Module B declares its own schema, in which it creates a new column (`new_id_column`) and changes the primary index to this column. Module B disables the original primary key and sets the `referenceId` to a value be different from PRIMARY. Although the `referenceId` is different, the real name of the primary key in the database remains PRIMARY.
```
fascinosum: > Module B disables the original primary key and sets the `referenceId` to a value be different from PRIMARY. Although the `referenceId` is different, the real name of the primary key in the database remains PRIMARY. 

Module B disables the original primary key and **sets a new primary key** with `referenceId` value that is different from PRIMARY. Although the `referenceId` is different, the real name of the primary key in the database remains PRIMARY.

guides/v2.3/extension-dev-guide/declarative-schema/migration-commands.md
```diff
@@ -25,6 +25,8 @@ bin/magento setup:install --convert-old-scripts=1
 bin/magento setup:upgrade --convert-old-scripts=1
 ```
 
+The `convert-old-scripts` option converts tables and columns only. Renaming tables is not supported.
```
fascinosum: This is a known issue, we have to emphasize this aspect. We are working on it.

misha-kotov: Agreed - currently it sounds like we made it that way on purpose. In reality, its' a known issue, meaning in the future we will expand support. I would probably remove the entire note about tables and columns, and renaming; just keep this info in the known issues list.

guides/v2.3/extension-dev-guide/declarative-schema/db-schema.md
```diff
@@ -245,20 +335,108 @@ onCreate="migrateDataFromAnotherTable(catalog_category_entity,entity_id)"
 
 The following example adds the `INDEX_SEVERITY` index to the `table_name` table.
```
meker12: Check:  In the example, the table name = "declarative_table"  --   Should this be: 

adds the `INDEX_SEVERITY` index to the `declarative_table` table?

guides/v2.3/extension-dev-guide/declarative-schema/db-schema.md
```diff
@@ -245,20 +335,108 @@ onCreate="migrateDataFromAnotherTable(catalog_category_entity,entity_id)"
...
+-               column="severity" referenceTable="severities" referenceColumn="severity_identifier" 
+-               HonDelete="CASCADE"/>
+    </table>
+</schema>
+```
+
+### Recreate a foreign key
+
+In this example, Module A defines a new table with primary key `id_column`. Module B declares its own schema, in which it creates a new column (`new_id_column`) and changes the primary index to this column. 
+Module B disables the original primary key and sets a new primary key with a `referenceId` value that is different from PRIMARY. Although this value is different, the real name of the primary key in the database remains PRIMARY.
```
meker12: "Although this value is different, the real name of the primary key in the database remains PRIMARY."  -- this is confusing -- Why does it keep the same primary key if the code sets a new primary key with a referenceId of NEW_PRIMARY?

guides/v2.3/extension-dev-guide/declarative-schema/migration-commands.md
```diff
@@ -108,7 +108,7 @@ Each CSV file contains a row that defines the column (or other database entity)
 
 ![Dump Example]({{ page.baseurl }}/extension-dev-guide/declarative-schema/images/dump_example.png)
 
-## Maintain backward compatibility
+## Create a schema whitelist {#create-whitelist}
```
meker12: There is an industry trend to replace "whitelist" with "allow list" and "blacklist" with block list.  We might want to consider that for our doc style guide.

guides/v2.3/extension-dev-guide/declarative-schema/migration-commands.md
```diff
@@ -108,7 +108,7 @@ Each CSV file contains a row that defines the column (or other database entity)
...
 
 Backward compatibility must be maintained. Therefore, declarative schema does not automatically delete database tables, columns or keys that are not defined in a `db_schema.xml` file. Declarative schema can't delete these elements because these items can be declared somewhere else, such as in an `Setup/UpgradeSchema.php` file.
```
meker12: can't ==>  cannot

guides/v2.3/extension-dev-guide/declarative-schema/db-schema.md
keharper: You're correct. I changed my mind on the table name midstream

guides/v2.3/extension-dev-guide/declarative-schema/migration-commands.md
keharper: The command name includes the word "whitelist". I can't change it, and it has nothing to do with allowing/blocking,

PR #3382, opened 2018-11-29
update the backward compatibility guide

fascinosum:
## This PR is a:

- [ ] New topic
- [x] Content update
- [ ] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

## Additional information

List all affected URLs 

- https://devdocs.magento.com/guides/v2.1/contributor-guide/backward-compatible-development/

### DIFF
--- a/guides/v2.1/contributor-guide/backward-compatible-development/index.md
+++ b/guides/v2.1/contributor-guide/backward-compatible-development/index.md
@@ -39,6 +39,12 @@ Mark the method with the `@deprecated` tag instead of removing it.
 
 Continue returning the same results from the method if possible, so the old functionality is preserved.
 
+If you need to remove the existing functionality and you do not need alternative implementation, you SHOULD change the method code to
+
+```php
+trigger_error('Function is deprecated', E_USER_WARNING);
+```
+
 #### Introduction of a method to a class or interface
 
 Create a new interface with a new method instead of introducing a method to a class or interface.
--- a/guides/v2.1/contributor-guide/backward-compatible-development/index.md
+++ b/guides/v2.1/contributor-guide/backward-compatible-development/index.md
@@ -42,7 +42,7 @@ Continue returning the same results from the method if possible, so the old func
 If you need to remove the existing functionality and there is no alternative implementation, you MUST change the method code to
 
 ```php
-trigger_error('THE_REASON_FOR_CHANGE', E_USER_WARNING);
+trigger_error('THE_REASON_FOR_CHANGE', E_USER_ERROR);
 ```
 
 #### Introduction of a method to a class or interface
--- a/guides/v2.1/contributor-guide/backward-compatible-development/index.md
+++ b/guides/v2.1/contributor-guide/backward-compatible-development/index.md
@@ -39,7 +39,7 @@ Mark the method with the `@deprecated` tag instead of removing it.
 
 Continue returning the same results from the method if possible, so the old functionality is preserved.
 
-If you need to remove the existing functionality and you do not need alternative implementation, you SHOULD change the method code to
+If you need to remove the existing functionality and there is no alternative implementation, you MUST change the method code to
 
 ```php
 trigger_error('Function is deprecated', E_USER_WARNING);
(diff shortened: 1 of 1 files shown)


magento-cicd2:

guides/v2.1/contributor-guide/backward-compatible-development/index.md
```diff
@@ -39,6 +39,12 @@ Mark the method with the `@deprecated` tag instead of removing it.
 
 Continue returning the same results from the method if possible, so the old functionality is preserved.
 
+If you need to remove the existing functionality and you do not need alternative implementation, you SHOULD change the method code to
+
+```php
+trigger_error('Function is deprecated', E_USER_WARNING);
```
buskamuza: Deprecation is not what happens in such situation, so I think deprecation warning will be misleading.
I'd recommend real exception or, even better, removal of entire method in such situation. So that it's visible to everyone that there is a breaking change.

guides/v2.1/contributor-guide/backward-compatible-development/index.md
```diff
@@ -39,6 +39,12 @@ Mark the method with the `@deprecated` tag instead of removing it.
 
 Continue returning the same results from the method if possible, so the old functionality is preserved.
 
+If you need to remove the existing functionality and you do not need alternative implementation, you SHOULD change the method code to
```
buskamuza: I'd rephrase, "and you do not need alternative implementation" to "and there is no alternative implementation" because developer of this part of code obviously don't need this functionality, but we need to care about other developers, and they may need it. So usually we should preserve the functionality even if core Magento doesn't need it. The only exception is when it's impossible to provide an alternative implementation.

buskamuza: Also, "SHOULD" should be changed to "MUST". There are no exceptions to this rule.

magento-cicd2:

hguthrie:
@fascinosum any status on this?

guides/v2.1/contributor-guide/backward-compatible-development/index.md
fascinosum: `trigger_error` is much better than Exception, you can't catch the user error in the catch block. Removing a  method is a backward incompatible implementation and sometimes it can't be implemented, when you have an inheritance

guides/v2.1/contributor-guide/backward-compatible-development/index.md
```diff
@@ -39,6 +39,12 @@ Mark the method with the `@deprecated` tag instead of removing it.
 
 Continue returning the same results from the method if possible, so the old functionality is preserved.
 
+If you need to remove the existing functionality and there is no alternative implementation, you MUST change the method code to
```
joni-jones: Actually, you can ADD `trigger_error` method but don't replace the existing. The method should still return the same result.

joni-jones: Actually, you can **add** `trigger_error` method but don't replace the existing. The method should still return the same result.

magento-cicd2:

guides/v2.1/contributor-guide/backward-compatible-development/index.md
fascinosum: It is impossible to return the same result and trigger error at the same time. I've changed the type of error

joni-jones: What type of error are you trying to use? Doesn't it work with `E_WARNING`?

magento-cicd2:

guides/v2.1/contributor-guide/backward-compatible-development/index.md
fascinosum: `E_USER_ERROR` is a real PHP error and this avoids hidden warnings in the log

joni-jones: @fascinosum why to don't use `E_DEPRECATED`? PHP core team uses it to "mark" deprecated functionality. Why the deprecated code should throw an error?

fascinosum: @joni-jones because this is not a deprecated code, but a backward incompatible change. You must not use this method anymore

joni-jones: BC guide should not contain approaches which introduce BIC.

akaplya: Why should we keep a method if we can not mimic an execution result?

PR #3403, opened 2018-12-04
Add Sitemap to fix duplicated URLs in Google search results (MAGEDOC-3283)

belbiy:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [X] Bug fix or improvement

## Summary

When this pull request is merged, it will add the `sitemap.xml` file with the 2.3 pages listed. 2.0, 2.1, 2.2 will not be included. Hopefully, this will help Google prioritize 2.3 over the old versions.

The sitemap is generated using the `jekyll-sitemap` plugin with the help of `jekyll-last-modified-at` to correct modification time of files.

In `_config.yml` i have listed to exclude the 2.0, 2.1, 2.2 versions from the sitemap:

```
 -
    scope:
      path: guides/v2.0
    values:
      sitemap: false
  -
    scope:
      path: guides/v2.1
    values:
      sitemap: false
  -
    scope:
      path: guides/v2.2
    values:
      sitemap: false
```
When the new version of Magento docs would be released, the 2.3 version should be excluded as well.

In the future branching strategy, we can still use this file but need to make sure that we only generate the sitemap for the current version somehow.

## Additional information

List all affected URLs: 
- Only sitemap.xml

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 
 gem 'github-pages'
+gem 'jekyll-last-modified-at'
 gem 'devdocs', :git => 'https://github.com/magento-devdocs/devdocs-theme.git', :branch => 'devdocs-stable'
 
 gem 'wdm' if Gem.win_platform?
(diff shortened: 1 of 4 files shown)


belbiy:
running tests

belbiy:
running tests

PR #3408, opened 2018-12-05
Implement autogeneration for third-party licenses topics

dshevtsov:
## This PR is a:

- Content fix or rewrite

## Summary

When this pull request is merged, it will implement autogenerated tables into topic about third-party licenses in the 2.3 documentation. Data sources for the tables are the corresponding `composer.lock` files.


guides/v2.3/release-notes/thirdparty-mccloud.md
```diff
@@ -1,101 +1,15 @@
 ---
 group: release-notes
-subgroup: Third-party licenses
-menu_title: Magento Commerce (Cloud) third-party licenses
 title: Magento Commerce (Cloud) third-party licenses
```
hguthrie:
```suggestion
title: Magento Commerce Cloud third-party licenses
```

keharper:
This is awesome. We need to consider other repositories, such as those for Community Engineering projects like MSI, GraphQL, and Page Builder. I spot checked the MSI `composer.json` file and noticed that `"magento/magento2-functional-testing-framework": "2.3.5",` was in the `require-dev` section, but it's not listed in the build Dima provided. 

We should also plan on updating these files every patch release, instead of only at major releases.

keharper:
This is awesome. We need to consider other repositories, such as those for Community Engineering projects like MSI, GraphQL, and Page Builder. I spot checked the MSI `composer.json` file and noticed that `"magento/magento2-functional-testing-framework": "2.3.5",` was in the `require-dev` section, but it's not listed in the build Dima provided. 

We should also plan on updating these files every patch release, instead of only at major releases.

dshevtsov:
@keharper this topic is for third-party packages. The list is filtered to avoid `magento/` packages.

dshevtsov:
@keharper this topic is for third-party packages. The list is filtered to avoid `magento/` packages.

keharper:
Fair enough. That item just jumped out at me. Here's an external one:

` "squizlabs/php_codesniffer": "3.3.1"`

dshevtsov:
running tests

PR #3414, opened 2018-12-06
Update guidelines regarding blocks

AlexMaxHorkun:
A PHPMD rule has been created to prevent this but deemed to be unreliable so we've decided to just update the guide.

## This PR is a:

- Content update

## Summary

Include an important (for security) update regarding blocks

## Additional information

https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/devdocs?pullRequest=3414) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/devdocs?pullRequest=3414) before we can accept your contribution.<br/>

magento-cicd2:

jeff-matthews:
@buskamuza, please review.

magento-cicd2:

magento-cicd2:

magento-cicd2:

jeff-matthews:
@buskamuza & @melnikovi, please review this week or we will need to close this PR.

guides/v2.2/coding-standards/technical-guidelines.md
```diff
@@ -520,6 +520,8 @@ class View extends Template
 
 6.2.4. Actions MUST NOT reference blocks declared in layout.
 
+6.2.5 Blocks MUST NOT assume user input has been validated in an action and, if they use user input directly, must validate it themselves.
```
melnikovi: I would suggest to make it more generic and move to security section. We threat user input this way everywhere.

jeff-matthews: @melnikovi, how would you suggest making it more generic?

melnikovi: I would say that in business logic we should not assume that input has been validated.

AlexMaxHorkun: > I would say that in business logic we should not assume that input has been validated.

1st of all that's too much of a broad statement - I doubt it will catch developers attention
2nd: it's better to validate user input right when you receive it (in a controller or a block) because otherwise you'd have to duplicate validations across all classes used to accomplish an action and perform redundant checks that were already performed prior to current method call.

buskamuza: It should be either in "Basic programming principles" or in "Security" section. How blocks are different from any other classes in terms of input validation?
IMO, data should be validated in the class that owns it (class that knows how to use it). In case of controllers/blocks/templates, it's a block.
From another side, if controller is manipulating the data, it should also validate it because it has some expectations about its format.

dobooth:
@AlexMaxHorkun Can you please fix the merge conflict in this PR?
We would like to merge or close this.

guides/v2.2/coding-standards/technical-guidelines.md
AlexMaxHorkun: I still think this statement should be inside the blocks section since there is a problem with people assuming they are being attached to a controller when in fact they are not. Also it wouldn't be true to say "validate every argument everywhere" since it's not a good practice to duplicate validation in every layer. In general classes that are being directly responsible for user input processing should do the validation since they are aware of the context. Classes down the line don't know the context and may not be aware on how to validate arguments.
Blocks can be used as user-input processor directly without a controller and therefor are responsible for user-input validation.

melnikovi: Blocks are not the only place where data might need to be validated. It can also be controller, model or some other place. This is why I would prefer to make it more generic and not only about blocks. Also agree with @buskamuza about moving to different section.

buskamuza: I wouldn't say "validate everything everywhere", but I'd like to avoid a situation where tomorrow we add another rule "Validate arguments in view models", and then in one month "Validate input in controllers", etc.
I'd put something like "Input validation SHOULD be performed by the class performing operation on it." in the "Basic programming principles" section.

AlexMaxHorkun: The original problem stems from developers not understanding that blocks might be used as the 1st level user-input processors without controllers and thus being responsible for validation. How do you counter that problem by moving it into a different section?

melnikovi: I think it might be easier if everything security related can be found under one section.

dobooth:
running tests

PR #3447, opened 2018-12-12
MAGEDOC-3040: RSS enhancements

jeff-matthews:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [X] Bug fix or improvement

## Summary

When this pull request is merged, it will add the last last modified date to each topic in the `feed.xml` file so updated topics show in a feed reader; not just new topics.


feed.xml
```diff
@@ -12,12 +12,14 @@ layout: null
...
     <generator>Jekyll v{{ jekyll.version }}</generator>
-    {% for page in site.html_pages %}
+    {% assign sorted_pages = site.html_pages | sort:"last_modified_at" | reverse %}
+    {% for page in sorted_pages %}
       {% unless page.tag == 'omit' %}
         <item>
           <title>{{ page.title | xml_escape }}</title>
           <description>{{ page.content | xml_escape | strip_html | truncatewords: 75 }}</description>
           <link>{{ page.url | prepend: site.url }}</link>
+          <pubDate>{{ page.last_modified_at | date_to_rfc822 }}</pubDate>
```
dshevtsov: `<pubDate>` seems to be for `publication date`, but in fact it is not.
Are there tags available for update or modification date?

jeff-matthews: Per RSS specification, the only tags that are available for the `<item>` element are `pubDate` and `<lastBuildDate>`.

feed.xml
```diff
@@ -12,12 +12,14 @@ layout: null
     <pubDate>{{ site.time | date_to_rfc822 }}</pubDate>
     <lastBuildDate>{{ site.time | date_to_rfc822 }}</lastBuildDate>
     <generator>Jekyll v{{ jekyll.version }}</generator>
-    {% for page in site.html_pages %}
+    {% assign sorted_pages = site.html_pages | sort:"last_modified_at" | reverse %}
+    {% for page in sorted_pages %}
```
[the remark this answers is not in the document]
jeff-matthews: Per best practices, we should limit the number of posts: https://webmasters.googleblog.com/2014/10/best-practices-for-xml-sitemaps-rssatom.html.

The question is, what number do we use?

feed.xml
```diff
@@ -12,12 +12,14 @@ layout: null
...
       {% unless page.tag == 'omit' %}
```
dshevtsov: if  we suppose to use the tag, it should be moved to assignment filters, for example:
```
{% assign sorted_pages = site.html_pages | where_exp: "page", "page.tag != 'omit'"| sort:"last_modified_at" | reverse %}
```
If we don't have real cases to use it, it's better to remove it for now to not overload the generation logic. We can add it when we'll have the corresponding cases.

jeff-matthews: I'm ok with that. I just wanted to make sure that we could ignore certain files if we needed to. We're not currently using this on devdocs. 

It may be necessary for the OMS docs site, however. For example, to exclude the German translation of pages based on the `lang:` metadata tag so that we build feeds for each language appropriately. Otherwise, the feed will include both languages, which is not good.

feed.xml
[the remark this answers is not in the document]
jeff-matthews: I set it to 100.

jeff-matthews:
running tests

PR #3463, opened 2018-12-13
Add a pliugin that generates canonical URLs

dshevtsov:
## This PR is a:

- Improvement

## Summary

When this pull request is merged, it will add a plugin that generates `page.canonical_url` unless it is set in the front matter.  `canonical_url` is an absolute URL to be used as a variable for canonical URL in layouts. For all pages in `guides` of releases lower than 2.3 the canonical URL will be replaces with link to 2.3. For example the `https://devdocs.magento.com/guides/v2.0/ftf/introduction.html` page will contain canonical URL `http://devdocs.magento.com/guides/v2.3/ftf/introduction.html`.
The goal is to prioritize the latest version of documentation in web search.


belbiy:
running tests

dshevtsov:
there is no corresponding check whether a generated pat exists in 2.3, it will make the generator very slow.
@belbiy confirmed that this behavior is ok for this task.

belbiy:
Right now all the canonical pages exist for 2.3 guides. I have checked with htmlproofer by outputing the canonical URL as an `<a>` tag on the page. 

We can probably configure the html proffer to check of `<link rel="canonical">` as well. But for now, this will work.

dshevtsov:
running tests

PR #3467, opened 2018-12-14
Optimize plugins to enhance performance

dshevtsov:
## This PR is a:

- Improvement

## Summary

When this pull request is merged, it will disable parameter generation made in plugins that are not needed in serving mode; optimize performance with more appropriate use of conditionals; add more documentation in plugins.

### DIFF
--- a/_plugins/html-check-hook.rb
+++ b/_plugins/html-check-hook.rb
@@ -8,25 +8,35 @@ require 'html-proofer'
 require 'yaml'
 
 Jekyll::Hooks.register :site, :post_write do |site|
-  # If 'jekyll serve' is run, read options for html-proofer
-  # in '_config.checks.yml' and add 'excludes' from Jekyll configurtiuon
-  # to the 'url_ignore' list if it exists.
-  #
-  if site.config['serving'] && site.config['check_links']
-    begin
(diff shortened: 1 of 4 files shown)


_plugins/html-check-hook.rb
```diff
@@ -8,25 +8,35 @@
...
-        jekyll_excludes_as_regex = jekyll_excludes.map { |item| Regexp.new Regexp.escape(item) }
-        url_ignore.push(jekyll_excludes_as_regex).flatten!.uniq!
-        checks_config['html-proofer'][:url_ignore] = url_ignore
-      end
-      options = checks_config['html-proofer']
-      # Run html-proofer to check the jekyll destination directory
-      HTMLProofer.check_directory(site.dest, options).run
-    rescue StandardError
-      puts 'Fix the broken links before you push the changes to remote branch.'.blue
+  # Do nothing unless serving mode is enabled
```
jeff-matthews: Do nothing "if" serving mode is enabled?

dshevtsov: We use the hook only in serving mode, it is for local development only.
Link testing that we use for building is run as a rake task, it doesn't use this plugin, but uses the same options.

dshevtsov:
running tests

PR #3471, opened 2018-12-16
Transform Whats New on Devdocs to a template working with data and add a news feed for the topic updates

dshevtsov:
## This PR is a:

- Improvement

## Summary

When this pull request is merged, it will transform the Whats New on DevDocs topic to a data with template model and add an Atom feed.

## Additional information

1. The content is generated automatically, so it is easier to maintain the data separately.
2. The data stored separately allows to implement a news feed (community request: dshevtsov/whats-new-on-devdocs#1)


whatsnew-feed.xml
```diff
@@ -0,0 +1,21 @@
+---
+layout: null
```
jeff-matthews: It's not obvious to me where you're using this file... What is its purpose? Are you proposing that we replace the existing `feed.xml` file with this?

dshevtsov: It is another channel that can be used. We had a community request to implement it.
You can add it to an RSS reader as well as `feed.xml` to track Whats New updates.
`feed.xml` contains any updates to files, and `whatsnew-feed.xml` contains only the updates highlighted by writers.

dshevtsov: Added an alternate link to the body per suggestion from @belbiy . It also works.

dshevtsov:
running tests

PR #3519, opened 2019-01-03
Added redirect for old ece-tools update topic

meker12:
For the ece-tools update topic, Zendesk Support ticket footers are pointing to  this URL :   https://devdocs.magento.com/guides/v2.2/cloud/composer-packages/ece-tools-update.html
which returns `Page Not Found`.

## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

Updated the current ece-tools update topic to add a redirect for the older version of the topic.

**Affected URLs**
https://devdocs.magento.com/guides/v2.1/cloud/project/ece-tools-update.html
https://devdocs.magento.com/guides/v2.2/cloud/project/ece-tools-update.html
https://devdocs.magento.com/guides/v2.3/cloud/project/ece-tools-update.html

### DIFF
--- a/guides/v2.1/cloud/project/ece-tools-update.md
+++ b/guides/v2.1/cloud/project/ece-tools-update.md
@@ -4,6 +4,8 @@ title: Update ece-tools version
 redirect_from:
   - guides/v2.1/cloud/composer-packages/ece-tools.html
   - guides/v2.2/cloud/composer-packages/ece-tools.html
+  - guides/v2.1/cloud/composer-packages/ece-tools-update.html
+  - guides/v2.2/cloud/composer-packages/ece-tools-update.html
 functional_areas:
   - Cloud
   - Upgrade


dshevtsov:
@meker12 I cannot find the `guides/v2.2/cloud/composer-packages/ece-tools-update.md` file in git history. Seems like it was a wrong link in Zendesk Support ticket only. Is it possible to update just ticket without adding new redirects?

hguthrie:
The file never existed in the `composer-packages` directory. I recently created in NEW in the `project` directory. I have no idea where the link originated. I feel it is more appropriate for them to update their signatures with the latest URL.

meker12:
> The file never existed in the `composer-packages` directory. I recently created this NEW file in the `project` directory. I have no idea where the link originated. I feel it is more appropriate for them to update their signatures with the latest URL.

I would agree except that they have more than a year of existing tickets with the incorrect link.  It seems like a simple solution to add the redirect here, so that references to the incorrect link do not return a 404.

dshevtsov:
This simplicity can be a reason of complexity in future (see #3521)
If it is possible to update the original link, it should be done there.

meker12:
running tests

hguthrie:
Regarding old links. I'm not worried about that because you will always encounter bad links in older records. I don't think that is avoidable. I completely understand that this is a quick fix, but if we continue to always do this, our topics may have redirect lists quite long. I'm hoping that a clean sweep of the docs in the future can help with this.

PR #3522, opened 2019-01-06
Add required value to safe-mode and data-restore

Vinai:
Using the `--safe-mode` and `--data-restore` arguments as boolean options without a value has no effect.
They have to be specified with an `=1` value (like for `--dry-run=1`.

## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [x] Bug fix or improvement

## Summary

When this pull request is merged, it will show the proper usage of the `--safe-mode` and `--data-restore` `setup:upgrade` arguments.

## Additional information

List all affected URLs 

- https://devdocs.magento.com/guides/v2.3/extension-dev-guide/declarative-schema/migration-commands.html


magento-cicd2:

keharper:
running tests

PR #3529, opened 2019-01-08
Fix #2756 update command

### LINKED ISSUE #2756: Feedback on page: /guides/v2.2/config-guide/prod/prod_file-sys-perms.html
# General issue

"Production file system ownership for shared hosting (one user)" permissions recommendation causes 500 server error with PHP-FPM & SuPHP (which run as user).

## Description:

This page suggests that User write permissions should be removed in Production for a number of directories and files. Unfortunately, with SuPHP and PHP-FPM, they run as the user so removing write permissions causes Apache to throw a 500 server error when loading storefront.

[17-Aug-2018 00:10:29 UTC] PHP Fatal error:  Uncaught Zend_Cache_Exception: cache_dir "/home/user/public_html/var/page_cache" is not writable in /home/user/public_html/vendor/magento/zendframework1/library/Zend/Cache.php:209

## Possible solutions:

This section doesn't account for hosting, in the One User sections, where the PHP handler runs as the hosting account?

hguthrie:
Bug fix related to #2756 

-  Updated the command to remove write permissions of a set of folders and files.
-  Also removed contractions and fixed the 2.1 version of the instructions

Affects Config Guide: https://devdocs.magento.com/guides/v2.2/config-guide/prod/prod_file-sys-perms.html

### DIFF
--- a/guides/v2.1/config-guide/prod/prod_file-sys-perms.md
+++ b/guides/v2.1/config-guide/prod/prod_file-sys-perms.md
@@ -11,7 +11,7 @@ functional_areas:
 
 This section discusses how to set up the owner or owners of the Magento file system for a development and production system. Before you continue, review the concepts discussed in [Overview of file system ownership and permissions]({{ page.baseurl }}/install-gde/prereq/file-sys-perms-over.html).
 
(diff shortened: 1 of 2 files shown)


hguthrie:
running tests

PR #3562, opened 2019-01-12
Link checker: Check for double forward slashes

jeff-matthews:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [x] Bug fix or improvement

## Summary

When this pull request is merged, it will add a custom test called `DoubleSlashCheck` to extend the html-proofer library, since the library doesn't natively check for this scenario. See https://github.com/gjtorikian/html-proofer/issues/386.

URLs with double slashes will cause 404s on our website because our hosting solution doesn't automatically strip extra forward slashes from URLs like normal web servers.

This PR also fixes existing instances of double forward slashes in URLs.


jeff-matthews:
The scenario is a writer adding an extra forward slash when entering a URL in an `<a>` element (e.g., after a Jekyll variable). The LinkChecker class **will not** identify this as a failure, so we end up with 404s on the website despite the LinkCheck class. That's why we need a custom test/class.

The expected behavior of a failing test is any occurrence of a double forward slash in an `<a>` element in the compiled HTML (_excluding_ double slashes around the protocol like `https://`).

For example:

Fail
`<a href="https://devdocs.magento.com//path/to/file.html">Fail</a>`

Pass
`<a href="https://devdocs.magento.com/path/to/file.html">Pass</a>`

The new class actually runs with the other checks (e.g., ImageCheck, ScriptCheck, LinkCheck, DoubleSlashCheck).

dshevtsov:
ah, now I see it
so, it is not available for the link checking hook

jeff-matthews:
Oops! That wasn't my intention. I assumed that the hook would invoke the new class. Let me know how we can change or optimize the new class. I was just looking for the quickest way to get this working. I'm open to suggestions.

dshevtsov:
Working on it

PR #3597, opened 2019-01-20
Added missing store part in url

mage2pratik:
## This PR is a:

- Content fix or rewrite

## Summary

When this pull request is merged, it will add missing store part in URL.

## Additional information

N/A


keharper:
@dobooth I thought I assigned this to myself, but it looks like I didn't. 

This is a bigger issue than reported in this PR. The sample usage statements need to be consistent across all of devdocs. I'm still considering what format we should use. I'll get back to this PR by the end of the week.

keharper:
@mage2pratik 

Let's make this the official format for sample usage statements:

`POST <host>/rest/<store_code>/V1/inventory/source-items-delete`

Where we list and discuss specific endpoints, omit everything before /V1 so that it matches Swagger.

These standards need to be applied to all the examples in the three topics you changed in this PR. 

If you wish to go beyond these three topics:

- Don't change the store code in the tutorials when the existing store code is important to the tutorial. I'm especially thinking of the Inventory Management tutorial, where we're bouncing around from one store code to another. This might apply to others

- Bulk and async endpoints have their own paths, so be sure to keep their structure while following the above guidelines

- There may be cases where `index.php` appears in the route, or the store code is omitted.  Make these instances consistent as well.

keharper:
@mage2pratik This is a good start. The other verbs need to be updated as well.

devops-devdocs:

keharper:
@mage2pratik Just one more: guides/v2.1/rest/notes.html

keharper:
running tests

keharper:
Thanks so much, @mage2pratik

PR #3595, opened 2019-01-18
Remove obsolete release notes

meker12:
Fixes #3594

## This PR is a:

- [ ] New topic
- [x] Content update
- [x] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

- Removed obsolete release notes topic in v2.1, 2.2, and 2.3.
- Added v2.1, v2.2, v2.3 redirects so that obsolete URLs point to the current release notes topic.
- Updated the Magento Commerce Cloud release information on the [Magento Commerce Release Information page](https://devdocs.magento.com/guides/v2.3/release-notes/bk-release-notes.html).

## Additional information

*Affected URLs *
https://devdocs.magento.com/guides/v2.1/cloud/release-notes/CloudReleaseNotes.html
https://devdocs.magento.com/guides/v2.2/cloud/release-notes/CloudReleaseNotes.html
https://devdocs.magento.com/guides/v2.3/cloud/release-notes/CloudReleaseNotes.html
https://devdocs.magento.com/guides/v2.2/cloud/release-notes/CloudReleaseNotes2.2.html
https://devdocs.magento.com/guides/v2.3/cloud/release-notes/CloudReleaseNotes2.2.html
https://devdocs.magento.com/guides/v2.2/release-notes/bk-release-notes.html
https://devdocs.magento.com/guides/v2.3/release-notes/bk-release-notes.html
https://devdocs.magento.com/guides/v2.1/cloud/release-notes/cloud-tools.html
https://devdocs.magento.com/guides/v2.2/cloud/release-notes/cloud-tools.html
https://devdocs.magento.com/guides/v2.3/cloud/release-notes/cloud-tools.html

whatsnew
Updated {{ site.data.var.ee}} and {{ site.data.var.ece }} release information to remove obsolete topics and add links to current {{ site.data.var.ece }} release information in [Release Notes for ece-tools](https://devdocs.magento.com/guides/v2.3/cloud/release-notes/cloud-tools.html).

PR #3660, opened 2019-01-30
Quick fix for release notes

hguthrie:
BUG

Quick fix for Cloud release notes. In #3595, looks like page conflict duplicated content. This cleans it up so the release notes are back to normal.


hguthrie:
running tests

PR #3666, opened 2019-01-31
Add new topic names for 2.3.1. magento

nuzil:
## This PR is a:

- [ ] New topic
- [x] Content update
- [ ] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

Topic names generation will changed with 2.3.1 release
https://github.com/magento/async-import/pull/21

This change is only for 2.3, will not be backported to 2.2


magento-cicd2:

PR #3674, opened 2019-02-01
Fix the $ dissapearing from the `bash` code blocks

belbiy:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [X] Bug fix or improvement

## Summary

When this pull request is merged, it will add back the `$` symbol to the bash code blocks.
Ideally, this code should live in the theme, and after we move all the code blocks to backticks markdown syntax, i will move it.

## Additional information

List all affected URLs 

- all with the `bash` code block


belbiy:
running tests

PR #3680, opened 2019-02-03
I feel the text needs to be highlighted

dani97:
I have no clue about declarative schema . When I went through the page I could not understand what the page tries to convey as I skimmed through this para as it is not highlighted. So I suggest this needs to be highlighted.

## This PR is a:

- [ ] New topic
- [ ] Content update
- [X] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

It will improve the readability.

## Additional information

List all affected URLs 

- ...
- ...


magento-cicd2:

danidelcar:
Hello @dani97 . It doesn’t seem like this change is valid. We normally only bold text that is clickable in the UI, or similar. I’m going to reject this change for now, but if you see a better way to make this text clearer to the reader please submit a new PR. Thanks!

PR #3685, opened 2019-02-04
module-load-order.md #3649

### LINKED ISSUE #3649: The example in module load order is not correct
# Bug report

The first example given to explain module load order is not correct. I think component A and component B should be switched in the example.

The current documentation is as follows:

> Component A introduces gadgetlayout.xml, which updates block gadgetBlock from component B. In this case, layout files from component A should be loaded before component B, so you should specify that in component B’s <sequence> entry in module .xml. In other words, component B is dependent on component A. That is to say:

### Steps to reproduce

Please visit the following link to see the erroneous example, https://devdocs.magento.com/guides/v2.3/extension-dev-guide/build/module-load-order.html

## Expected result

I think that the correct paragraph should be as follows:
Component A introduces gadgetlayout.xml, which updates block gadgetBlock from component B. In this case, layout files from component B should be loaded before component A, so you should specify that in component A’s <sequence> entry in module .xml. In other words, component A is dependent on component B. That is to say:

## Possible solutions

I have just put the solution above.

## Additional information

I might misunderstood the example. What I understand is the following:
There are two components called **Component A** and **Component B**. **Component B** has a block defined in it. That block is called **gadgetBlock**. **Component B** has no idea about **Component A**. But, **Component A** would like to update the block **gadgetBlock**, which is defined in **Component B** in the first place. So, **Component A** should depend on **Component B**. Because, if there were not any **Component B**, then there would not be any block for **Component A** to update.

irajneeshgupta:
Fixes #3649 

## This PR is a:

- Content fix or rewrite

## Summary

The first example given to explain module load order was not correct. Rectified this example as per #3649 author suggested.

## Additional information

List all affected URLs 

- https://devdocs.magento.com/guides/v2.3/extension-dev-guide/build/module-load-order.html

https://devdocs.magento.com/guides/v2.3/extension-dev-guide/build/module-load-order.html

### DIFF
--- a/guides/v2.1/extension-dev-guide/build/module-load-order.md
+++ b/guides/v2.1/extension-dev-guide/build/module-load-order.md
@@ -23,7 +23,7 @@ If you change the component load order using `<sequence>`, you must regenerate t
 
 Assume you have a component that needs a configuration file from another component:
 
-__Component A__ introduces `gadgetlayout.xml`, which updates block `gadgetBlock` from __component B__. In this case, layout files from __component A__ should be loaded before __component B__, so you should specify that in __component B's__ `<sequence>` entry in {% glossarytooltip c1e4242b-1f1a-44c3-9d72-1d5b1435e142 %}module{% endglossarytooltip %}.xml. In other words, __component B__ is dependent on __component A__. That is to say:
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/devdocs?pullRequest=3685) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/devdocs?pullRequest=3685) before we can accept your contribution.<br/>

magento-cicd2:

jeff-matthews:
Thanks for the contribution @irajneeshgupta, but you must complete the template when opening a PR so that we can efficiently process it.

irajneeshgupta:
> Thanks for the contribution @irajneeshgupta, but you must complete the template when opening a PR so that we can efficiently process it.

@jeff-matthews 
I have added the required field summary in PR template.
Thank you sir.

dobooth:
Thanks for the fix!

PR #3694, opened 2019-02-05
MAGEDOC-3502: Elastic search config

jeff-matthews:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [x] Bug fix or improvement

## Summary

When this pull request is merged, it will remove the reference to Magento 2.2.x in the 2.3.x topic and more precisely identify the supported version of Elasticsearch 5.x, which is 5.2.

## Additional information

List all affected URLs 

- https://devdocs.magento.com/guides/v2.3/config-guide/elasticsearch/es-overview.html

### DIFF
--- a/guides/v2.3/config-guide/elasticsearch/es-overview.md
+++ b/guides/v2.3/config-guide/elasticsearch/es-overview.md
@@ -86,43 +86,39 @@ The tasks discussed in this section require the following:
 
 This section discusses how to install Elasticsearch 5.2 from their repository. You can select a more recent version of 5.x, but 5.2 has been tested.
 
-1.	Log in to your Magento server as a user with `root` privileges.
-2.	_CentOS_: Enter the following commands in the order shown:
+1. Log in to your Magento server as a user with `root` privileges.
+1. _CentOS_: Enter the following commands in the order shown:
 
-		rpm --import https://artifacts.elastic.co/GPG-KEY-elasticsearch
-		vim /etc/yum.repos.d/Elasticsearch.repo
-
-    Add the following:
-
-    ```
-    [elasticsearch-5.2]
-    name=Elasticsearch repository for 5.2 packages
-    baseurl=https://artifacts.elastic.co/packages/5.2/yum
-    gpgcheck=1
-    gpgkey=https://artifacts.elastic.co/GPG-KEY-elasticsearch
-    enabled=1
-    autorefresh=1
-    type=rpm-md
+    ```bash
+    sudo yum install java-1.8.0-openjdk
     ```
 
-    Enter the following commands:
+    ```bash
+    curl -L -O https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-5.2.0.rpm
+    ```
+  
+    ```bash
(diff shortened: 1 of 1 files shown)


hguthrie:
I bypassed all instinct to address markdown preferences. ;P

jeff-matthews:
Haha! Thank you.

jeff-matthews:
running tests

keharper:
I have some concerns about these changes.

Version 5.2 is rather old. When I made significant updates to this topic about a year ago, there were complaints about having to install such an ancient version. Development said the installation procedures should work for any version of ES 5 or 6. "Should work"  isn't good enough, so I listed version 5.2, the most recent version tested. I added this statement as a compromise to help satisfy those complaints:

`This section discusses how to install Elasticsearch 5.2 from their repository. You can select a more recent version of 5.x, but 5.2 has been tested.`

At minimum, the "5.x" in the preceding paragraph should remain. "A more recent version of 5.2 implies 5.2.1, 5.2.2, etc.

The code sample came from here:

https://www.elastic.co/guide/en/elasticsearch/reference/5.2/rpm.html

I don't know if the `X` in the `baseurl` parameter can simply be replaced with a `2`. That would need to be tested.

Also, this image shows the corresponding attribute in Admin. Note that the selected value of the Search Engine is 5.0+. Development really believes that any version of 5 or 6 should work. ;) But basically, this is another reason why devdocs says 5.X.

https://docs.magento.com/m2/ee/user_guide/Resources/Images/config-catalog-catalog-search-elasticsearch-1_thumb_0_0.png

jeff-matthews:
Thanks @keharper! I'll revert the line about using a more recent version of 5.x as you mention. Then, I'll test the installation commands to see if they work by adding the "2".

guides/v2.3/config-guide/elasticsearch/es-overview.md
```diff
@@ -14,29 +14,28 @@ functional_areas:
...
+* Supports real-time data and analysis
+* Can be used as a document-oriented data store
+* Applications in framework beyond search, including reporting, personalization, performance, and storage
 
 ### Supported versions {#es-spt-versions}
-{{site.data.var.ee}} version 2.2.x supports the following Elasticsearch versions:
 
-*	Elasticsearch [5.x](https://www.elastic.co/downloads/past-releases/elasticsearch-5-2-2){:target="_blank"}
-*	Elasticsearch [2.x](https://www.elastic.co/downloads/past-releases/elasticsearch-2-4-5){:target="_blank"}
+{{site.data.var.ee}} 2.3.x supports the following Elasticsearch versions:
```
keharper: Change the variable to `ce`. As of 2.3, Open Source supports ES

keharper: Actually, get rid of those `target="_blank"`s while you're there.

jeff-matthews: Done. Thanks @keharper!

jeff-matthews:
running tests

PR #3695, opened 2019-02-05
MAGEDOC-3450: Page builder config

jeff-matthews:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [x] Bug fix or improvement

## Summary

- Add a scope for `page-builder/` to the `_config.yml` file so that we can start building and publishing Page Builder docs using the `master` branch of devdocs.
- Exclude all non-multi-repo doc directories so we can avoid needlessly processing files with the `jekyll-relative-links` plugin, which minimizes build time.
- Add the `jekyll-relative-links` and `jekyll-optional-front-matter` gems, which are required to build multi-repo docs.
 - Update `Gemfile.lock` with new dependencies

We will need a separate PR later to add the PB docs to the top nav (`_data/main-nav.yml`).


jeff-matthews:
I've been doing some testing, and I think relative includes are going to be a problem for our multi-repo approach. I know that `mftf/` docs don't exist in a separate repo right now, but that's probably going to change when we realize our ultimate goal.

I think we need to convert relative includes to regular includes and handle them just like the TOC YAML files for page-builder. That means adding a build automation step to copy includes from one repo to another. I'm open to ideas, but I don't think there's a workaround here.

jeff-matthews:
@dshevtsov, the problem doesn't seem to be the include itself. It seems to be this link in the include:

> [Find out your version](./introduction.html#find-version)

jeff-matthews:
I don't think includes are that confusing actually. I'd like to be able to use them because they're a powerful tool for single-sourcing content. I'm also reluctant to ignore links in includes. By updating the link syntax, I was able to eliminate the link check errors.

jeff-matthews:
Do we need to discuss this further or can we merge this?

dshevtsov:
This can work as a workaround for this case, because the affected files won't go to the code repo.
In general, we will need to use a link format that will work on GitHub, which is a link relative to current page and pointing to `.md` file.

Maybe it will make sense to consider moving page-builder configuration (like scope) in its config in future.

jeff-matthews:
running tests

jeff-matthews:
Build time doubled!!! I don't want to merge this without more research into how these changes are affecting build time.

dshevtsov:
Each page now is processed by added plugins.

jeff-matthews:
It's the `jekyll-relative-links` plugin. 

```Generating: JekyllRelativeLinks::Generator finished in 854.932448895 seconds.```

That's adding ~14 minutes to the build time.

jeff-matthews:
When https://github.com/benbalter/jekyll-relative-links/pull/50 is merged, we'll be able to limit the plugin to processing the page-builder directory only and get build times back under control.

jeff-matthews:
running tests

jeff-matthews:
running tests

PR #3753, opened 2019-02-15
Replace github-pages gem with jekyll gems

dshevtsov:
## This PR is a:

Improvement

## Summary

When this pull request is merged, it will replace the github-pages gem with only the gems that we need to build the site. It will allow to break dependency on the github-pages gem and to use latest releases of Jekyll and its plugins.

## Additional details

Internal preview for branch with the changes is available on internal staging environment.


dshevtsov:
running tests

PR #3760, opened 2019-02-16
Correct description of renaming a table

Vinai:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [x] Bug fix or improvement

## Summary

The previous description of `migrateDataFromAnotherTable` was incorrect.  
Also added a small correction about disabling modules - that isn't done in the admin console, but rather using `bin/magento module:disable` to record the change in the `config.php` file.

## Additional information

List all affected URLs:
I'm only aware of
https://devdocs.magento.com/guides/v2.3/extension-dev-guide/declarative-schema/db-schema.html


magento-cicd2:

jcalcaben:
running tests

Vinai:
Hi, I'd like to take the survey, but in addition to allowing the app to read my user email, unfortunately I'd have to give it read and write access to all my organization repositories, and I don't want to do that.
Maybe you can update the app so it requests less privileges via oauth?

sidolov:
Hi @Vinai , we take a look on this, thank you for the input!

PR #3813, opened 2019-02-22
MAGEDOC-3133: Expand Algolia index coverage

jeff-matthews:
## This PR is a:

- [ ] Bug fix or improvement

## Summary

When this pull request is merged, it will configure Algolia to start indexing code samples and other HTML elements that were previously not being indexed. 

We upgraded our Algolia plan, which increases the number of records we can include in our index to 1M.


_config.yml
```diff
@@ -178,6 +178,10 @@ algolia:
     - guides/m1x/**/*.md
     - swagger
     - redoc
+  # Index HTML elements, especially code samples.
+  nodes_to_index: 'p,li,td,code'
+  # Indexing fails if this isn't set because some of our records are very large.
+  max_record_size: 400000
```
belbiy: Good to have this, but we need to be careful not to add large chunks of code into the index with this somehow. Like have a rule in that beyond N number of lines the code needs to be either split into manageable chunks or referenced somehow. What do you think, @jeff-matthews ?

jeff-matthews: That's a good idea @belbiy. I think that the largest file size is currently `/guides/v2.2/rest/tutorials/orders/order-create-order.html` at `346.85 Kb`, which is caused by a massive API sample response in one of our tutorials: https://devdocs.magento.com/guides/v2.2/rest/tutorials/orders/order-create-order.html#review-order

Are you aware of any custom options for selectively excluding specific elements from the index?  I know that we can ignore certain _files_, but I think that's a little heavy-handed.

jeff-matthews:
running tests

_config.yml
belbiy: Algolia have hooks that we can use when each record gets processed https://community.algolia.com/jekyll-algolia/hooks.html#before-indexing-each

We can potentially use that to exclude certain code snippets by adding some sort of class name like `.no-index` and make the plugin that will skip that record.

jeff-matthews: Nice!

PR #3853, opened 2019-03-02
MAGEDOC-3553: Make 'Edit this page on GitHub' work in docs from subrepos

dshevtsov:
## This PR is a:

Bug fix

## Summary

When this pull request is merged, it will add support of 'Edit this page on GitHub' to docs from sub repositories such as `mbi`.

## Additional information

Added the github-path plugin that generates `page.github_path` parameter.
Enabled the **Give us feedback** link in the MBI docs.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -69,7 +69,7 @@ defaults:
       path: mbi
     values:
       group: mbi
-      github_link: false
+      github_files: https://github.com/magento/devdocs-mbi/blob/master/
       feedback_link: false
 
   -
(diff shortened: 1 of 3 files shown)


dshevtsov:
@jeff-matthews, good point.
I had doubts about enabling **Give us feedback** this way, because the generated title of the issue points to path inside devdocs, not inside _devdocs-mbi_
But now I see cases how we can use that, so I'll add it to this PR.
screen shot 2019-03-04 at 09 29 07

dshevtsov:
running tests

PR #3861, opened 2019-03-04
Fix command for running all test

slackerzz:
Since we are in `dev/tests/acceptance` we need to prepend `../../../` to the command or we will get an error

## This PR is a:

- [ ] New topic
- [ ] Content update
- [x] Content fix or rewrite
- [x] Bug fix or improvement

## Summary

## Additional information

List all affected URLs 

- https://devdocs.magento.com/mftf/2.3/getting-started.html#prepare-environment


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/devdocs?pullRequest=3861) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/devdocs?pullRequest=3861) before we can accept your contribution.<br/>

magento-cicd2:

osrecio:
Hi @slackerzz . I checked your change... but some lines below we have:

```
vendor/bin/mftf run:test AdminLoginTest --remove
```
should be changed too?

Thanks!

slackerzz:
Hi @osrecio, 
```
vendor/bin/mftf run:test AdminLoginTest --remove
```
can be launched from magento root dir, this is just an example of how run a single test with "cleanup".

Instead the line i changed is the last of a sequence of commands that comes just after a `cd` shell command.

dobooth:
Hi @slackerzz 
Thanks for this fix.
We just (a few minutes ago) moved the MFTF docs to the MFTF repo.
That means this PR won't work any more as the folder structure is different.
But to the fix, in the docs, we are trying to be consistent with commands. 
We want to assume that they are running commands at the Magento root, which would negate the need for ../../../. See the note here: https://devdocs.magento.com/mftf/docs/commands/codeception.html
If we still need a fix, it would explaining that these commands are run from the Magento root, rather than changing the command paths.
Any new MFTF PRs can be logged here: https://github.com/magento/magento2-functional-testing-framework
I will have to close this PR since it will fail now. Sorry for the delay and thanks for the help.

PR #3871, opened 2019-03-05

paliarush:
## This PR is a:

-  Content fix or rewrite

## Summary


magento-cicd2:

guides/v2.2/coding-standards/technical-guidelines.md
```diff
@@ -496,7 +496,7 @@ class View extends Template
 
 5.16. If a method uses system resources (such as files, sockets, streams, etc.), the code MUST be wrapped with a `try` block and the corresponding `finally` block. In the `finally` sections, all resources SHOULD be properly released.
 
-5.17. `LocalizedException` SHOULD only be thrown in the Presentation layer (Controllers, Blocks).
+5.17. Exceptions which need to be displayed to the user MUST be sub-types of `LocalizedException`. Any other types of exceptions MUST be wrapped with `LocalizedException` before being displayed to the user.
```
melnikovi: I think we also want to use LocalizedException only in presentation layer, after change it's not clear.

paliarush: We use it in service contracts, which are directly exposed as web API, what is the alternative proposal for service contracts?

melnikovi: We can convert service contract exceptions to LocalizedException before displaying to the user. Can we keep somehow part that LocalizedException should be only be thrown in presentation layer.

paliarush: What will be the conversion rule? We have a hierarchy of LocalizedException classes. Where will the translation be done?

Also this will be backward incompatible change to @api since most service contracts have LocalizedException declared via @throws

dobooth:
running tests

dshevtsov:
@dobooth please add `whatsnew`

PR #3883, opened 2019-03-06
Use build number instead of branch name on staging

dshevtsov:
## This PR is a:

- Improvement

## Summary

When this pull request is merged, it will update link checker to use a build number instead of branch name for a baseurl to test on staging.

### DIFF
--- a/rakelib/lib/link-checker.rb
+++ b/rakelib/lib/link-checker.rb
@@ -11,7 +11,7 @@ module LinkChecker
   # Ignore baseurl if $branch is available in environment
   def self.options
     config = YAML.load_file('_config.checks.yml')
-    baseurl = ENV['branch']
+    baseurl = ENV['BUILD_NUMBER']
     return config['html-proofer'] unless baseurl
     url_swap = { url_swap: { %r{\A/#{baseurl}} => '' } }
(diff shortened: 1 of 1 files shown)


jcalcaben:
If someone creates a staging build on a PR and that PR gets updated and built again. Does that mean there exists 2 versions of the branch but different build numbers?

jcalcaben:
What is the cleanup logic in the AWS staging server for all these builds? Some writers do a lot of iterations and builds before final publication. I'm concerned with hitting the server storage limit with this approach.

dshevtsov:
@jcalcaben yes
both previews will be available

jeff-matthews:
@jcalcaben, since it's AWS S3, there are no storage limits, per se. But you raise a valid concern about usage since we get charged not only by API operations but also by how much storage we're using.  That's why I asked the Platforms Ops team to configure a 14-day lifecycle policy on the staging bucket when setting up these environments months ago; so we can purge old builds automatically.

I admit that I haven't verified that builds are being purged per the policy, so if we discover that old builds are _not_ being purged, I'll contact Platform Ops to fix it.

dshevtsov:
running tests

PR #3941, opened 2019-03-16
Add a plugin to generate page.versions

dshevtsov:
## This PR is a:

- Improvement

## Summary

When this pull request is merged, it will add a plugin that generates `page.versions` which contains an array of URLs of the topic in all available versions on devdocs. It is aimed to make the version switcher to use only existing topics.


dshevtsov:
running tests

PR #3966, opened 2019-03-19
Update created vanilla Magento instance name.

likemusic:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [ ] Content fix or rewrite
- [x] Bug fix or improvement

## Summary

When this pull request is merged, it will show actual vanilla magento instance name pattern.

## Additional information

List all affected URLs 

- https://devdocs.magento.com/guides/v2.2/contributor-guide/contributing.html

But seems that files for urls below should be updated also:
- https://devdocs.magento.com/guides/v2.1/contributor-guide/contributing.html
- https://devdocs.magento.com/guides/v2.3/contributor-guide/contributing.html

Seems that the other urls for magento instanses should be updated too.

And btw, why common blocks from pages by urls above not separated into common included blocks?

### DIFF
--- a/guides/v2.1/contributor-guide/contributing.md
+++ b/guides/v2.1/contributor-guide/contributing.md
@@ -154,7 +154,7 @@ For `version`, the currently supported values are [version tags](https://github.
 
 **Admin access:**
 
-- http://i-xxx.engcom.dev.magento.com/admin
+- https://pr-{$id_pr}.instances.magento-community.engineering/admin
 - Admin Credentials:
     - Username: admin
     - Password: 123123q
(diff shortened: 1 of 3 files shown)


devops-devdocs:

jeff-matthews:
@likemusic:

> And btw, why common blocks from pages by urls above not separated into common included blocks?

That's a good question. It would be interesting to run a diff on all three versions of the file and identify all common text or "blocks" as you put it. If there's a lot of sections of text that are the same, we can review creating include files for those sections.

osrecio:
@jeff-matthews Done!

PR #3934, opened 2019-03-15
Fix wrong path and update path of indexing

mageprince:
## This PR is a:

- Content fix or rewrite

## Summary

When this pull request is merged, it will fix and add proper configuration path.

PR #3972, opened 2019-03-20
Update wrong configuration path

mageprince:
## This PR is a:

- Content fix or rewrite

## Summary

When this pull request is merged, it will update the wrong configuration path as per @dobooth suggestion: https://github.com/magento/devdocs/pull/3934#issuecomment-473296471


devops-devdocs:

PR #4069, opened 2019-03-27
Updated example of logical AND and OR search

mage2pratik:
## This PR is a:

- [ ] New topic
- [ ] Content update
- [x] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

When this pull request is merged, it will updated example of logical AND and OR search.
Currently display wrong result(The query returns 4 items). So, I've updated example.

## Additional information

List all affected URLs 

- https://devdocs.magento.com/guides/v2.3/graphql/queries.html


devops-devdocs:

keharper:
Hmm. The original example used to work. I've tested that your example works, but I want to understand why the original no longer works before I merge this PR.

mage2pratik:
@keharper I'm not sure but may be sample data has been updated.

keharper:
I can't figure out why the old query failed, but I've done my due diligence. Thanks for the updates, @mage2pratik

PR #4142, opened 2019-04-02
Document GET request support

keharper:
## This PR is a:

- [ ] New topic
- [x] Content update
- [ ] Content fix or rewrite
- [ ] Bug fix or improvement

## Summary

When this pull request is merged, it will add documentation describing HTTP GET support for GraphQL queries

## Additional information

List all affected URLs 
https://devdocs.magento.com/guides/v2.3/graphql/send-request.html


guides/v2.3/graphql/send-request.md
```diff
@@ -3,7 +3,25 @@ group: graphql
 title: GraphQL requests
 ---
 
-In Magento 2.3.1, you must send all queries and mutations as HTTP POST requests. 
+Magento GraphQL supports the HTTP GET and POST methods. You can send a query as a GET or POST request. GET requests can be cached, but Magento does not support cacheing POST requests. Mutations must be POST requests.
```
danielrenaud: Maybe we should remove the mention of caching. It gives the ability to cache, but people might take this to me that Magento does some caching of GET requests.

keharper: This won't be published on devdocs until 2.3.2. I reworded the passage to 

GET requests can be cached, although Magento does not currently support this feature.

When we support Varnish cacheing, I'll update this passage and include a link to any Varnish documentation that needs to be written in the future.

guides/v2.3/graphql/send-request.md
```diff
@@ -3,7 +3,25 @@ group: graphql
 title: GraphQL requests
 ---
 
-In Magento 2.3.1, you must send all queries and mutations as HTTP POST requests. 
+Magento GraphQL supports the HTTP GET and POST methods. You can send a query as a GET or POST request. GET requests can be cached, although Magento does not currently support this feature. Mutations must be POST requests. 
+
+You can optionally send a GET query request in a URL. In these requests, you must specify `query` as the query string. You might need to encode the query.  
+
+`http://<magento_host>/graphql?query=%7Bproducts(filter%3A%7Bsku%3A%7Blike%3A%2224-WB%25%22%7D%7D)%7Bitems%7Bsku%7D%7D%7D`
```
erikmarr:
```suggestion
`http://<host>/graphql?query=%7Bproducts(filter%3A%7Bsku%3A%7Blike%3A%2224-WB%25%22%7D%7D)%7Bitems%7Bsku%7D%7D%7D`
```
our rest documentation just uses `host` at the beginning of the request URL. Do GraphQL URL requests need to be different?

guides/v2.3/graphql/send-request.md
```diff
@@ -3,7 +3,25 @@ group: graphql
 title: GraphQL requests
 ---
 
-In Magento 2.3.1, you must send all queries and mutations as HTTP POST requests. 
+Magento GraphQL supports the HTTP GET and POST methods. You can send a query as a GET or POST request. GET requests can be cached, although Magento does not currently support this feature. Mutations must be POST requests. 
+
+You can optionally send a GET query request in a URL. In these requests, you must specify `query` as the query string. You might need to encode the query.
```
erikmarr: Is it obvious when a user would choose to use a URL vs mutli-line query?

erikmarr: Would I need to encode the query if there are special characters?

PR #4162, opened 2019-04-03
Fix unclear layer names in Arch guide

### LINKED ISSUE #3295: Architecture layers names are switched the whole time
# Topic clarification/correction request

Consistent naming of the layers is something that will help understanding this topic

## Topic Link

https://devdocs.magento.com/guides/v2.2/architecture/archi_perspectives/present_layer.html
https://devdocs.magento.com/guides/v2.2/architecture/archi_perspectives/service_layer.html

## Description

Initially you have 4 layers : 
- Presentation layer
- Service layer
- Domain layer
- Persistence layer

So far so good!

But in Present layer it seems that there is a 5th layer Product layers and Domain layer is now also Business logic layer. This term is used before it is explained that it is the same thing later on.

In the Service layer you suddenly have the model layer which i think is also the domain layer since it has domain logic. So now we have 3 names used for the same layer. 

What the product layer is is still to be guessed.
Since it is talking about underlying product layers we can guess that this is the Service layer even though it does not make any sense to call it a product layer

## Possible solutions

Keep the same terminology that you are explaining.

## Additional information

Do not get me wrong here please. I love the devdocs but i feel this should be improved on

dobooth:
## This PR is a:

- Content fix or rewrite

## Summary

Update confusing layer names within the Arch guide.
Closes #3295 

## Additional information

List all affected URLs 

https://devdocs.magento.com/guides/v2.3/architecture/archi_perspectives/service_layer.html
https://devdocs.magento.com/guides/v2.3/architecture/archi_perspectives/present_layer.html


dobooth:
@paliarush Can you please review?

PR #4220, opened 2019-04-11
Remove file-based redirect

jeff-matthews:
## Scope of this PR

- Site improvements

## Purpose of this PR (required)

Remove the file-based redirect for the Product release page from the Jekyll project and add a routing rule condition to the S3 configuration instead to handle the redirect "server side".

Here's the snippet I added to the AWS S3 config:

```
{
        "Condition": {
           "KeyPrefixEquals": "availability"
         },
         "Redirect": {
           "HttpRedirectCode": "301",
           "ReplaceKeyPrefixWith": "release"
         }
      }
```

I tested on staging first (build 291) and the S3 redirect is working as expected.

TODO:

- [ ] Remove the following line from the deploy stage in the production pipeline
  ```sh 'mv _site/availability.html _site/availability && aws s3 cp _site/availability s3://devdocs.magento.com/availability --content-type text/html'```

## Affected URLs (required)

- https://devdocs.magento.com/availability


dshevtsov:
@jeff-matthews we have a convention to not publish links to internal resources.

jeff-matthews:
Yes, that's true @dshevtsov, but we're not 100% consistent since some team members tweet links to staging and we provide links to staging for EAP participants.

dshevtsov:
I think we shouldn't use internal links in any public channels.
I guess, we use separate channels to communicate with EAP participants and the participants have access to staging.

jeff-matthews:
Our staging environment is not restricted, so technically, everyone has access.

dshevtsov:
I didn't know that links to https://devdocs.magedevteam.com/ were available for anyone.
Just checked from a smart phone. It means that the preview links are public resources, not internal.
Thank you for the information.

jeff-matthews:
running tests

PR #4237, opened 2019-04-12
MAGECLOUD-3205: Provide ElasticsSuite auto-configuration in ECE-Tools

shiftedreality:
## Scope of this PR

- __Major Update__: Significant new info: new section in existing topic, etc.

## Purpose of this PR

https://magento2.atlassian.net/browse/MAGECLOUD-3205

### DIFF
--- a/guides/v2.2/cloud/env/variables-deploy.md
+++ b/guides/v2.2/cloud/env/variables-deploy.md
@@ -398,6 +398,45 @@ stage:
       _merge: true
 ```
 
+### `ELASTICSUITE_CONFIGURATION`
+
+-  **Default**—_Not set_
+-  **Version**—Magento 2.2.0 and later
+
+Use this variable to retain customized ElasticSuite service setting between deployments.
+All configs provided in this configuration are nested in main ElasticSuite config section 'system/default/smile_elasticsuite_core_base_settings'    
+
+```yaml
+stage:
+  deploy:
+    ELASTICSUITE_CONFIGURATION:
+      es_client:
+        servers: 'remote-host:9200'
+      indices_settings:
+        number_of_shards: 1
+        number_of_replicas: 0      
+```
+
+{% include cloud/merge-configuration.md %}
+
+The following example merges a new value to the existing configuration:
+
+```yaml
+stage:
+deploy:
+  ELASTICSUITE_CONFIGURATION:
+      indices_settings:
+        number_of_shards: 3
+        number_of_replicas: 3
+      _merge: true
+```
+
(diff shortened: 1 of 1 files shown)


devops-devdocs:

guides/v2.2/cloud/env/variables-deploy.md
```diff
@@ -398,6 +398,45 @@ stage:
...
+        number_of_replicas: 3
+      _merge: true
+```
+
+#### Known limitations
+There are some known issues you should be aware of once `ElasticSuite` module is installed:
+
+* Changing `search engine` to anything unlike `elasticsuite` would cause failed deploy with appropriate validation error
+* Removing `ElasticSearch` service would cause failed deploy with appropriate validation error
+
```
NadiyaS: Should not we mention somewhere how to add ElasticSuite with link to its documentation? and once you added this package to the project - your search engine is changed to elasticsuite automatically (and on cloud) and you should uninstall elasticsuite if wants to switch to another engine

hguthrie: DevDocs have been told to state specifically that we do not support ElasticSuite. Has this changed?

hguthrie: See here we even call it out: https://devdocs.magento.com/guides/v2.2/cloud/project/project-conf-files_services-elastic.html#add-elasticsearch-plugins

Please let me know if that has changed

shiftedreality: Yes, we added functionality to correctly set up this module

jeff-matthews: _**If**_ Magento's support position on Elasticsuite _**has**_ changed, we'll also need an update here: https://devdocs.magento.com/guides/v2.3/cloud/project/project-conf-files_services-elastic.html#elasticsearch-plugins.

shiftedreality: Maybe we can call it *limited support*, need advise from @YPyltiai

jeff-matthews: Seems like we should also run this by someone in Support.

YPyltiai: We **do not** Support ElasticSuite module itself as well as issues in it. We **do** configure it properly for Cloud, though which is the only thing we "Support". Being short - as long as the configuration of module and connection to ES servers are set up correctly - our support for this module ends at this point.

PR #4260, opened 2019-04-16
GraphQL cache tagging

keharper:
This PR adds documentation for https://github.com/magento/graphql-ce/issues/230


guides/v2.3/graphql/develop/resolvers.md
```diff
@@ -209,3 +209,42 @@ A sample response:
   }
 }
 ```
+
+## Identity resolvers
```
danielrenaud: We changed the naming of this to avoid using "Resolver" so we don't confuse people that this is some kind of resolver. So maybe a new title could be "Identity interfaces" and update things according to the changes in this commit: https://github.com/magento/magento2ce/commit/962ad5e055717c22af8c8f6414d49bc178aaaee3

keharper:
Superceeded by #4402

PR #4295, opened 2019-04-19
Added note about using ">=2.3.1 <2.3.2" syntax for composer

markshust:
## Purpose of this PR

The format of the composer package `magento/magento-cloud-metapackage` version of `">=2.3.1 <2.3.2"` is necessary to ensure the ECE tool is properly updated. There should be a note in the readme that calls this out.


devops-devdocs:

rogyar:
Hi @markshust. Thank you for your contribution. That's a really good point. Please check my review comments for minor adjustments

guides/v2.3/cloud/project/project-upgrade.md
```diff
@@ -27,6 +27,8 @@ If you use PHP version 7.2, you must remove the `mcrypt` extension from the [`ex
     ```bash
     composer require magento/magento-cloud-metapackage <requiredversion> --no-update
     ```
+    
+    > Be sure to use the `">=2.3.1 <2.3.2"` syntax for the value of `<requiredversion>`. This is necessary, as failing to do so will not properly update the ECE Tool.
```
rogyar: Please, rename "ECE Tool" to "ECE Tools".

guides/v2.3/cloud/project/project-upgrade.md
```diff
@@ -27,6 +27,8 @@ If you use PHP version 7.2, you must remove the `mcrypt` extension from the [`ex
     ```bash
     composer require magento/magento-cloud-metapackage <requiredversion> --no-update
```
hguthrie:
```suggestion
    composer require "magento/magento-cloud-metapackage":">=2.3.1 <2.3.2" --no-update
```

guides/v2.3/cloud/project/project-upgrade.md
hguthrie: I can understand why you would want this note here.  I recommend removing that note. I updated the step instead. I also suggested providing the actual command, so it is easy for others to copy.

markshust: i think the note is needed -- it's important to know why you would need to do this rather than blindly copying/pasting something. if this note wasn't there i would most likely just use the single version method.

hguthrie:
@rogyar Can you give this another look and update your review?

hguthrie:
running tests

PR #4298, opened 2019-04-19
Fix broken link for ModSecurity rules

hguthrie:
This PR fixes a broken link to the ModSecurity Rules for [Fastly WAF topic](https://devdocs.magento.com/guides/v2.3/cloud/cdn/fastly-waf-service.html#how-it-works).

I think this is the appropriate link, since the Trustwave site links to it.


hguthrie:
@dobooth this one is ready for writer clearance to merge with small changes, so adding you.

PR #4323, opened 2019-04-23
Updated contributor guide

akaplya:


devops-devdocs:

osrecio:
Hi @akaplya Can you update the [PR Template](https://github.com/magento/devdocs/blob/master/.github/PULL_REQUEST_TEMPLATE.md) to fill it

guides/v2.1/contributor-guide/contributing_dod.md
```diff
@@ -166,17 +166,19 @@ The reviewer should check whether the task meets the original acceptance criteri
 * All backward-incompatible changes on code level are approved by an architect. The following cases currently cause SVC failures but do NOT require architect approval:
   * Removal of private constant
   * Override of the method defined in parent class
-  * Adding optional parameter to the constructor, except for the classes explicitly meant for extension, including the following ones:
+  * Adding optional parameter to the constructor, except for the classes explicitly meant for extension, including the following ones and their children:
```
kandy: I don't think that we need to add this "and their children", it's too strict. Can you provide a reason for that?

akaplya: For the most scenarios that we have in models, resource models and blocks the inheritance is the only way how a developer can modify behaviour. So, we have to expect that for instance the Product model will be inherited. This is a legal operation, so we have to protect extensions which are using this approach.

kandy: There's no reason to extends from product model, and we also have an interface for it, so don't think is a good idea to not allow to improve product model class

akaplya: We do not forbid extend models, so we have to protect extension that did it. Anyway if you believe that nobody needs to extend models nobody will be negatively affected by this rule.

joni-jones: I agree with @kandy, we should not disable a possibility to modify children constructors of mentioned classes. It reduces refactoring possibilities as well.

devops-devdocs:

dobooth:
@akaplya @buskamuza  Is this ready to go?

dobooth:
Hi @akaplya We recently dropped support for v2.1, so we cannot accept PRs that hit that branch. Can you please move these changes to v2.2?

buskamuza:
I believe we agreed to the previous (existing) statement and these changes are not needed anymore.

paliarush:
This PR is outdated and MUST NOT be merged.

PR #4369, opened 2019-04-27
Update index.md

ryantfowler:
## Purpose of this PR

I updated the file suggested to edit to enable/disable modules in the way shown from `composer.json` to `config.php` which is the accurate file for this method of disabling modules.

## Affected URLs

- https://devdocs.magento.com/extensions/inventory-management/


devops-devdocs:

PR #4370, opened 2019-04-28
Removed M1 class path in the documentation

maximbaibakov:
## Purpose of this PR

Removed M1 class path in the documentation

## Affected URLs

- https://devdocs.magento.com/guides/v2.3/extension-dev-guide/build/di-xml-file.html


devops-devdocs:

PR #4389, opened 2019-05-01
Corrected magnifier option name

yogeshsuhagiya:
## Purpose of this PR

The magnifier option name is `eventType` instead of `hover`. Actually `hover` is the possible value not the property itself.

## Affected URLs

- https://devdocs.magento.com/guides/v2.1/javascript-dev-guide/widgets/widget_gallery_mg.html
- https://devdocs.magento.com/guides/v2.2/javascript-dev-guide/widgets/widget_gallery_mg.html
- https://devdocs.magento.com/guides/v2.3/javascript-dev-guide/widgets/widget_gallery_mg.html

## Links to source code

- https://github.com/magento/magento2/blob/2.1/app/design/frontend/Magento/luma/etc/view.xml#L224

### DIFF
--- a/guides/v2.1/javascript-dev-guide/widgets/widget_gallery_mg.md
+++ b/guides/v2.1/javascript-dev-guide/widgets/widget_gallery_mg.md
@@ -56,7 +56,7 @@ Enable magnifier.
 
 Height of the magnifier block in pixels.
 
-### `hover` {#opt_hover}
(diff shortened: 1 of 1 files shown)


devops-devdocs:

yogeshsuhagiya:
Hi, @ronak2ram Updated URLs in the comment as per your request

PR #4403, opened 2019-05-02
Remove node as a dependency

belbiy:

This pull request (PR) will remove node as a dependency. 
This means that our repo will be more secure. The downside is that we loose CSS/JS minification, which we not need anyway. So no downsides I guess.

## Affected DevDocs pages

All

- ...
- ...


jcalcaben:
What is this in response to? Why not just remove dependencies that are vulnerable?

jeff-matthews:
For vulnerability reference:

```

Known high severity security vulnerability detected in tar < 4.4.2 defined in package-lock.json.
package-lock.json update suggested: tar ~> 4.4.2.

```

belbiy:
@jcalcaben , i've tried fixing the dependencies, but it seems that `node-sass` cannot be fixed. It depends on `node-gyp` and the `tar` package that's affected. It's a known problem and they need to resolve it.

But the point is that we don't really need the fancy node scripts on such a small amount of code. I'm kind of tired of dealing with security vulnerabilities it brings :-)

belbiy:
running tests

PR #4412, opened 2019-05-03
Add font-display CSS rule to Using custom fonts docs

adifucan:

Font-display property is added to CSS.
The font-display descriptor determines how a font face is displayed based on whether and when it is downloaded and ready to use.
Now by default the behavior of font loading remains the same as by default font-display rule is auto. But for Blank theme (and for Luma as it extends our Blank theme) we made font-display: swap
It means that if the font face is not loaded, any element attempting to use it must render a fallback (default or system) font face. If custom (in our case it is Open Sans) font face successfully loads during this period, it is used normally. 

This pull request (PR) ...

## Affected DevDocs pages

- guides/v2.3/frontend-dev-guide/css-topics/using-fonts.md

- https://github.com/magento/magento2ce/pull/4146/files


devops-devdocs:

adifucan:
Thanks @rogyar! You are correct. Changes are also applicable for 2.2.x and 2.1.x I guess. Shall I update a PR?

erikmarr:
Hi @adifucan yes please update your PR so the 2.1 version is updated. Because 2.2.x and 2.3.x are symlinks back to the 2.1 version, you'll only need to make the change there.

adifucan:
@erikmarr update my PR. I made it only for 2.3 because this documentation is only about changes in Magento 2.3

erikmarr:
@rogyar can you please remove or dismiss your comment as @adifucan has indicated this is only for v2.3.x?

erikmarr:
running tests

erikmarr:
running tests

erikmarr:
Thanks @adifucan!

PR #4424, opened 2019-05-04
Update queries.md

sanjay-wagento:

Previous logic was only for AND condition. If we remove or then also get same result but using my modified query we can get AND OR result with 8 items.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/graphql/queries.html

- N/A


devops-devdocs:

sanjay-wagento:
I am working on this at #dmcdindia19

rogyar:
From the technical standpoint, the provided change is correct.
We need to have logical AND between price and (name OR sku) as it's mentioned in the description.
Currently we have logical AND between price AND name OR sku

keharper:
I updated the example further because this phrase was invalid.

pair of women’s shorts in size 10 (WSH%10%)

There are no women's shorts in Luma in that size. To ensure the search returns meaningful results, I changed the query to search for pants (WP%).

keharper:
running tests

keharper:
Thanks @sanjay-wagento

PR #4425, opened 2019-05-06
Remove rules duplicating method signature info

Vinai:
Remove rules that duplicate information in method signatures.

This pull request (PR) removes the rules enforcing useless phpdoc annotations.
These useless annotations are generally not liked and add no information beyond what already is supplied by the class, interface or method name and the method signature.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.2/coding-standards/docblock-standard-general.html
- https://devdocs.magento.com/guides/v2.3/coding-standards/docblock-standard-general.html

### DIFF
--- a/guides/v2.2/coding-standards/docblock-standard-general.md
+++ b/guides/v2.2/coding-standards/docblock-standard-general.md
@@ -33,7 +33,7 @@ The following is assumed by default:
 The documentation should follow two simple principles:
 
 1. Be as short as possible.
-2. Include all necessary information.
+2. Include all necessary information without duplication.
 
 ### Short documentation
 
@@ -183,7 +183,7 @@ For example: namespace, class, interface, function, property, method, and so on.
 
 If the source code file has one and only one standalone structural element (class, interface, function, and so on), as it may be required by language-specific coding standard, the file DocBlock is to be reused for this element.
 
-So in general case, classes that are declared in dedicated files, must have one DocBlock, which refers to the class and file at the same time.
+So classes that are declared in dedicated files, must have either no DocBlock or exactly one DocBlock, which refers to the class and file at the same time.
 
 **DocBlock for a Class**
 
@@ -257,7 +257,9 @@ require_once __DIR__ . '/../../functions.php';
 ### Classes and interfaces
 {:#classes-interfaces}
 
-Classes and interfaces must have a short description that is a human-understandable intention of the class.
+Classes and interfaces should have a short description that is a human-understandable intention of the class.
+If the short description adds no additional information beyond what the type name already supplies, the
+short description must be omitted.
 
 **Good:**
 
@@ -288,7 +290,7 @@ use Magento\Stdlib\DateTime as StdlibDateTime;
 /**
  * @var Logger
  */
-protected $_logger;
+private $logger;
 
 /**
  * Description of method here.
@@ -297,7 +299,7 @@ protected $_logger;
  * @param StdlibDateTime $dateTime
  * @param int            $number
  */
-protected function doSomething(Random $mathRandom, StdlibDateTime $dateTime, $number)
+private function doSomething(Random $mathRandom, StdlibDateTime $dateTime, $number)
 {
 
 }
@@ -326,32 +328,33 @@ class Profiler
 ### Functions and methods
 {:#functions-methods}
 
-Functions and methods must have:
+In general, typed method signature must be preferred over PHPDoc annotations whenever possible.
 
-* Short description
-* Long description that explains the motivation behind the implementation.
+Functions and methods should have:
+
+* A short description in case it adds meaningful information beyond the method name.
+* If the purpose of the method is not obvious, a long description that explains the motivation behind the implementation.
+  The comment must describe why method is implemented and not how.
   For example:
 
    * If a workaround or hack is implemented, explain why it is necessary and include any other details necessary to understand the algorithm.
    * For non-obvious implementations where the implementation logic is complicated or does not correspond to the Technical Vision or other known best practices, include an explanation in the doc block's description.
      An implementation is non-obvious if another developer has questions about it.
 
-* Declaration of all arguments (if any) using `@param` tag.
-  Appropriate argument type must be specified.
-* Declaration of return type using `@return` tag.
-  If there is no such operator, the `@return` tag must have `void` as the return value.
+* The declaration of all arguments (if any) using `@param` tag, unless the argument type is indicated in the method signature.
+  All `@param` annotations must include the appropriate argument type.
+  If any argument requires a `@param` annotation, all arguments must be listed (all or none).  
+  The `@param` annotations must be in the same order as the method arguments.
+* The declaration of the return type using the `@return` tag must only be added if the method return type signature
+  does not supply all necessary information (see below for more information on return types).
 * Declaration of possibly thrown exception using `@throws` tag, if the actual body of function triggers throwing an exception.
-  All occurrences of `@throws` in a DocBlock must be after `@param` and `@return` (if any).
+  All occurrences of `@throws` in a DocBlock must be after `@param` and `@return` annotations.
 
 **Exceptions to these rules:**
 
-* Constructors may not have short and/or long description
+* Testing methods in Unit tests may have doc blocks to describe the purpose of the test, for example referencing github issues.
 
-* Testing methods in Unit tests may not have doc blocks if the test's method name follows the convention (test<MethodName>)
-
-   * If the test does not follow the convention, it should have a doc block describing the covered methods
-
-   * Non-testing methods should have a doc block with description. It includes data providers and any helper methods
+* Test method annotations may include data providers and other testing annotations.
 
 #### Things to include
 
@@ -364,7 +367,7 @@ Functions and methods must have:
     For example: `@return Config|null`.
     The DockBlock needs to explain what situations return `null`.
 
-    Another example: `@param FileInterface | null`.
+    Another example: `@param FileInterface|null`.
     The DocBlock needs to explain what happens when the value of the parameter is `null`.
 
     Ideally, implementations such as these should be avoided.
@@ -425,7 +428,7 @@ In general, use the `@throws` tag when the code uses *throw*:
  *
  * @param  string $elementId
  * @param  string $attribute
- * @param  mixed  $value
+ * @param  int|string|float|bool|object|null  $value
  * @return self
  * @throws \InvalidArgumentException
  */
@@ -492,25 +495,51 @@ public function deleteDirectory($path)
 #### @return tag
 {:#return}
 
-If there is no explicit return statement in a method or function, a `@return void` should be used in the documentation.
+In general method return type signatures should be prefered over `@return` type annotations.
+If that is not possible due to ambiguous return types or because of backward compatibility constraints, the `@return` type annotation must be used.
 
-If the method returns itself, `return $this` should be used.
+If there is no explicit return statement in a method or function or a return statement without a value, a `void` return type must be declared in the method signature. For example:
+
+```php
+function setName(string $name): void
+{
+   $this->name = $name;
+}
+```
+
+If the method returns itself, the method signature return type must be `self`. Here is an example:
+
+```php
+function withField(string $fieldName): self
+{
+   $this->fields[] = $fieldName;
+   return $this;
+}
+```
+
+If for backward compatibility reasons no return type can be added to the method signature, a `@return $this` annotation must be used.
+
 
 ### Constants
 {:#constants}
 
-Constants must have short description.
+Constants may have a short description.
+If the short description adds no additional information beyond what the constant name already supplies, the
+short description must be omitted.
 
 For example, a global constant:
 
 
 ```php
 /**
- * Directory separator shorthand
+ * Directory separator shorthand, intended to make code more readable.
  */
 define('DS', DIRECTORY_SEPARATOR);
+```
 
 Or constants in a class:
+
+```php
 class Profiler
 {
     /**
@@ -530,7 +559,7 @@ It's encouraged to replace existing DocBlock templates with regular DocBlock com
 ## Structure of documentation space
 {:#documentation-space}
 
-`@category`, `@package`, and `@subpackage` MUST NOT be used.
+`@author` ,`@category`, `@package`, and `@subpackage` MUST NOT be used.
 Documentation is organized with the use of namespaces.
 
 ## Other DocBlock tags
@@ -539,16 +568,12 @@ Documentation is organized with the use of namespaces.
 ### @inheritdoc tag
 {:#inheritdoc}
 
-Whenever possible the `@inheritdoc` tag MUST be used for children to avoid duplication of doc blocks.
-
-Even Though PHPDocumentor understands inheritance and uses the parent doc block by default (without `@inheritdoc` tag specified), including the tag helps ensure that the doc block is not missed at all.
-
-Rules for usage of the tag:
-
-* Use `@inheritdoc` (notice no braces around) to indicate that the entire doc block should be inherited from the parent method.
-* Use the inline `{@inheritdoc}` tag (with braces around) in long descriptions to reuse the parent's long description. The tagged method MUST have its own short description.
+The `@inheritdoc` tag SHOULD NOT be used.
+If a child class method requires a long description to explain its purpose, it may use `@inheritdoc` to indicate the new description is intended as an addition to the parent method description.
+In general such method overrides are a code smell and should be used as an incentive to make the code more self-documenting if possible.
 
 **DocBlock for the Interface**
+
 ```php
 /**
  * Interface for mutable value object for integer values
@@ -556,22 +581,16 @@ Rules for usage of the tag:
 interface MutableInterface
 {
     /**
-     * Get value
-     *
      * Returns 0, if no value is available
-     *
-     * @return int
      */
-    public function getVal();
+    public function getVal(): int;
 
     /**
-     * Set value
-     *
      * Sets 0 in case a non-integer value is passed
-     *
-     * @param int $value
+     * 
+     * @param int|string|bool|float|null $value
      */
-    public function setVal($value);
+    public function setVal($value): void;
 }
 ```
 
@@ -582,19 +601,16 @@ interface MutableInterface
  */
 class LimitedMutableClass implements MutableInterface
 {
-    /**
-     * @inheritdoc
-     */
-    public function getVal()
+    public function getVal(): int
     {
     }
-
+    
     /**
-     * Set value
-     *
-     * Sets 0 in case the value is bigger than max allowed value. {@inheritdoc}
+     * Sets 0 in case a non-integer value is passed
+     * 
+     * @param int|string|bool|float|null $value
      */
-    public function setVal($value)
+    public function setVal($value): void
     {
     }
 }


devops-devdocs:

Vinai:
Please hold this PR until after it was discussed in an architetcure meeting.

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -326,27 +328,29 @@ class Profiler
...
-  If there is no such operator, the `@return` tag must have `void` as the return value.
+* The declaration of all arguments (if any) using `@param` tag, unless the argument type is indicated in the method signature.
+  All `@param` annotations must include the appropriate argument type.
+  If any argument requires a `@param` annotation, all arguments must be listed (all or none).
+* The declaration of the return type using the `@return` tag should only be added if the method return type signature
+  does not supply all necessary information.
+  If the method does not `return` a value, the method type signature should be `void`, e.g. `function example(): void {}`.
 * Declaration of possibly thrown exception using `@throws` tag, if the actual body of function triggers throwing an exception.
-  All occurrences of `@throws` in a DocBlock must be after `@param` and `@return` (if any).
+  All occurrences of `@throws` in a DocBlock must be after any `@param` and `@return` annotations (if any).
```
joshuaadickerson: Remove the last "(if any)" or make it "must be after all".

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -539,39 +546,27 @@ Documentation is organized with the use of namespaces.
...
 
-Whenever possible the `@inheritdoc` tag MUST be used for children to avoid duplication of doc blocks.
-
-Even Though PHPDocumentor understands inheritance and uses the parent doc block by default (without `@inheritdoc` tag specified), including the tag helps ensure that the doc block is not missed at all.
-
-Rules for usage of the tag:
-
-* Use `@inheritdoc` (notice no braces around) to indicate that the entire doc block should be inherited from the parent method.
-* Use the inline `{@inheritdoc}` tag (with braces around) in long descriptions to reuse the parent's long description. The tagged method MUST have its own short description.
+The `@inheritdoc` tag MUST NOT be used.
```
joshuaadickerson: This should be `SHOULD NOT` as the tag is supposed to be used for inline documentation. You may want to document the parent as _part_ of the child. Say for instance your child method just throws a bunch of exceptions. You'd use `@inheritdoc` and then `@throws` after it.

buskamuza: Throwing new exception may be not the best example as it may break clients that depend on the parent. But adding long description that explains purpose of the method may require using `@inheritdoc`.

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: Done.

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: done.

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -257,7 +257,9 @@ require_once __DIR__ . '/../../functions.php';
 ### Classes and interfaces
 {:#classes-interfaces}
 
-Classes and interfaces must have a short description that is a human-understandable intention of the class.
+Classes and interfaces should have a short description that is a human-understandable intention of the class.
+If the short description adds no additional information beyond what the type name already supplies, the
+short description should be ommitted.
```
DrewML: s/ommitted/omitted

orlangur:
@buskamuza this PR is about removing ALL THE useless stuff.

> I'd suggest to allow `@inheritdoc`

Please provide some reasoning.

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -257,7 +257,9 @@ require_once __DIR__ . '/../../functions.php';
 ### Classes and interfaces
 {:#classes-interfaces}
 
-Classes and interfaces must have a short description that is a human-understandable intention of the class.
+Classes and interfaces should have a short description that is a human-understandable intention of the class.
+If the short description adds no additional information beyond what the type name already supplies, the
+short description should be omitted.
```
orlangur: MUST be omitted

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -326,27 +328,29 @@ class Profiler
 ### Functions and methods
 {:#functions-methods}
 
-Functions and methods must have:
+In general, typed method signature should be preferred over PHPDoc annotations when possible.
```
orlangur: SHOULD -> MUST
when -> whenever

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -326,27 +328,29 @@ class Profiler
...
 
-* Short description
-* Long description that explains the motivation behind the implementation.
+Functions and methods should have:
+
+* A short description in case it adds meaningful information beyond the method name.
+* If the purpose of the method is not obvious, a long description that explains the motivation behind the implementation.
```
orlangur: Some sentence to involve "describe why method is implemented and not how", please

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -326,27 +328,29 @@ class Profiler
...
    * For non-obvious implementations where the implementation logic is complicated or does not correspond to the Technical Vision or other known best practices, include an explanation in the doc block's description.
      An implementation is non-obvious if another developer has questions about it.
 
-* Declaration of all arguments (if any) using `@param` tag.
-  Appropriate argument type must be specified.
-* Declaration of return type using `@return` tag.
-  If there is no such operator, the `@return` tag must have `void` as the return value.
+* The declaration of all arguments (if any) using `@param` tag, unless the argument type is indicated in the method signature.
+  All `@param` annotations must include the appropriate argument type.
+  If any argument requires a `@param` annotation, all arguments must be listed (all or none).
```
orlangur: Disagree. Let's specify `@param` only for necessary arguments, when it is more accurate than type hinting.

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -326,27 +328,29 @@ class Profiler
...
+* The declaration of the return type using the `@return` tag should only be added if the method return type signature
```
orlangur: should -> must

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -326,27 +328,29 @@ class Profiler
...
+  does not supply all necessary information.
+  If the method does not `return` a value, the method type signature should be `void`, e.g. `function example(): void {}`.
```
orlangur: Something like "does not have return statement" would be more clear. should -> must

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -326,27 +328,29 @@ class Profiler
...
 * Declaration of possibly thrown exception using `@throws` tag, if the actual body of function triggers throwing an exception.
-  All occurrences of `@throws` in a DocBlock must be after `@param` and `@return` (if any).
+  All occurrences of `@throws` in a DocBlock must be after any `@param` and `@return` annotations.
 
 **Exceptions to these rules:**
 
-* Constructors may not have short and/or long description
-
 * Testing methods in Unit tests may not have doc blocks if the test's method name follows the convention (test<MethodName>)
```
orlangur: `test<MethodNameInWhichCircumstances>` convention must be mandatory, thus doc block must be not allowed in most cases

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -492,21 +496,24 @@ public function deleteDirectory($path)
 #### @return tag
 {:#return}
 
-If there is no explicit return statement in a method or function, a `@return void` should be used in the documentation.
+In general method return type signatures should be prefered over `@return` type annotations.
+If that is not possible due to ambiguous return types, the `@return` type annotation must be used.
+
+If there is no explicit return statement in a method or function, a `void` return type should be declarted in the method signature.
```
orlangur: Hm, this one sounds better, should -> must and please remove the part "If the method does not `return` a value, the method type signature should be `void`, e.g. `function example(): void {}`."

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -492,21 +496,24 @@ public function deleteDirectory($path)
...
 
-If the method returns itself, `return $this` should be used.
+If the method returns itself, `@return $this` should be used, unless the return type is specified in the method signature.
```
orlangur: Should -> must. Looks like such annotation will remain for the sake of BC only.

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -539,39 +546,29 @@ Documentation is organized with the use of namespaces.
...
 
-Whenever possible the `@inheritdoc` tag MUST be used for children to avoid duplication of doc blocks.
-
-Even Though PHPDocumentor understands inheritance and uses the parent doc block by default (without `@inheritdoc` tag specified), including the tag helps ensure that the doc block is not missed at all.
-
-Rules for usage of the tag:
-
-* Use `@inheritdoc` (notice no braces around) to indicate that the entire doc block should be inherited from the parent method.
-* Use the inline `{@inheritdoc}` tag (with braces around) in long descriptions to reuse the parent's long description. The tagged method MUST have its own short description.
+The `@inheritdoc` tag SHOULD NOT be used.
```
orlangur: MUST NOT

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -539,39 +546,29 @@ Documentation is organized with the use of namespaces.
...
+If a child class method requires a long description to explain its purpose, it may use `@inheritdoc` to indicate the new description is intended as an addition to the parent method description.
```
orlangur: We may allow `{@inheritdoc}` inline version only but I believe it would be better to forbid completely.

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -539,39 +546,29 @@ Documentation is organized with the use of namespaces.
...
+    public function getVal(): int;
 
     /**
-     * Set value
-     *
      * Sets 0 in case a non-integer value is passed
-     *
-     * @param int $value
+     * 
+     * @param mixed $value
```
orlangur: I believe `mixed` must be avoided in favor of specific types (even if there are 3-4 of them).

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -582,19 +579,16 @@ interface MutableInterface
...
     }
-
+    
     /**
-     * Set value
-     *
-     * Sets 0 in case the value is bigger than max allowed value. {@inheritdoc}
+     * Sets 0 in case a non-integer value is passed
+     * 
+     * @param mixed $value
```
orlangur: Same here for `mixed`

Vinai:
@orlangur She did - I added that to the PR:

> The `@inheritdoc` tag SHOULD NOT be used.  
> If a child class method requires a long description to explain its purpose, it may use `@inheritdoc` to indicate the new description is intended as an addition to the parent method description.  
> In general such method overrides are a code smell and should be used as an incentive to make the code more self-documenting if possible.

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: In my experience that leads to less readable code.

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: A return statement without a return value would not be covered by that expression.

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: But what if they are an infinite number of possible types, for example anything that can be cast to a string?

orlangur:
@Vinai also, I'll double check regarding v2.1 and v2.2 in terms of used sniffs but my gut feeling says that we should only revise v2.3 here (I would be able to apply automated PHPDoc change for all the three release lines though).

Vinai:
@orlangur I used the 2.1 version of the page because 2.2 and 2.3 just referenced it.
Should I try to copy the page contents of 2.1 into 2.3 so it no longer is a reference, and make the changes there, instead?

orlangur:
Ok, let me check automated checks in 2.1/2.2 first and then will ask somebody from DevDocs team.

guides/v2.1/coding-standards/docblock-standard-general.md
orlangur: True :) "should be avoided whenever possible" then. Still, I would get rid all of `mixed` from code samples in doc though.

guides/v2.1/coding-standards/docblock-standard-general.md
orlangur: Aha, let me reach Olga in Slack.

guides/v2.1/coding-standards/docblock-standard-general.md
orlangur: Some examples? I don't think `@param` annotations are for humans at all, they are to give maximum accurate information for IDE. Besides constructor, any method don't contain more than 4-5 parameters usually, so, I believe such block with partial types for arguments missing real type hinting will be readable anyway.

guides/v2.1/coding-standards/docblock-standard-general.md
orlangur: @Vinai,

> But what if they are an infinite number of possible types

Changed my mind. Some interface or pseudo-type must be used. If there are too many different classes possible, just use `object`.

So, I would forbid `mixed` completely but this and few other my items can be added to separate PR so that this one only concentrates on stripping redundant declarations.

guides/v2.1/coding-standards/docblock-standard-general.md
joshuaadickerson: I know default PhpStorm will complain about partial doc blocks. We could change this, but a lot of developers will not understand it and it will lead to inconsistencies. Not sure how IDEs or PhpDoc will handle the order of the arguments then.

guides/v2.1/coding-standards/docblock-standard-general.md
joshuaadickerson: Agreed. I don't think I've ever seen someone using inline inheritdoc without it looking weird in the documentation.

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -364,7 +368,7 @@ Functions and methods must have:
     For example: `@return Config|null`.
     The DockBlock needs to explain what situations return `null`.
 
-    Another example: `@param FileInterface | null`.
+    Another example: `@param FileInterface|null`.
```
joshuaadickerson: Any reason you didn't opt for the nullable operator? `@param ?FileInterface`

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: I don't think there is an absolute truth to be reached in this regard (as is often the case in regards to "understandability". Who can decide this matter?

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: What's the current state in regards to `@inheritdoc`? I'm all for getting rid of it, but if @buskamuza wants to allow it I'm okay with that, too, as long as it's discouraged to use it situations where it is only noise.

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: I got rid of `mixed` annotations in the examples.

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: I find it strange to use nullable on `@param` annotations. Is that a good idea? I honestly don't know. So far I've never encountered that in annotations. Happy to change it if you think it's good and PHPStorm understands it.

Vinai:
@orlangur I've applied most changes you requested in https://github.com/magento/devdocs/pull/4425#pullrequestreview-234103595

However:

> let's make no unnecessary alignment mandatory

and

> Let's make `/** @var Profiler */` the only form allowed.

I'm against those changes - they give me the feeling as if I'm beeing micro managed by someone oppressive.
As long as the style is consistent within a file I think it's better to allow either format. I've been discouraged from opening PRs or following through with PRs in the past by overzealous coding standard rules (e.g. in the phpunit/mink project).

orlangur:
@Vinai thanks, no problem, to me those make sense if they are supported by auto-formatting tooling. Will try to address them separately.

Asked Olga in Slack regarding inheritdoc and will double-check the overall document in next few days. Huge thanks on your efforts to aggregate all the pain points related to redundancy 👍

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -326,32 +328,34 @@ class Profiler
...
 
 **Exceptions to these rules:**
 
-* Constructors may not have short and/or long description
-
-* Testing methods in Unit tests may not have doc blocks if the test's method name follows the convention (test<MethodName>)
+* Testing methods in Unit tests may have doc blocks to describe the purpose of the test, for example referencing github issues.
 
-   * If the test does not follow the convention, it should have a doc block describing the covered methods
+* Test emthod annotations may include data providers and other testing annotations.
```
lenaorobei:
``` suggestion
* Test method annotations may include data providers and other testing annotations.
```

guides/v2.1/coding-standards/docblock-standard-general.md
lenaorobei: Can we also add a rule that `@param` annotations should be in the same order as method arguments?

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: thanks, fixed!

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: @lenaorobei done

Vinai:
@buskamuza Your requested change to allow `@inheritdoc` is in. Are there other changes you would like to request?

guides/v2.1/coding-standards/docblock-standard-general.md
orlangur: @Vinai,

> In my experience incomplete lists of params in the annotations lead to less readable code.

Do you have some examples? There are not a lot of situations when we will need annotation for parameter at all (only more accurate type like `array` -> `Product[]` and such). Methods beside constructor usually won't contain more than 3-4 arguments.

Vinai: I don't have any examples at hand, since I changed those to contain all parameters instead.
But here is a constructed example:

Single `@param` annotation:
```php
/**
 * @param SectionSourceInterface[] $sectionSourceMap
 */
public function __construct(
    ObjectManagerInterface $objectManager,
    SectionIdentifier $identifier,
    array $sectionSourceMap = []
) {
    $this->objectManager = $objectManager;
    $this->identifier = $identifier;
    $this->sectionSourceMap = $sectionSourceMap;
}
```

A complete list of parameters annotations:
```php
/**
 * @param ObjectManagerInterface $objectManager
 * @param SectionIdentifier $identifier
 * @param SectionSourceInterface[] $sectionSourceMap
 */
public function __construct(
    ObjectManagerInterface $objectManager,
    SectionIdentifier $identifier,
    array $sectionSourceMap = []
) {
    $this->objectManager = $objectManager;
    $this->identifier = $identifier;
    $this->sectionSourceMap = $sectionSourceMap;
}
```

Vinai:
@orlangur I'm happy to do whatever in regards to `@inheritdoc` - please tell me what to do and I'll adjust the PR accordingly.

guides/v2.1/coding-standards/docblock-standard-general.md
buskamuza: I agree with keeping just "The `@inheritdoc` tag SHOULD NOT be used." if it's ok for everyone.
I don't see why it should be "MUST NOT", as it doesn't break anything. But I'm ok with discouraging.

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -425,7 +430,7 @@ In general, use the `@throws` tag when the code uses *throw*:
  *
  * @param  string $elementId
  * @param  string $attribute
- * @param  mixed  $value
+ * @param  int|string|float|bool|object|null  $value
```
paliarush: Why is this change needed? It is not exactly equivalent transformation since `mixed` is broader.

Vinai: This was a request by @orlangur - I personally think `mixed` is fine, but don't feel strongly either way.

orlangur: @paliarush do you mean it can be `resource`? Generally `mixed` does not smell well.

orlangur:
@Vinai @buskamuza in which case we need `@inheritdoc`? It does not bring any value and complicates review process.

@Vinai another thing we discussed earlier - could you provide example when partial parameters documentation may be confusing? We will need parameter type in PHPDoc only in rare cases, like a typified array `string[]` or smth.

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -326,32 +328,35 @@ class Profiler
...
      An implementation is non-obvious if another developer has questions about it.
 
-* Declaration of all arguments (if any) using `@param` tag.
-  Appropriate argument type must be specified.
-* Declaration of return type using `@return` tag.
-  If there is no such operator, the `@return` tag must have `void` as the return value.
+* The declaration of all arguments (if any) using `@param` tag, unless the argument type is indicated in the method signature.
+  All `@param` annotations must include the appropriate argument type.
+  If any argument requires a `@param` annotation, all arguments must be listed (all or none).
+* The `@param` annotations must be in the same order as the method arguments.
```
orlangur: I would combine it with "all arguments must be listed" item

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -326,32 +328,35 @@ class Profiler
...
+* The declaration of the return type using the `@return` tag must only be added if the method return type signature
```
orlangur: It is unclear whether `@return void` is still mandatory or not?

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: Besides `resoruce`, the `@mixed` annotation includes any future types, too.

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -326,32 +328,35 @@ class Profiler
...
+  does not supply all necessary information (see below for more information on return types).
 * Declaration of possibly thrown exception using `@throws` tag, if the actual body of function triggers throwing an exception.
-  All occurrences of `@throws` in a DocBlock must be after `@param` and `@return` (if any).
+  All occurrences of `@throws` in a DocBlock must be after any `@param` and `@return` annotations.
```
orlangur: "any" may be omitted here at all.

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -326,32 +328,35 @@ class Profiler
...
 
 **Exceptions to these rules:**
 
-* Constructors may not have short and/or long description
-
-* Testing methods in Unit tests may not have doc blocks if the test's method name follows the convention (test<MethodName>)
+* Testing methods in Unit tests may have doc blocks to describe the purpose of the test, for example referencing github issues.
```
orlangur: Honestly, I like previous version more. My understanding is that we should insist  descriptive enough `test<MethodName>` method name without any PHPDoc is preferred.

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -326,32 +328,35 @@ class Profiler
...
 
-   * If the test does not follow the convention, it should have a doc block describing the covered methods
+* Test method annotations may include data providers and other testing annotations.
```
orlangur: This can be removed I believe.

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -326,32 +328,35 @@ class Profiler
...
 
-   * Non-testing methods should have a doc block with description. It includes data providers and any helper methods
+* Non-testing methods may have a doc block with description according to the rules above.
```
orlangur: This is not needed aswell: anything which is not forbidden is allowed.

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -492,21 +497,24 @@ public function deleteDirectory($path)
 #### @return tag
 {:#return}
 
-If there is no explicit return statement in a method or function, a `@return void` should be used in the documentation.
+In general method return type signatures should be prefered over `@return` type annotations.
```
orlangur: I would say

```
Return type signature MUST be used whenever possible.
`@return` annotation MUST be omitted when it brings no new information compared to type signature.
```

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -492,21 +497,24 @@ public function deleteDirectory($path)
...
+If that is not possible due to ambiguous return types, the `@return` type annotation must be used.
+
+If there is no explicit return statement in a method or function or a return statement without a value, a `void` return type must be declared in the method signature.
 
-If the method returns itself, `return $this` should be used.
+If the method returns itself, `@return $this` must be used, unless the return type is specified in the method signature.
```
orlangur: So, what to use for `function FooBar(): self` ? I think no `@return` needed in this case.

When there is not type hint, I'm a bit concerned whether we need explicit `@return $this` or PhpStorm may resolve correct type anyway.

guides/v2.1/coding-standards/docblock-standard-general.md
```diff
@@ -492,21 +497,24 @@ public function deleteDirectory($path)
...
 
 ### Constants
 {:#constants}
 
-Constants must have short description.
+Constants must have short description if they add information beyond what the constant name supplies.
```
orlangur: It's better like in another section: `MUST be omitted when it brings no additional information`

Vinai:
@orlangur Regarding partial vs. complete parameter annotations: I believe it boils down to familiarity which scenaerio is more or less confusing. We will not reach an agreement with examples here.

The use cases that require type annotations beyond what PHP is capable of typing are:

* typed array value annotations (like you said), e.g. `string[]` 
* quasi union types that have no common super type, e.g. `Foo|Bar`
* multiple scalar types, e.g. `string|int`

Just saying, because in my experience they aren't such a rare ocurance...

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: done

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: Does the sentence exclude `void`in some way? To me it seems clear.
Should `void` be made explicit?
e.g: If a method return type signature is `void` then the `@return void` annotation must be omitted.

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: done

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: A test name with a class name often is not sufficient. Since tests are documentation, I think it's good to be more permissive than the previous version of the sentence.

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: Because it is implicit?

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: removed

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: Restricting types make code harder to evolve over time. I believe it is better to sometimes NOT specify a return type until a method signature is very stable. This more permissive statement allows for this.

orlangur:
> Just saying, because in my experience they aren't such a rare occurrence...

@Vinai sure, but it won't happen for too many parameters of the same method. And such accurate type specification is more for IDE type resolution than human reading.

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: Good point. I'll expand the section accordingly.

guides/v2.1/coding-standards/docblock-standard-general.md
Vinai: done

Vinai:
Moved all changes to `guides/v2.2/coding-standards/docblock-standard-general.md` instead of the `2.1` file.

dobooth:
running tests

orlangur:
@Vinai glad to finally see it in, you rock 🎉 Special thanks for squashing prior to merge 😉 

Let's start the biggest cleanup ever made for core codebase.

PR #4490, opened 2019-05-11
[2.3.3] Updated lock config and lock env short commands

shikhamis11:
## This PR is a:

- Content fix or rewrite

## Summary

When this pull request is merged, it will update the short commands for lock env and lock config
as updated in Pull Request on magento 2 core 
https://github.com/magento/magento2/pull/22720

## Additional information
reference page of configuration command - /config-guide/cli/config-cli-subcommands-config-mgmt-set.html
is only available from magento 2.2 version onwards  

List all affected URLs 

https://devdocs.magento.com/guides/v2.2/config-guide/cli/config-cli-subcommands-config-mgmt-set.html
https://devdocs.magento.com/guides/v2.3/config-guide/cli/config-cli-subcommands-config-mgmt-set.html

PR #4507, opened 2019-05-13
Revert "[2.3.3] Updated lock config and lock env short commands"

dobooth:
Reverts magento/devdocs#4490


dobooth:
Reverting this commit because it references the 2.3.3 codebase. We will merge it into the correct branch soon.

PR #4519, opened 2019-05-15
Update multirepo script

jeff-matthews:

This pull request (PR) removes the fetch command from the multirepo script.

It seems unnecessary to fetch all branches from a repo if we only need the latest version of the main branch. The `init` stage on our pipelines is hanging periodically for some builds. My expectation after merging this PR is that we will reduce times for the `init` stage.

We don't have the visibility into the network infrastructure to troubleshoot the long stage times, so this seems like the most effective way to eliminate one possible cause for the long `init` stage times.

### DIFF
--- a/scripts/docs-from-code.sh
+++ b/scripts/docs-from-code.sh
@@ -13,7 +13,7 @@ echo 'Initiating git in the directory'
 git init
 
 echo "Adding a remote repository: $repo"
-git remote add origin --fetch "$repo"
+git remote add origin "$repo"
 
 if $sparse; then
   echo 'Enabling sparse checkout'


dshevtsov:
This will create empty local repositories and as a result - empty folders.
You have to fetch to download git history from a remote repository.

dshevtsov:
To test changes to the script locally, remove all directories that contain subrepositories and run `rake init`.

jeff-matthews:
Bleh... you're right.

jeff-matthews:
Since we're removing 2.0 content next week, hopefully this issue will become a moot.

dshevtsov:
To avoid fetching the repo for 2.0, we can use `git worktree add -f guides/v2.0 2.0` instead.
I plan to implement this to the script later when I have time.

PR #4539, opened 2019-05-16
Prefer exception expectations in body

Vinai:

This pull request (PR) gives guidance on how to declare expected exceptions in tests in a forward compatible manner.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/graphql/functional-testing.html

### DIFF
--- a/guides/v2.3/graphql/functional-testing.md
+++ b/guides/v2.3/graphql/functional-testing.md
@@ -202,7 +202,8 @@ Your functional tests should include events that cause exceptions. Since your te
 - The body of the test
 - The test function annotation
 
-It's better to declare expected exceptions in the test method body, as declaring expected exceptions with annonations has been deprecated in PHPUnit 8. Existing tests that use annotations will have to be updated when Magento upgrades PHPUnit.
+{: .bs-callout .bs-callout-tip }
+We recommend that you declare expected exceptions in the test method body, as declaring expected exceptions with annonations has been deprecated in PHPUnit 8. Existing tests that use annotations will have to be updated when Magento requires that version of PHPUnit or higher.
 
 ### Exception messages in the body of a test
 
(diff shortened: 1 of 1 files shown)


devops-devdocs:

keharper:
@atwixfirster based on @Vinai's comment, should we remove the annotations section?

atwixfirster:
> @atwixfirster based on @Vinai's comment, should we remove the annotations section?

The purpose of my update was - display all of possible ways how developers can set exceptions for tests. 

@keharper , if you want I can add a `bs-callout` message with best practices advice - set exception in the test body.

What do you think?

Thank you!

keharper:
@atwixfirster Yes, this text should be highlighted. Use `tip` instead of `info` 

I would also prefer to change "when Magento upgrades PHPUnit" to "when Magento requires that version of PHPUnit"

atwixfirster:
> @atwixfirster Yes, this text should be highlighted. Use `tip` instead of `info`
> 
> I would also prefer to change "when Magento upgrades PHPUnit" to "when Magento requires that version of PHPUnit"

@keharper , I was unable to make a commit into Vinai's PR. So, I've created this one - https://github.com/Vinai/devdocs/pull/1

Could you please merge?

Thank you!

keharper:
@atwixfirster I can't merge over there either. I'll update the file in this PR.

Vinai:
I'm a little confused - is there anything to do to the PR or are you taking care of all things?

> `bs-callout`
What does "bs" stand for? :D

@keharper 
> I would also prefer to change "when Magento upgrades PHPUnit" to "when Magento requires that version of PHPUnit"

To be correct it would need to be "that version of PHPUnit or newer" I think...
What is wrong with "upgrades PHPUnit" ? Just curious, happy to learn.

keharper:
@Vinai 

The `bs-callout...` is the Kramdown tagging that renders text as a note/warning/tip.

As to "upgrades PHPUnit", I know it's highly improbable that we start requiring PHPUnit 7 but it's technically possible.

keharper:
Thanks @Vinai

PR #4591, opened 2019-05-22
Add warning about the list widget is deprecated since version 2.2.0

serhiyzhovnir:

The list widget is deprecated since version 2.2.0 and this pull request adds warning about this.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/javascript-dev-guide/widgets/widget_list.html
- https://devdocs.magento.com/guides/v2.2/javascript-dev-guide/widgets/widget_list.html

- https://github.com/magento/magento2/blob/2.2/lib/web/mage/list.js#L7
- https://github.com/magento/magento2/blob/2.3/lib/web/mage/list.js#L7


devops-devdocs:

rogyar:
Hi @serhiyzhovnir. Thank you for the provided clarification. It would be awesome if we could shed some light on what should be used instead of the list widget if it's deprecated. I've lack of ideas in this particular case. Maybe you have thoughts on this matter? 
If it's the case where there are no obvious alternatives, we will leave it as it is. 

Thank you

serhiyzhovnir:
Hi @rogyar 
We have tried to find a place where Magento used the list widget, but we did not find it.
After this, we have checked the functionality of this component and the test for this component https://github.com/magento/magento2/blob/2.1/dev/tests/js/JsTestDriver/testsuite/mage/list/jquery-list-test.js.
The list widget is used to add and remove list items.
Also, it allows configuring item template, list limit and so on.
This functionality is really similar to the DynamicRows component (https://devdocs.magento.com/guides/v2.2/ui_comp_guide/components/ui-dynamicrows.html)
@rogyar should we suggest the DynamicRows component as an alternative?
Thank you!

rogyar:
For the admin area, I believe the DynamicRows is the best and commonly used alternative. So, yes, let's recommend it instead of the list widget, thank you.

serhiyzhovnir:
@rogyar I have added the suggestion about using the DynamicRows component as an alternative for the admin area.
Thank you!

dobooth:
running tests

dobooth:
running tests

dobooth:
running tests

dobooth:
5th times the charm.

PR #4594, opened 2019-05-22
Add spell check

dshevtsov:

This pull request adds a rake task `rake spellcheck` that runs the [forspell](https://github.com/kkuprikov/forspell) spell checker (which uses [Hunspell](https://en.wikipedia.org/wiki/Hunspell)). If the task is run without arguments it will check all the recently changed but not committed `.md` and `.rb` files.
If the `path` argument is added, the corresponding path will be checked recursively.

The forspell tool cannot be installed as a part of Gemfile because of conflicting dependencies with Jekyll. The tool uses kramdown 2 and Jekyll uses kramdown 1. Hope this will be resolved in the next Jekyll 4. Thus, the rake task uses `forspell` as a separately installed tool.

Install the forspell gem to enable spell checking:

```
gem install forspell
```

Custom dictionary is specified in the `forspell.dict` file. For more details about dictionaries, refer to [forspell dictionaries](https://github.com/kkuprikov/forspell#dictionaries) 

## Testing

1. Edit and save one or more Markdown files in the repository.
2. Run `rake spellcheck`
3. Add words you want to exclude from detection to `forspell.dict`
4. Run `rake spellcheck` and ensure that the excluded words are not in the list now
5. Run `rake spellcheck path=community` to check topics in the `community` folder
6. Exclude more words and ensure that they are not detected in the next run of the task

### DIFF
--- a/forspell.dict
+++ b/forspell.dict
@@ -1,0 +1,8 @@
+Changelog
+Chocolatey
+codebase
+Codeception
+decrypted: encrypted
+Homebrew
+Magento
+subdirectory: directory
--- a/Rakefile
+++ b/Rakefile
@@ -65,10 +65,21 @@ desc 'Optimize images in modified files, or by path (rake image_optim path=path/
 task :image_optim do
   path = ENV['path']
   unless path
-    staged_files = `git ls-files --modified --others --exclude-standard`.split("\n")
-    # staged_md_files = staged_files.select { |file| File.extname(file) == '.md' }
-    abort 'Didn\'t find any modified files.'.blue if staged_files.empty?
-    path = staged_files.join(' ')
+    modified_files = `git ls-files --modified --others --exclude-standard`.split("\n")
+    abort 'Didn\'t find any modified files.'.blue if modified_file.empty?
+    path = modified_file.join(' ')
   end
   system "bin/image_optim --no-pngout --no-svgo --recursive #{path}"
+end
+
+desc 'Pull docs from external repositories'
+task :spellcheck do
+  path = ENV['path']
+  unless path
+    modified_files = `git ls-files --modified --others --exclude-standard`.split("\n")
+    modified_md_files = modified_files.select { |file| File.extname(file).match?(/.(md|rb)/) }
+    abort 'Didn\'t find any modified Markdown or Ruby files.'.blue if modified_md_files.empty?
+    path = modified_md_files.join(' ')
+  end
+  system "forspell #{path}"
 end


jcalcaben:
If the `forspell` gem is required as a separate install. We should probably note that requirement in the README file or add this step in the init script.

shrielenee:
I reiterate my comment on the `rake` image PR---is this documented somewhere that I can refer to in the future?

forspell.dict
```diff
@@ -0,0 +1,8 @@
+Changelog
+Chocolatey
+codebase
+Codeception
+decrypted: encrypted
```
meker12:
```suggestion
decrypted: encrypted
Fastly
```

forspell.dict
```diff
@@ -0,0 +1,8 @@
+Changelog
```
meker12:
```suggestion
allowlist
blocklist
Changelog
```

forspell.dict
```diff
@@ -0,0 +1,8 @@
...
+Chocolatey
+codebase
+Codeception
```
meker12: cron

dobooth:
Been playing around with this. A couple notes:
The plugin can generate a custom dictionary of terms it would flag as errors. We can use this as a starting point for our own custom dict. The file is attached here.
[forspell.txt](https://github.com/magento/devdocs/files/3258293/forspell.txt)
Note that a large percentage of the terms are contributors names or companies. Most of these terms are coming from */release_notes/*. 
We will likely have to include these names in the dict file so they don't get flagged. We should debate whether including names/private repos should continue in the future.

Also, there seems to be a bug when there is code between words.
For instance: 
./guides/v2.1/mrg/ee/Staging.md:38: campaigna
./guides/v2.1/mrg/ee/Staging.md:41: staginga

If you check these files, you will see that `campaign` is on line 38, then some HTML. It's grabbing the `a` from the next line. Same with `staging` on line 41.

Another example: 
./guides/v2.3/config-guide/cache.md:28: backendsa 
`backendsa` comes from `backends&mdash;a`. The entity is being ignored.

dshevtsov:
The task has to be moved to the `check` namespace, for example `rake check:spell`, or `rake check:en`.

dobooth:
@dshevtsov Any update on this PR?

dshevtsov:
@dobooth cannot get to this for a while, busy with other projects.

dobooth:
Closing for now until Dima can get back to it.

PR #4645, opened 2019-05-30
Fix for listens property

OleksiiBulba:
Fix for listens property where key can use string templates but not value.

This pull request (PR) fix description for listens property.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/ui_comp_guide/concepts/ui_comp_linking_concept.html (sym link to the next page in the list)
- https://devdocs.magento.com/guides/v2.2/ui_comp_guide/concepts/ui_comp_linking_concept.html
- not https://devdocs.magento.com/guides/v2.1/ui_comp_guide/concepts/ui_comp_linking_concept.html (it has the right description)


devops-devdocs:

PR #4691, opened 2019-06-05
Add support of multirepo search

belbiy:

This pull request makes possible searching for multirepo content on the site using Algolia search.

@dshevtsov  wrote an awesome `_plugins/page-versionless.rb` that adds the `versionless: true` parameter to the pages that are not in `guides/v` folder. 

That parameter then used for faceting the search queries. 

When you search, it will pull results that are versioned with the current `guide_version` that you have on page and by `versionless: true` as well.

## Affected DevDocs pages

All pages.


belbiy:
running tests

PR #4752, opened 2019-06-13
Added 'rake check' to Rakefile

dobooth:

Adds a new command `rake check` that is used for checking devdocs before committing.
Currently runs image optimizer but we plan to add more checks soon.

## Additional information

The existing `image_optim` task has been moved to the `check:image_optim`.


dshevtsov:
running tests

PR #4760, opened 2019-06-14
Add GraphQL vault documentation

keharper:

This pull request (PR) documents the query and mutation added in https://github.com/magento/graphql-ce/pull/307

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/graphql/reference/vault.html

whatsnew
GraphQL supports the [Vault endpoint](https://devdocs.magento.com/guides/v2.3/graphql/reference/vault.html), which allows customers to store payment information with a service such as Braintree Vault.


_includes/graphql/customer-payment-tokens.md
```diff
@@ -0,0 +1,16 @@
+The `CustomerPaymentTokens` output object contains an array of `items`.
+
+Attribute | Data Type | Description
+--- | --- | ---
+`items` | [PaymentToken] | Contains an array of customer payment tokens
```
erikmarr: What's the difference between a "customer payment token" and "customer payment method"? Payment method makes more sense to me. Can we use that throughout?

guides/v2.3/graphql/reference/vault.md
```diff
@@ -0,0 +1,136 @@
...
+        }
+      ]
+    }
+  }
+}
+```
+
+## Mutation
+
+The `deletePaymentToken` mutation deletes a payment token from the system.
```
erikmarr: "deletes a payment method..."?

_includes/graphql/customer-payment-tokens.md
```diff
@@ -0,0 +1,16 @@
...
+
+#### PaymentToken attributes
+
+The `PaymentToken` object defines characteristics of a token stored in the payment vault.
+
+Attribute | Data Type | Description
+--- | --- | ---
+`details` | String | Stored account details
+`payment_method_code` | String | The payment method code associated with the token
+`public_hash` | String | The public hash of the token
```
erikmarr: Should we mention this is automatically generated like an auth token or customer cart ID? Maybe that's only necessary if we add a mutation to add payment token.

_includes/graphql/customer-payment-tokens.md
keharper: I can't change the terminology. Tokens are managed here, not payment methods.

_includes/graphql/customer-payment-tokens.md
keharper: Mentioned that the vault provider generates the token

_includes/graphql/customer-payment-tokens.md
joni-jones: @erikmarr , payment token and method are different. The payment token - tokenized on Payment Gateway credit card, payment method - an integration to provide a possibility to make a payment transaction, there are no terms like "customer payment method" because the payment method doesn't belong to any customer.

guides/v2.3/graphql/reference/vault.md
```diff
@@ -0,0 +1,136 @@
+---
+group: graphql
+title: Vault endpoint
+---
+
+When the vault feature is enabled in the payment method configuration for Braintree, customers have the option during checkout to save their credit card information. During subsequent checkouts, the customer is presented with a list of saved payment options. If Instant Purchase is enabled, customers can even by-pass the two-step checkout process and place the order from the product page.
```
joni-jones: Not only Braintree supports Vault. Also, there is might be 3rd party extensions which support Vault features. I would prefer to say: "When vault feature is supported by a payment integration and enabled, ...".

guides/v2.3/graphql/reference/vault.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Query
+
+The `customerPaymentTokens` query returns an array of stored payment methods.
```
joni-jones: "payment token" is not equal to "payment method"

guides/v2.3/graphql/reference/vault.md
keharper: Updated description.

PR #4765, opened 2019-06-17

ryantfowler:
To clarify the Technical Guidelines for using `declare(strict_types=1);` and strict return type declarations in PHP Interfaces. This comes as a result of a conversation with @maghamed .

## Affected DevDocs pages
- https://devdocs.magento.com/guides/v2.2/coding-standards/technical-guidelines.html
- https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html


devops-devdocs:

guides/v2.2/coding-standards/technical-guidelines.md
```diff
@@ -43,7 +43,7 @@ Use [RFC2119] to interpret keywords like:
 
 1.3. Type hints for scalar arguments SHOULD be used.
 
-1.3.1. All new PHP files MUST have strict type mode enabled by starting with `declare(strict_types=1);`. All updated PHP files SHOULD have strict type mode enabled. PHP interfaces SHOULD NOT have this declaration.
+1.3.1. All new PHP files MUST have strict type mode enabled by starting with `declare(strict_types=1);`. All updated PHP files SHOULD have strict type mode enabled. PHP interfaces MAY have this declaration.
```
buskamuza: IMO, this doesn't change much. "SHOULD NOT" still allows to have it. "MAY" still allows to omit it.
We had a discussion some time ago and decided that it's better to not have strict types in interfaces because it doesn't make sense there.
Both sides have some arguments, and none are 100% objective, but more a preference. We can vote here and decide based on majority to make sure most people are satisfied.

ryantfowler: I think the change of language makes it a bit clearer, because "SHOULD NOT" seems more like what it says, should not ... but then there's discrepancies between core and MSI interfaces, so for someone looking to the core for a reference and cross-referencing that with the Technical Guidelines, it's quite confusing. Beyond that.... even though from a technical perspective having strict types declared on Interfaces, it does convey intent.

buskamuza:
Screen Shot 2019-07-10 at 1 50 13 PM

dobooth:
running tests

PR #4786, opened 2019-06-19
Separate indexing configuration into `_config.index.yml`

belbiy:

- This pull request (PR) prevents the code blocks line numbers from being indexed by Jekyll Algolia.
- Separates indexing config from public search config (should be left at `_config.yml`)

We have to change the Jenkins indexing job in order for this to work:
```
ALGOLIA_API_KEY='$var' bundle exec jekyll algolia --config _config.yml,_config.index.yml
```

## Affected DevDocs pages

none

### DIFF
--- a/_config.index.yml
+++ b/_config.index.yml
@@ -1,0 +1,24 @@
+# Jekyll Algolia settings for indexing https://community.algolia.com/jekyll-algolia/options.html
+algolia:
+  files_to_exclude:
+    - guides/m1x/**/*.html
+    - guides/m1x/**/*.md
+    - swagger
+    - redoc
+  # Index HTML elements, especially code samples.
+  nodes_to_index: 'p,li,td,code'
+  # Override 10KB default and use our allowed max of 20KB
+  max_record_size: 20000
+  settings:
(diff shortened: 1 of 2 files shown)


jeff-matthews:
@belbiy, are you separating the config because we expect it to get larger/more complex and this will make it easier to maintain?

Also, to update the Jenkins pipeline, do you mind creating a PR here?

https://git.corp.adobe.com/magento-docs/infrastructure/blob/master/jenkins/pipeline-scripts/devdocs-algolia-reindex.groovy

belbiy:
@jeff-matthews , the main problem i'm trying to solve is indexing of the code line numbers:

Screen Shot 2019-06-20 at 11 26 11 AM

From my point of view, separate config is not much cleaner and maintainable, but it's the only way to achieve the difference with rendering for production and rendering for indexing.

And i will make PR to the jenkins pipeline, thank you for the link!

belbiy:
Pipeline script updated. https://git.corp.adobe.com/magento-docs/infrastructure/pull/15

belbiy:
running tests

PR #4801, opened 2019-06-21
Updated Backward compatibility guidelines

buskamuza:
- clarified public vs private code
- virtual types to be private by default
- added static view files

Improve and add more rules to backward compatibility guidelines.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/architecture/extensibility.html
- https://devdocs.magento.com/guides/v2.3/extension-dev-guide/versioning/codebase-changes.html
- https://devdocs.magento.com/guides/v2.3/extension-dev-guide/versioning/dependencies.html

❗️ please don't merge before architectural review.


devops-devdocs:

guides/v2.1/architecture/extensibility.md
```diff
@@ -32,6 +32,22 @@ Adherence to standard best practices for [PHP](https://glossary.magento.com/php)
 
 Magento has well-defined upgrade and versioning strategies that can help you avoid any problems with software component dependencies. Add modules after confirming that the [module](https://glossary.magento.com/module) version is compatible with the Magento Framework version.
 
+## Extension Points
+
+While each module can introduce its own extensibility points guided by the above principles, there is a number of extension points provided by Magento Framework or Magento core modules that are not specific to a certain feature.
+Below is a list of such generic extension points:
+* PHP interfaces marked with `@api`
+* Plugins
```
joni-jones: Why plugins are extensions points?

guides/v2.1/extension-dev-guide/versioning/codebase-changes.md
```diff
@@ -108,14 +109,19 @@ Use this table to understand what changes Magento can make and which version num
 | | Changed event property | MAJOR|
 | | Removed event property | MAJOR|
 | | Changed event ordering | MAJOR|
-| **Virtual Type**| Virtual type removed | MAJOR|
+| **Virtual Type (marked with `@api`)**| Virtual type removed | MAJOR|
 | | Virtual type added | MINOR|
-| **URL Paths** | Path removed | MAJOR|
-| | Removed/renamed a request parameter| MAJOR|
```
buskamuza: @paliarush : why "removed" is MAJOR - it will not fail requests.
@kokoc : removal removes some functionality, so it's a MAJOR change.
@buskamuza : functional changes should be considered separately from technical changes.

guides/v2.1/architecture/extensibility.md
```diff
@@ -32,6 +32,22 @@ Adherence to standard best practices for [PHP](https://glossary.magento.com/php)
...
+* XML configuration. Some common configuration files are:
+  * `di.xml`
+  * `config.xml`
+  * `system.xml`
+  * `mview.xml`
+* Events and observers
```
joni-jones: Why observers are extensions points? I agree about events but observers more like private consumers.

guides/v2.1/extension-dev-guide/versioning/codebase-changes.md
```diff
@@ -108,14 +109,19 @@ Use this table to understand what changes Magento can make and which version num
...
-| | New required request parameter | MAJOR|
-| | New optional request parameter | MINOR|
+| **URL Paths (Store-Front and Admin)** | Path removed | MAJOR|
+| | Removed/renamed a request parameter (GET)| MAJOR|
+| | New required request parameter (GET) | MAJOR|
+| | New optional request parameter (GET) | MINOR|
+| | Path removed (POST) | PATCH|
```
buskamuza: @kokoc, @paliarush : POST is also public API.
Change to the same as GET.

guides/v2.1/extension-dev-guide/versioning/codebase-changes.md
```diff
@@ -150,6 +156,5 @@ Use this table to understand what changes Magento can make and which version num
 | | Index added/changed| PATCH|
 | | Foreign key added| MAJOR|
 | | Temporary tables added/removed/changed | PATCH|
-  
-  
-[private]: http://php.net/manual/en/language.oop5.visibility.php
+| **Static view files** | JS/CSS/LESS file removed/moved | MAJOR|
```
buskamuza: @joni-jones : Marked with `@api`

guides/v2.1/extension-dev-guide/versioning/codebase-changes.md
```diff
@@ -108,14 +109,19 @@ Use this table to understand what changes Magento can make and which version num
...
+| **Virtual Type (marked with `@api`)**| Virtual type removed | MAJOR|
 | | Virtual type added | MINOR|
-| **URL Paths** | Path removed | MAJOR|
-| | Removed/renamed a request parameter| MAJOR|
-| | New required request parameter | MAJOR|
-| | New optional request parameter | MINOR|
+| **URL Paths (Store-Front and Admin)** | Path removed | MAJOR|
+| | Removed/renamed a request parameter (GET)| MAJOR|
+| | New required request parameter (GET) | MAJOR|
+| | New optional request parameter (GET) | MINOR|
```
viktym: @buskamuza  considering a new optional request parameter (GET/POST) as MINOR change technically blocking delivery a significant number of bug fixes in the patch release. I agree that the new parameter adds new or change somehow current behavior,  but the behavior can be changed without adding this parameter as well. I believe we should apply this limitation only for URL paths that intended for access by 3-rd party extensions or there is a high probability of such case

dobooth:
@buskamuza can you please resolve this merge conflict?

guides/v2.1/extension-dev-guide/versioning/codebase-changes.md
buskamuza: Will discuss with the community. By default, I propose to consider such changes MINOR/MAJOR for now.

guides/v2.1/contributor-guide/contributing_dod.md
```diff
@@ -15,12 +15,20 @@ At a high level, the criteria defined in our "definition of done" revolve around
...
+### Functional Backward Compatibility
 
 Existing product features and functionality must be retained during any changes to the code.
-If backward-incompatible functional changes are made, the documentation should explain the justification and provide the "business value".
+Any backward-incompatible functional changes must be approbved by a product owner.
+The documentation should explain the justification and provide the "business value".
+
+### Technical Backward Compatibility
+
+Technical backward compatibility must be retained between PATCH versions of Magento products. It should also be retained between MINOR releases if possible.
```
pdohogne-magento: Why is MINOR downgraded to "should" instead of "must"?  SemVer rules (which we say we follow) state that all incompatible changes are MAJOR.  If this is referring to marketing version instead of module version it should be clarified (since we allow MAJOR-level module changes in MINOR-level marketing releases; e.g. module goes from 100.x.x to 101.0.0 between Magento 2.2 and 2.3).

guides/v2.1/extension-dev-guide/versioning/codebase-changes.md
```diff
@@ -9,40 +9,40 @@ redirect_from:
...
 
 Changes in public code always trigger MINOR or MAJOR version increases.
 
 In most cases, modifications to private code will trigger PATCH version increases.
 On rare occasions, if the Magento development team made significant modifications to private code, they will increase the MINOR or MAJOR version.
 
-Third-party modules should not use or change private code.
-If this is unavoidable, you must depend on the patch version of the core modules used.
+Third-party modules should not rely on private code.
+If this is unavoidable, you must depend on the PATCH version of the core modules used.
```
pdohogne-magento: We should discuss what this means if we start using four-segment/hotfix version numbers

guides/v2.1/extension-dev-guide/versioning/codebase-changes.md
pdohogne-magento: @viktym Do we have a way of marking which URL paths are intended for 3rd-party access?

guides/v2.1/extension-dev-guide/versioning/codebase-changes.md
```diff
@@ -108,14 +110,19 @@ Use this table to understand what changes Magento can make and which version num
...
 | | Removed event property | MAJOR|
 | | Changed event ordering | MAJOR|
-| **Virtual Type**| Virtual type removed | MAJOR|
+| **Virtual Type (marked with `@api`)**| Virtual type removed | MAJOR|
 | | Virtual type added | MINOR|
-| **URL Paths** | Path removed | MAJOR|
-| | Removed/renamed a request parameter| MAJOR|
-| | New required request parameter | MAJOR|
-| | New optional request parameter | MINOR|
+| **URL Paths (Store-Front and Admin)** | Path removed | MAJOR|
```
pdohogne-magento: Do we use PUT/UPDATE/DELETE endpoints anywhere?

guides/v2.1/extension-dev-guide/versioning/codebase-changes.md
```diff
@@ -150,6 +157,18 @@ Use this table to understand what changes Magento can make and which version num
 | | Index added/changed| PATCH|
 | | Foreign key added| MAJOR|
 | | Temporary tables added/removed/changed | PATCH|
-  
-  
-[private]: http://php.net/manual/en/language.oop5.visibility.php
+| **Static view files (marked with `@api`)** | JS/CSS/LESS file removed/moved | MAJOR|
+| | JS/CSS/LESS file added | PATCH|
+
+### PHP Classes Intended for Extension
```
pdohogne-magento: We should note what's different about versioning for these classes.  Also, having some annotation in the code for such classes would be very useful (both for extension development and semver tooling).

guides/v2.1/contributor-guide/contributing_dod.md
```diff
@@ -15,12 +15,20 @@ At a high level, the criteria defined in our "definition of done" revolve around
...
+## Backward Compatibility 
+
+### Functional Backward Compatibility
 
 Existing product features and functionality must be retained during any changes to the code.
-If backward-incompatible functional changes are made, the documentation should explain the justification and provide the "business value".
+Any backward-incompatible functional changes must be approbved by a product owner.
+The documentation should explain the justification and provide the "business value".
+
+### Technical Backward Compatibility
```
pdohogne-magento: We need to define the difference between functional and technical backwards compatibility.

guides/v2.1/extension-dev-guide/versioning/codebase-changes.md
```diff
@@ -62,6 +62,7 @@ public function save()
 
 This table lists code change scenarios and which version number it affects.
 Use this table to understand what changes Magento can make and which version number gets increased for that change.
+Any change not listed below is considered a PATCH level change.
```
pdohogne-magento: Has this list been checked against our semver tool's rules (including Magento-custom rules and semverchecker's defaults)?

guides/v2.1/contributor-guide/contributing_dod.md
```diff
@@ -15,12 +15,20 @@ At a high level, the criteria defined in our "definition of done" revolve around
...
 The following sections provide additional details about each of these criteria:
 
-## Functional Backward Compatibility
+## Backward Compatibility 
+
+### Functional Backward Compatibility
 
 Existing product features and functionality must be retained during any changes to the code.
-If backward-incompatible functional changes are made, the documentation should explain the justification and provide the "business value".
+Any backward-incompatible functional changes must be approbved by a product owner.
```
scottsb: Typo in "approved."

guides/v2.1/extension-dev-guide/versioning/codebase-changes.md
buskamuza: This document should be a base for SemVer checker. A team is currently working on improving the tool and they are checking it with the document.
cc @irenelagno

guides/v2.1/extension-dev-guide/versioning/codebase-changes.md
buskamuza: If we will, all documents that mentioning versioning should be reviewed and updated. Agree.

guides/v2.1/extension-dev-guide/versioning/codebase-changes.md
```diff
@@ -9,40 +9,40 @@ redirect_from:
...
+All other code is considered private.
 
 Changes in public code always trigger MINOR or MAJOR version increases.
 
 In most cases, modifications to private code will trigger PATCH version increases.
 On rare occasions, if the Magento development team made significant modifications to private code, they will increase the MINOR or MAJOR version.
 
-Third-party modules should not use or change private code.
-If this is unavoidable, you must depend on the patch version of the core modules used.
+Third-party modules should not rely on private code.
```
scottsb: This wording change seems confusing to me. Your module will _rely_ on private code even if you're only using it indirectly via public @api extension points. I'm not sure what the intended clarification versus the old version is, but is there a clearer/more explicit way of saying it?

guides/v2.1/contributor-guide/contributing_dod.md
buskamuza: Yes, it's about marketing version. That's why "versions of Magento products" in the first sentence. But I'll rephrase to make all more clear.

guides/v2.1/extension-dev-guide/versioning/codebase-changes.md
buskamuza: I think in WebAPI only. It's not covered by these items.

guides/v2.1/architecture/extensibility.md
buskamuza: Removed those changes to not complicate this PR. Will deliver this separately.

guides/v2.1/extension-dev-guide/versioning/codebase-changes.md
buskamuza: We don't currently. So all of therm are public API.

viktym: Could we define @api URL paths first and only then apply limitations regarding optional request parameters? It will help to avoid blocking of delivery bugfixes in patch releases

viktym: > We don't currently. So all of them are public API.

In that case, the new URL path should be treated as a MINOR also since some extensions can start using it.
So we are not allowed to add in a patch a new URL-path or optional request parameter. It's a strong limitation.
I believe it should be applied only to @api URL-path.

buskamuza:
Related Twitter discussion - https://twitter.com/buskamuza/status/1144638525234655232

nathanjosiah:
Can we add a rule for adding translation strings being a MINOR change?

guides/v2.2/extension-dev-guide/versioning/dependencies.md
```diff
@@ -84,8 +85,9 @@ Use this table to set the appropriate version dependency on a module based on ho
 | | Extend from a class | MINOR|
 | | Override a method | MINOR|
 | | Subscribe to an event| MINOR|
-| **Virtual Type**| Configure a virtual type in the `di.xml` file as a class dependency | MAJOR|
-| **URL Paths** | Link to from custom pages | MAJOR|
+| **Virtual Type (marked with `@api`)**| Use an existing virtual type in the `di.xml` file as a class dependency | MAJOR|
+| **URL Paths** | Link to a URL from custom pages | MAJOR|
+| | Submit a POST request to the URL | MAJOR|
```
joni-jones: I still don't agree that changing POST params is a MAJOR change. In most cases, such requests are handled by controllers which are not public API.

buskamuza: Controller is private code in terms that nobody should call `Controller->execute()` directly. But an HTTP request is public API (or at least it should be considered as a different thing).

buskamuza: What level of change do you suggest for POST params? Then probably, for URLs as well (because somebody can't just rely on a POST URL, while not specifying the params).
MINOR or PATCH?

joni-jones: We have clear separation on URL level for WEB API and other requests. Changing POST params WEB API requests should be MAJOR (such entry-points exposed by service contracts), other POST requests - PATCH.

viktym: Defining any change in any POST request structure as MAJOR it's a very strong restriction.
How about new optional POST parameters? I still believe it should be a PATCH.
Extension developers should verify correct work of their extensions with the lowest patch version of the compatible minor release. In such way, they'll avoid problems with relying on some code that can appear in a patch release (approved or hidden BIC).

viktym: Another point regarding why all POST HTTP requests should be considered as public API?
I think the statement that somebody can create dependency on them is not enough.
In the earlier stages, I see that it is more of a necessity than a desired state. It's like with our private and public code. In the early stages, we changed all classes in a backward-compatible way, regardless of whether they are API or not. But now we follow BC only for code marked as API. The same should be for POST request. If we follow BC for POST request then we should document a list of maintained parameters with their descriptions like for WEB API. My opinion is that only WEB API requests should be under BC and possibly some shortlist of other requests. At least it can be our desired state.

pdohogne-magento: I have no input on whether or not we should consider POST requests public API, though if we do a subset it needs to be documented somewhere and annotated in code (both for usage/development and semver tooling purposes).

If we do have POST requests as part of the API, I would apply the same rules to them as to API functions in PHP: adding required params major, optional params minor, removing any params major according to standard semver rules.

guides/v2.2/extension-dev-guide/versioning/codebase-changes.md
```diff
@@ -146,6 +149,21 @@ Use this table to understand what changes Magento can make and which version num
...
 | | Temporary tables added/removed/changed | PATCH|
-  
-  
-[private]: http://php.net/manual/en/language.oop5.visibility.php
+| **Static view files (marked with `@api`)** | JS/CSS/LESS file removed/moved | MAJOR|
+| | JS/CSS/LESS file added | PATCH|
+
+### PHP Classes Intended for Extension
+
+Though inheritance is discouraged in Magento, the following classes are still acceptable for extension at the moment:
```
pdohogne-magento: We should annotate these classes in code as well

guides/v2.2/extension-dev-guide/versioning/codebase-changes.md
```diff
@@ -108,7 +108,7 @@ Any change not listed below is considered a PATCH level change.
 | | Changed event ordering | MAJOR|
 | **Virtual Type (marked with `@api`)**| Virtual type removed | MAJOR|
 | | Virtual type added | MINOR|
-| **URL Paths (Store-Front and Admin, any HTTP method)** | Path removed | MAJOR|
+| **URL Paths (Store-Front and Admin, GET)** | Path removed | MAJOR|
```
pdohogne-magento: Why is removing other types of requests not MAJOR as well?

guides/v2.2/extension-dev-guide/versioning/codebase-changes.md
```diff
@@ -77,8 +78,8 @@ Use this table to understand what changes Magento can make and which version num
 | | New optional method argument | MINOR|
 | | Removed a non-last argument| MAJOR|
 | | New required constructor object argument | MINOR|
-| | New optional constructor argument for [extendable @api][extendable-api]{:target="_blank"} | MINOR|
-| | New optional constructor argument | PATCH|
+| | New optional constructor argument in a class intended for extension | MINOR|
```
paliarush: Please give reference to the section listing these classes.

guides/v2.2/extension-dev-guide/versioning/codebase-changes.md
buskamuza: We didn't come to an agreement on this. I'll prepare a separate PR for POST requests to not block this one.

guides/v2.2/extension-dev-guide/versioning/codebase-changes.md
buskamuza: Added

dobooth:
running tests

PR #4884, opened 2019-07-03
Add missing word

erikhansen:

Change this:
> Magento auto-generates this file **it** does not detect

to this:
> Magento auto-generates this file **if it** does not detect

This pull request (PR) ...

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/cloud/reference/discover-deploy.html


devops-devdocs:

jeff-matthews:
@erikhansen, thanks for the PR, but we no longer accept PRs to 2.1 docs.

For us to accept this PR, you'll need to make the change to the 2.2 file.

erikhansen:
@jeff-matthews Good catch, thanks. I've updated the PR so that the 2.2 file is edited.

PR #4898, opened 2019-07-08
Removed unclear/strange sentence from Security, Performance, and Data Handling

### LINKED ISSUE #4804: Missing link to the glossary item / possibly unfinished section
# Bug report

## Description

Security, Performance, and Data Handling page contains a section covering using a correct object as a source to get correct data.

It has a sentence "For example, get $1 data from the Product instead of the Order object." which doesn't really explain how does one decide on which object to use to get data and in what cases. It also mentions **$1** but doesn't provide any explanation to what it is. It links to the missing page in the glossary.

### Steps to reproduce

1. Visit https://devdocs.magento.com/guides/v2.3/ext-best-practices/extension-coding/security-performance-data-bp.html#get-the-correct-data-from-the-correct-object
2. Review the section
3. Try tapping on **$1** to make sure it links to a missing page 

## Expected result

The section should have correct links, use terms that are easier to understand and have a deeper dive text covering the topic. 

## Additional information

dobooth:

This pull request (PR) removes an unclear/not helpful sentence from Devdocs.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/ext-best-practices/extension-coding/security-performance-data-bp.html

Closes #4804.


guides/v2.2/ext-best-practices/extension-coding/security-performance-data-bp.md
```diff
@@ -23,16 +23,13 @@ Make sure that your Magento application does not directly use any PHP supergloba
...
   $GLOBALS, $_SERVER, $_GET, $_POST, $_FILES, $_COOKIE, $_SESSION, $_REQUEST, $_ENV
   ```
+
 Instead use the [`Magento\Framework\HTTP\PhpEnvironment\Request`]({{ site.mage2bloburl }}/{{ page.guide_version }}/lib/internal/Magento/Framework/HTTP/PhpEnvironment/Request.php){:target="_blank"} wrapper class to safely access these values.
 
 ## Use the correct MySQL data types
 
 MySQL offers a range of numeric, string, and time data types. If you are storing a date, use a DATE or DATETIME field. Using an INTEGER or STRING can make SQL queries more complicated, if not impossible. It is often tempting to invent your own data formats; for example, storing serialized PHP objects in string. Database management may be easier, but MySQL will become a dumb data store and it may lead to problems later.
 
-## Get the correct data from the correct object
```
meker12: Issue requested the following: ```The section should have correct links, use terms that are easier to understand and have a deeper dive text covering the topic.```

The section does seem like it's providing some useful information -- that is, you want to ensure that you are pulling the correct data from the Magento data base. For this example,  I am guessing that there is an order collection and a product collection which probably share the same data fields, and if you meant to get product data and got order data you would run into issues.   It seems like this could be stated more clearly with pointers to tutorial or other information that explained about pulling data from different collections in Magento.

dobooth:
Discussed this with Olga. She suggested that we just remove it. It doesn't enhance anything and there isn't really anything to be added here.

dobooth:
running tests

PR #4933, opened 2019-07-10
Adding key "defaults" to examples

### LINKED ISSUE #4932: Devdocs should give an exact example of where we have to put import/export/listen properties
# Clarify the usage of import, export and listen

## Topic Link

https://devdocs.magento.com/guides/v2.3/ui_comp_guide/concepts/ui_comp_linking_concept.html

## Description

I think that it should have an exact example of where we need to put the export/import/listen properties

## Possible solutions

Add another example, like this one:
```javascript
{
  "defaults": {
    "imports": {
     "visible": "${ $.provider }:visibility"
    }
  }
}
```

LucasCalazans:

This pull request (PR) adds the `defaults` key to all examples used in js files

## Affected DevDocs pages

v2.3:
https://devdocs.magento.com/guides/v2.3/ui_comp_guide/concepts/ui_comp_linking_concept.html

v2.2
https://devdocs.magento.com/guides/v2.2/ui_comp_guide/concepts/ui_comp_linking_concept.html

Here we have some core codes using theses props inside a `default` key:

Magento_Downloadable: https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Downloadable/view/adminhtml/web/js/components/is-downloadable-handler.js#L11

Magento_Catalog: https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Catalog/view/base/web/js/product/list/columns/price-box.js#L17

Magento_Ui: https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Ui/view/base/web/js/form/form.js#L163

## Issue

Closes #4932


devops-devdocs:

PR #4979, opened 2019-07-17
Reverting markdown to fix the build

dobooth:
Somehow older experimental code snuck in and is breaking the build.
I am reverting back to the last known good markdown.

- https://devdocs.magento.com/guides/v2.2/extension-dev-guide/versioning/check-version.html


dobooth:
running tests

PR #5003, opened 2019-07-19
MAGECLOUD-3932 update bin docker commands

hguthrie:
This PR updates the `bin/docker` command list in the Docker quick reference topic.

Also, provided an example of `bin/docker` usage and improved the ece-tools build and deploy description.

Added PHP to one command. Some cloud environments cannot interpret the command without PHP, so this is to be safe.

whatsnew
Updated the `bin/docker` command list in the [Docker quick reference](https://devdocs.magento.com/guides/v2.2/cloud/docker/docker-quick-reference.html#bindocker) topic.


hguthrie:
running tests

PR #5076, opened 2019-07-30
Small Fix declarative schema setup_version

osrecio:
Omit setup_version if you not use declarative schema

This pull request (PR) fix a small typo. If you don't use a declarative schema you don't need add a `setup_version` in `module.xml`

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/extension-dev-guide/build/create_component.html#add-component-xml

- https://github.com/magento/magento2/blob/2.3-develop/lib/internal/Magento/Framework/Module/etc/module.xsd#L46  (use="optional")


devops-devdocs:

guides/v2.3/extension-dev-guide/build/create_component.md
```diff
@@ -27,7 +27,7 @@ The smallest working `module.xml` file would look something like this:
 </config>
 ```
 
-The `name` parameter defines the name of your component. It is required for all components. If you do not use [Declarative Schema]({{ page.baseurl }}/extension-dev-guide/declarative-schema/index.html)  to help manage the installation and upgrade processes for your component, then you must also add the  `setup_version` parameter to the `module` line. Set the `setup_version` value to your module's [database schema](https://glossary.magento.com/database-schema) version. Omit the `setup_version` parameter if you implement Declarative Schema.
+The `name` parameter defines the name of your component. It is required for all components. If you do not use [Declarative Schema]({{ page.baseurl }}/extension-dev-guide/declarative-schema/index.html)  to help manage the installation and upgrade processes for your component, then you must also add the  `setup_version` parameter to the `module` line. Set the `setup_version` value to your module's [database schema](https://glossary.magento.com/database-schema) version. Omit the `setup_version` parameter if you no implement Declarative Schema.
```
jeff-matthews: This contradicts what we're saying in the third sentence:

> If you do not use [Declarative Schema]({{ page.baseurl }}/extension-dev-guide/declarative-schema/index.html)  to help manage the installation and upgrade processes for your component, then you must also add the  `setup_version` parameter to the `module` line.

osrecio: It should be the opposite. I will change it . If you don't use Declarative Schema you don't need add the attribute `setup_version`. If you use it you need add the attribute `setup_version`

jeff-matthews:
Thanks @osrecio!

davidalger:
@jeff-matthews This PR seems to have made DevDocs inaccurate literally doing nothing more than inverting the technical aspect of the paragraph which indicated that `setup_version` is now optional when Declarative Schema is used.

slack-imgs

The `setup_version` property of `module.xml` has been around since the inception of Magento 2 and it's no longer needed when Declarative Schema is used since the schema updates are applied when they differ from the database (something which `setup:db:status` will check) rather than relying on the `setup_version` (aka schema version) to determine if upgrade routines must be run.

Unless I'm totally missing something here, this PR really needs to be reverted. There is perhaps a cleaner way to call this out (perhaps mentioning it ON the page about declarative schema), but the original text is I believe accurate here.

osrecio:
Hi @davidalger ,

I propose, instead of reversing it, that I could agree with you. Depending on how you understand the context.

What I would do is simply comment that the use of the version is no longer necessary due to the Declarative Schema. Since it's 2.3 documentation, there is always a mention of previous versions of Magento and I think that should be enough. In this way, you encourage the reader to use the Declarative Schema over the old way of Setup.

How do you see this proposal?

jeff-matthews:
At what point did using `setup_version` become optional for declarative schema? Was it a specific 2.3.x release?

davidalger:
@osrecio The reason I'm suggesting it be reversed is this PR basically makes it (inaccurately) state the following:

* If you use Declarative Schema… then you **must add** the  `setup_version`
* Omit the `setup_version` parameter if you **do not use** Declarative Schema

Quite the opposite is true. When declarative schema is used, the `setup_version` **should be** omitted. Prior to this PR the doc (correctly) stated the following:

* If you **do not use** Declarative Schema... then you **must add** the `setup_version` parameter
* Omit the `setup_version` parameter **if you implement** Declarative Schema

If there is a better way to word the note, I'm very open to it, but this PR didn't clarify anything it changed the meaning of the text entirely to be the opposite of what it was before.

@jeff-matthews The use of `setup_version` became optional in the XSD with the release of 2.3.0 which introduced Declarative Schema. See the following links:

* [XSD for 2.3.0](https://github.com/magento/magento2/blob/2.3.0/lib/internal/Magento/Framework/Module/etc/module.xsd#L46)
* [XSD for 2.2.11](https://github.com/magento/magento2/blob/2.2.11/lib/internal/Magento/Framework/Module/etc/module.xsd#L46)

osrecio:
@davidalger Can you revert the PR and assign it to me? 
I'll be happy to accept it.

jeff-matthews:
Sorry it's taken so long to get to this. I just created a PR.

davidalger:
Thanks @jeff-matthews!

jeff-matthews:
Thanks for reporting it @davidalger!

PR #5101, opened 2019-08-01
Change the PWA studio url in Federated Search

belbiy:

This pull request (PR) changes the link to PWA studio website in federated search

## Affected DevDocs pages

- ...

- ...


belbiy:
running tests

_includes/layout/header-scripts.html
```diff
@@ -27,7 +27,7 @@
     {
       label: "PWA",
       name: "pwa-devdocs",
-      baseUrl: "https://magento-research.github.io/pwa-studio"
+      baseUrl: "https://magento.github.io/pwa-studio"
```
jcalcaben:
```suggestion
      baseUrl: "https://pwastudio.io"
```
it looks like you were right, eric must have changed the behavior.
this can be changed to: `pwastudio.io` and it will redirect correctly: https://pwastudio.io/peregrine/reference/hooks/useQuery/

belbiy: I think it's better to avoid redirects than have a pretty url in the search. It's not a big deal, but i just dont think redirecting is a good idea on a regular basis.

belbiy:
running tests

PR #5108, opened 2019-08-02
Disable external link checking

jeff-matthews:

This pull request (PR) disables external link checking because it's causing unexpected behavior that's blocking the test pipeline and therefore blocking production deployments.

html-proofer is identifying new mftf files as external links even though they're not.

This PR also fixes the PWA link, which is necessary after the repo move.


jeff-matthews:
I pushed a new commit after starting the test run, so I need to run it again.

jeff-matthews:
running tests

PR #5130, opened 2019-08-06
Update to supported PHP version

dobooth:

This pull request (PR) updates the supported version of PHP that 2.2.10 will support.
Missing Oxford comma caused the initial update to be incorrect.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.2/install-gde/prereq/php-settings.html


dobooth:
running tests

hostep:
I'm not in the loop with what's going to happen with 2.2.10, but according to this PR https://github.com/magento/magento2-sample-data/pull/124 it will still support PHP 7.0, can that be?

Rrego6:
7.0 is no longer officially supported as we will no longer test with 7.0. We will retain the 7.0 PHP dependency in order to support upgrades from older versions of Magento. However, merchants should upgrade to PHP 7.1.x

hostep:
Ok thanks for the explanation @Rrego6 !

PR #5144, opened 2019-08-08
[site] Review project implementation and tools

dshevtsov:

This pull request:
1. Organizes plugins into directories depending on their purpose
2. Moves the `lib` directory (that contains helper methods) from `rakelib` (that contains rake tasks) to the root of the project for more clarity, because the methods have been used not only in rake tasks.
3. Removes the `glossary-tooltip` plugin and converts all remaining tags to the recent format (`[term](https://glossary.magento.com/term)`)
4. Extends the `rake whatsnew` task with the `since` argument (`rake whatsnew since='jul 4'`) and adds a binstub for the whatsup_github tool to be able to run it with `bin/whatsup_github`.
5. Changes filenames to follow Ruby conventions
6. Excludes unneeded files from website


dshevtsov:
running tests

_config.yml
```diff
@@ -212,7 +212,7 @@ exclude:
   - bin
   - node_modules
   - vendor
-  - .git
+  - .github
```
jeff-matthews: `.git` is a hidden directory and can contain sensitive info. shouldn't we continue excluding it?

jeff-matthews: Actually, it doesn't matter because we already have S3 configured to redirect away from `.git/` for security purposes. Also, since none of the files in the `.git` directory include frontmatter, Jekyll won't include them in the output directory anyway. Disregard this comment :).

PR #5167, opened 2019-08-12
magento/devdocs#: Data Migration Tool. Data migration command. Preparation for delta step.

atwixfirster:

This pull request (PR) adds information about back-end processes when customer runs 

```
bin/magento migrate:data
```

command.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.2/migration/migration-migrate-data.html
- https://devdocs.magento.com/guides/v2.3/migration/migration-migrate-data.html

- [Migration\ResourceModel\Adapter\Mysql::createDelta](https://github.com/magento/data-migration-tool/blob/2.3-develop/src/Migration/ResourceModel/Adapter/Mysql.php#L378)

Thank you!


devops-devdocs:

rogyar:
Hi @atwixfirster. Thank you for providing that information. Considering the fact, that the raw SQL requests might be harder to understand for some people in comparison with the simplified verbal explanation, we should introduce both explanations: simple and extended. 

So, my suggestion is to put a few sentences about how the process works. For example:

> Additional tables will contain information about the changed records after the last migration execution. The database triggers are used to fill these extra tables, so once a new operation is being performed on the particular table (a record is added/modified/removed), the database trigger added by the Migration Tool saves information about this operation to the extra table.
If we run a delta migration process, the Tool checks these tables for the unprocessed records and ...

and put the details in SQL queries format below. So we have a polished rock-solid explanation of the process :)

Thank you!

atwixfirster:
> Please, check my suggestion above.

@rogyar , your suggestions have been implemented.

Could you please verify?

Many thanks!

dobooth:
running tests

meker12:
@dobooth  Missing whatsnew on this one.

PR #5213, opened 2019-08-19
GraphQL: update description of deprecated directive

keharper:

This pull request (PR) updates the description of the `deprecated` keyword.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/graphql/develop/create-graphqls-file.html

- ...

### DIFF
--- a/guides/v2.3/graphql/develop/create-graphqls-file.md
+++ b/guides/v2.3/graphql/develop/create-graphqls-file.md
@@ -99,11 +99,7 @@ url_key: String @doc(description: "The url key assigned to the product")
 product_count: Int @doc(description: "The number of products")
 ```
 
-Use the `@deprecated` directive to mark a query, mutation, or attribute as deprecated:
-
-```text
-@deprecated(reason: "description")
-```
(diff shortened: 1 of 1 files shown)


guides/v2.3/graphql/develop/create-graphqls-file.md
```diff
@@ -99,11 +99,7 @@ url_key: String @doc(description: "The url key assigned to the product")
 product_count: Int @doc(description: "The number of products")
 ```
 
-Use the `@deprecated` directive to mark a query, mutation, or attribute as deprecated:
-
-```text
-@deprecated(reason: "description")
-```
+Use the `@deprecated` directive to deprecate attributes and enum values. The GraphQL specification does not permit deprecating input values or arguments. The `reason` keyword allows you to specify which attribute or enum should be used instead.
```
avattam06: Can we say **fields** and enum values  here and in the sentence here "The `reason` keyword allows you to specify which **attribute/field** or enum **value** should be used

PR #5223, opened 2019-08-20
Link to new marketing PDF from release info page

jeff-matthews:

This pull request (PR) adds a link to a new marketing PDF that provides more detail about Magento release cycles. Magento Product Management requested that we add a link only rather than replace the contents of the devdocs page with this info.

## Affected DevDocs pages

- https://devdocs.magento.com/release/policy


jeff-matthews:
running tests

PR #5227, opened 2019-08-21
Remove email example from username credentials

ihor-sviziev:
Magneto admin user has separate fields - "username" and "email". These fields might be different.
Currently in docs for getting token for API we have example `customer1@example.com` that could be treated as email, but there actually should be username.

This pull request (PR) changes example username from `customer1@example.com` to `customer` to prevent people confusing.

## Affected DevDocs pages
- https://devdocs.magento.com/guides/v2.3/get-started/authentication/gs-authentication-token.html
- https://devdocs.magento.com/guides/v2.2/get-started/authentication/gs-authentication-token.html

- ...


devops-devdocs:

rogyar:
Hi @ihor-sviziev. Good catch ;)

ihor-sviziev:
Not sure if it’s clear enough, but looks like my fix fixing issue with
admin tokens, but adds not clear description for customer tokens.

Or it’s clear that there should be customer email?

PR #5297, opened 2019-08-30
Fixed broken link for Emogrifier README link

sanganinamrata:
## Purpose of this pull request

This pull request (PR) if to **Fix broken link for Emogrifier README link**

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/templates/template-email.html

PR #5295, opened 2019-08-29
updating output of this command bin/magento queue:consumers:list

aman3103:
## Purpose of this pull request

This pull request (PR) is updating the out of bin/magento queue:consumers:list mentioned in issue #5294 i.e https://github.com/magento/devdocs/issues/5294

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.3/config-guide/mq/manage-message-queues.html
- ...

## Links to Magento source code

- ...

PR #5292, opened 2019-08-29
Remove unsupported PHP versions from examples

Vinai:
## Purpose of this pull request

This PR makes the example `composer.json` code more usable as a template for custom modules.

References:
- https://www.php.net/supported-versions.php
- https://devdocs.magento.com/guides/v2.2/install-gde/system-requirements-tech.html#php
- https://devdocs.magento.com/guides/v2.3/install-gde/system-requirements-tech.html#php

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/extension-dev-guide/package/package_module.html
- https://devdocs.magento.com/guides/v2.2/extension-dev-guide/package/package_module.html

PR #5301, opened 2019-08-30
Small changes

jeff-matthews:
- #5205 
- #5292 
- #5295 
- #5297


guides/v2.2/extension-dev-guide/package/package_module.md
```diff
@@ -51,7 +51,7 @@ The following example is a `composer.json` for a metapackage:
     "version": "2.0.0",
     "type": "metapackage",
     "require": {
-        "php": "~5.5.0|~5.6.0|~7.0.0",
+        "php": "~7.1.3|~7.2.0",
```
dshevtsov: It is different for _magento2_ v2.2 though:
https://github.com/magento/magento2/blob/2.2/composer.json#L11

jeff-matthews:
running tests

guides/v2.2/extension-dev-guide/plugins.md
```diff
@@ -99,7 +99,7 @@ Use the following method names for the `_construct` method in the plugin class:
 
 Magento runs all before methods ahead of the call to an observed method. These methods must have the same name as the observed method with 'before' as the prefix.
 
-You can use before methods to change the arguments of an observed method by returning a modified argument. If there is more than one argument, the method should return an array of those arguments. If the method does not change the argument for the observed method, it should return `null`.
+You can use before methods to change the arguments of an observed method by returning a modified argument. If there is more than one argument, the method should return an array of those arguments. If the method does not change the argument for the observed method, it should return a `null` value.
```
dshevtsov: To be consistent with the change in next comment (`_after_ methods`), change to: 
`before methods` to `_before_ methods`

guides/v2.2/extension-dev-guide/plugins.md
```diff
@@ -137,7 +137,7 @@ class ProductAttributesUpdater
 }
 ```
 
-After methods have access to all the arguments of their observed methods. When the observed method completes, Magento passes the result and arguments to the next after method that follows. If observed method does not return a result (`@return void`), then it passes `null` to the next after method.
+The after methods have access to all the arguments of their observed methods. When the observed method completes, Magento passes the result and arguments to the next after method that follows. If the observed method does not return a result (`@return void`), then it passes a `null` value to the next after method.
```
dshevtsov: Change to:
`_after_ methods`

guides/v2.2/extension-dev-guide/plugins.md
```diff
@@ -236,9 +236,9 @@ class ProductAttributesUpdater
 }
 ```
 
-When you wrap a method which accepts arguments, your plugin must also accept those arguments and you must forward them when you invoke the <code>proceed</code> callable. You must be careful to match the default parameters and type hints of the original signature of the method.
+When you wrap a method which accepts arguments, your plugin must accept those arguments and you must forward them when you invoke the <code>proceed</code> callable. You must be careful to match the default parameters and type hints of the original signature of the method.
 
-For example, the following code defines a parameter of type <code>SomeType</code> which is nullable:
+For example, the following code defines a parameter of type <code>SomeType</code>, which is nullable:
```
dshevtsov: Replace HTML formatting with Markdown

guides/v2.2/extension-dev-guide/plugins.md
```diff
@@ -288,23 +288,23 @@ class MyUtilityUpdater
 
 ### Prioritizing plugins
 
-The `sortOrder` property for plugins determine when their before, after, or around methods get called when more than one plugins are observing the same method.
+The `sortOrder` property for plugins determines when to call the `before`, `around`, or `after` methods when more than one plugin is observing the same method.
```
dshevtsov: The before and after methods formatting is inconsistent with the changes above.

guides/v2.2/extension-dev-guide/package/package_module.md
jeff-matthews: thanks for catching this @dshevtsov. I'll break the symlink and restore the 2.2 version.

keharper: I almost said something about this, but I decided against, since it's unlikely anyone is developing new modules for 2.2 at this point.

guides/v2.2/extension-dev-guide/plugins.md
jeff-matthews: Done

guides/v2.2/extension-dev-guide/plugins.md
jeff-matthews: @dshevtsov, it looks like you're suggesting that we italicize the method names instead of using inline code format. Is that right? That's not consistent with the changes on line 291.

guides/v2.2/extension-dev-guide/plugins.md
keharper: I disagree with this. I don't know why after is italicized above. There are plenty of instances where before, after, and around are not italicized. I would remove the italics above.

guides/v2.2/extension-dev-guide/plugins.md
jeff-matthews: I changed all `before`, `after`, and `around` formatting to inline code for consistency.

guides/v2.2/extension-dev-guide/plugins.md
jeff-matthews: I changed all before, after, and around formatting to inline code for consistency.

guides/v2.2/extension-dev-guide/plugins.md
jeff-matthews: I changed all before, after, and around formatting to inline code for consistency.

guides/v2.2/extension-dev-guide/plugins.md
```diff
@@ -55,7 +55,7 @@ The following elements are optional:
 
 By applying code before, after, or around a public method, a plugin extends or modifies that method's behavior.
 
-The first argument for the before, after, and around methods is an object that provides access to all public methods of the observed method's class.
+The first argument for the before, after, and `around` methods is an object that provides access to all public methods of the observed method's class.
```
keharper:
```suggestion
The first argument for the `before`, `after`, and `around` methods is an object that provides access to all public methods of the observed method's class.
```

guides/v2.2/extension-dev-guide/plugins.md
```diff
@@ -116,13 +116,13 @@ class ProductAttributesUpdater
 }
 ```
 
-#### After methods
+#### `after` methods
 
-Magento runs all after methods following the completion of the observed method. Magento requires these methods have a return value and they must have the same name as the observed method with 'after' as the prefix.
+Magento runs all `after` methods following the completion of the observed method. Magento requires these methods have a return value and they must have the same name as the observed method with 'after' as the prefix.
```
keharper:
```suggestion
Magento runs all `after` methods following the completion of the observed method. Magento requires these methods have a return value and they must have the same name as the observed method with `after` as the prefix.
```

guides/v2.2/extension-dev-guide/plugins.md
hguthrie: So, typically one would emphasize a first occurrence for a definition. After that, it should not be emphasized. When it is used as an inline code, that is referring to the method in the code.

guides/v2.2/extension-dev-guide/plugins.md
hguthrie: Consistency for inline code format is preferred provided the reference is to the code. In some cases in this topic, it may not be referring to the code, but rather the definition of the method. You'd have to look at the topic holistically to make that kind of decision.

jeff-matthews: Well, this has been, thanks everyone. I'll revert my changes.

guides/v2.2/extension-dev-guide/plugins.md
```diff
@@ -191,26 +191,26 @@ class WebsitesLogger
...
+#### `around` methods
 
-Magento runs the code in around methods before and after their observed methods. Using these methods allow you to override an observed method. Around methods must have the same name as the observed method with 'around' as the prefix.
+Magento runs the code in `around` methods before and after their observed methods. Using these methods allow you to override an observed method. `around` methods must have the same name as the observed method with 'around' as the prefix.
 
 {:.bs-callout .bs-callout-warning}
-Avoid using around method plugins when they are not required because they increase stack traces and affect performance.
-The only use case for around method plugins is when the execution of all further plugins and original methods need termination.
-Use after method plugins if you require arguments for replacing or altering function results.
+Avoid using `around` method plugins when they are not required because they increase stack traces and affect performance.
```
keharper:
```suggestion
Avoid using `around` method plugins when they are not required, because they increase stack traces and affect performance.
```

guides/v2.2/extension-dev-guide/plugins.md
```diff
@@ -236,9 +236,9 @@ class ProductAttributesUpdater
 }
 ```
 
-When you wrap a method which accepts arguments, your plugin must also accept those arguments and you must forward them when you invoke the <code>proceed</code> callable. You must be careful to match the default parameters and type hints of the original signature of the method.
+When you wrap a method which accepts arguments, your plugin must accept those arguments and you must forward them when you invoke the <code>proceed</code> callable. You must be careful to match the default parameters and type hints of the original signature of the method.
```
keharper:
```suggestion
When you wrap a method that accepts arguments, your plugin must accept those arguments, and you must forward them when you invoke the `proceed` callable. You must be careful to match the default parameters and type hints of the original signature of the method.
```

guides/v2.2/extension-dev-guide/plugins.md
```diff
@@ -338,7 +338,7 @@ The execution flow will be as follows:
 
 Classes and interfaces that are implementations of or inherit from classes that have plugins will also inherit plugins from the parent class.
 
-Magento uses plugins defined in the global scope when the system is in a specific area (i.e. frontend, backend, etc). You can also extend or override these global plugin configuration via an area's `di.xml`.
+Magento uses plugins defined in the global scope when the system is in a specific area (i.e. frontend, backend, etc). You can extend or override these global plugin configurations with an area's `di.xml` file.
```
keharper:
```suggestion
Magento uses plugins defined in the global scope when the system is in a specific area (such as frontend or backend). You can extend or override these global plugin configurations with an area's `di.xml` file.
```

guides/v2.2/extension-dev-guide/plugins.md
jeff-matthews: Outdated

guides/v2.2/extension-dev-guide/plugins.md
jeff-matthews: Outdated

guides/v2.2/extension-dev-guide/plugins.md
jeff-matthews: Outdated

guides/v2.2/extension-dev-guide/plugins.md
jeff-matthews: Outdated

guides/v2.2/extension-dev-guide/plugins.md
jeff-matthews: Outdated

jeff-matthews: Actually, I was able to make that one.

jeff-matthews:
running tests

jeff-matthews:
I know there are some outstanding opinions about how we should handle the formatting for methods, but I think this is a global issue that isn't isolated to this topic. Also, the formatting is question doesn't necessarily enhance or diminish the fundamental meaning of these changes and is out of scope for this PR.

Issue #4647, opened 2019-05-30
Provide information about how to (re-)create mysql triggers


hostep:
# New topic request

## Description

This comes from a request over here: https://github.com/magento/magento2/issues/23023

It's not documented yet how mysql triggers are created in the database. It would be beneficial that this could be found easily so people who are missing triggers in their database know how to (re-)create them.

## Content checklist

Not sure where this could go, I think either in the [mysql section](https://devdocs.magento.com/guides/v2.3/install-gde/prereq/mysql.html), or in the [indexing section](https://devdocs.magento.com/guides/v2.3/extension-dev-guide/indexing.html)?

So to create triggers in the database, you have to make sure the indexers are set to "Update by Schedule". If they are already set to this mode and you are still missing triggers, you can first set them to "Update on Save" and then change them back to "Update by Schedule".

This change of indexer modes can be done in the backend using System => Index Management.
Or via the command line: `bin/magento indexer:set-mode <mode> [<indexer1> <indexer2> ...]` where `mode` is either `realtime` or `schedule`.

jeff-matthews:
Thanks @hostep. What do you think about enhancing the descriptions for both modes in the following topic: https://devdocs.magento.com/guides/v2.3/config-guide/cli/config-cli-subcommands-index.html#config-cli-subcommands-index-reindex?
 
By adding references to "triggers" and/or "missing triggers" specifically, people looking for this info should be able to find it easily using our site search.

hostep:
Hi @jeff-matthews 

Ah yes, that might be a good idea, but I would put it in the [Configure indexers](https://devdocs.magento.com/guides/v2.3/config-guide/cli/config-cli-subcommands-index.html#configure-indexers-1) section of that page, since that's where the `indexer:set-mode` command is being explained.

And then maybe also add a link to that section from [Mysql](https://devdocs.magento.com/guides/v2.3/install-gde/prereq/mysql.html#instgde-prereq-mysql-intro) page? In this bullet point for example:
> Magento uses MySQL database triggers to improve database access during reindexing. Magento does not support any custom triggers in the Magento database because custom triggers can introduce incompatibilities with future Magento versions.

Where we can mention that triggers are only created when the indexers mode are set to `Update by schedule` with a link to how to do that.

jeff-matthews:
[...]

PR #5337, opened 2019-09-08
information added for how to reinstall triggers

konarshankar07:

This pull request (PR) provide information about how can you reinstall triggers in mysql

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/config-guide/cli/config-cli-subcommands-index.html#configure-indexers-1

Fixed https://github.com/magento/devdocs/issues/4647

### DIFF
--- a/guides/v2.3/config-guide/cli/config-cli-subcommands-index.md
+++ b/guides/v2.3/config-guide/cli/config-cli-subcommands-index.md
@@ -19,7 +19,7 @@ bin/magento indexer:info
 
 The list displays as follows:
 
-```
+```terminal
 design_config_grid                       Design Config Grid
 customer_grid                            Customer Grid
 catalog_category_product                 Category Products
@@ -59,7 +59,7 @@ bin/magento indexer:status
 
 Sample result:
 
-```
+```terminal
 +----------------------+------------------+-----------+---------------------+---------------------+
 | Title                | Status           | Update On | Schedule Status     | Schedule Updated    |
 +----------------------+------------------+-----------+---------------------+---------------------+
@@ -106,7 +106,7 @@ bin/magento indexer:reindex
 
 Sample result:
 
-```
+```terminal
 Design Config Grid index has been rebuilt successfully in <time>
 Customer Grid index has been rebuilt successfully in <time>
 Category Products index has been rebuilt successfully in <time>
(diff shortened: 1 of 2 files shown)


devops-devdocs:

guides/v2.3/config-guide/cli/config-cli-subcommands-index.md
```diff
@@ -120,6 +120,20 @@ Product Price index has been rebuilt successfully in <time>
 Catalog Search index has been rebuilt successfully in <time>
 ```
 
+To reinstall triggers:
```
rogyar: Hi @konarshankar07. Thank you for providing the information. 
Currently, as a reader, I find this information a little bit confusing. I'm not sure what's a trigger, why I need to rebuild the triggers? So we a little bit out of context here. 

I would suggest extending the [following](https://devdocs.magento.com/guides/v2.3/config-guide/cli/config-cli-subcommands-index.html#configure-indexers-1) section with a mention that
> The indexers-related database triggers are being added when the indexer mode is set to `schedule` and removed when the indexer mode is set to `realtime`.

Thank you!

konarshankar07: Hello @rogyar ,
I have updated the content as per your above feedback. My question is how we can reference this in MySQL section which is mention in the task?
Thanks

hostep:
Thanks for taking this on @konarshankar07!

Looks good so far.
As for your question where to reference it from the [mysql section](https://devdocs.magento.com/guides/v2.3/install-gde/prereq/mysql.html#instgde-prereq-mysql-intro), maybe something like this:

Magento uses MySQL database triggers to improve database access during reindexing. **These get created when [the indexer mode is set to `schedule`](https://devdocs.magento.com/guides/v2.3/config-guide/cli/config-cli-subcommands-index.html#configure-indexers-1).** Magento does not support any custom tr...

The part in bold. This is just a suggestion 🙂

konarshankar07:
Hello @hostep ,
Thanks for the feedback. I'll update the changes for mention feedback but still as a user point of view its very difficult to find exact information which they are looking for like "reinstall triggers or reinstall MySQL triggers" might the be keywords user will search which is currently missing.

hostep:
@konarshankar07 

> but still as a user point of view its very difficult to find exact information which they are looking for like "reinstall triggers or reinstall MySQL triggers" might the be keywords user will search which is currently missing.

Maybe we should add this additional line then:
> The indexers-related database triggers are being added when the indexer mode is set to `schedule` and removed when the indexer mode is set to `realtime`.
**If the triggers are missing somehow from your database while the indexers are set to `schedule`, then a way to reinstall them is by first changing the indexers to `realtime` and afterwards changing them back to `schedule`. That will reinstate them.**

Just a suggestion again 🙂

konarshankar07:
Thanks @rogyar and @hostep for the feedbacks.

dobooth:
running tests

Issue #2370, opened 2018-06-26
RabbitMQ version recommended by M2.2 is EOF for more than 1 year


valugi:
## This issue is a:
- [ ] Bug on the website
- [ ] New topic request
- [ ] Topic clarification request
- [ ] New DevDocs feature request
- [x ] Other

## Description:
M2.2 [recommends](https://devdocs.magento.com/guides/v2.2/install-gde/prereq/install-rabbitmq.html) installing RabbitMQ 3.5.x, which [has finished](https://www.rabbitmq.com/which-erlang.html) his life in 2017-09-11 and its use is not recommended. Apart of this, recommended installation does not work and there are lots of problems with installing such old dependecies of erlang. 

### Steps to reproduce
Follow the installation steps on magento [rabbitmq page](https://devdocs.magento.com/guides/v2.2/install-gde/prereq/install-rabbitmq.html).

## Expected result:
A fully working RabbitMQ.

## Possible solutions:
Upgrade M2 to use more recent RabbitMq and Erlang.

jeff-matthews:
Related to #2369

lorikrell:
Not related to Cloud specifically. More system requirements for install.

shrielenee:
Hey @valugi - Are you installing on CentOS or Ubuntu?

valugi:
correct. Centos 7

dobooth:
From Misha via Slack " since we still will be releasing security patches for 2.2.x next year, I believe it makes sense to do the RabbitMQ upgrade for it too, so I will create a ticket for this"

PR #5345, opened 2019-09-10
Documentation updated for installing RabbitMQ on CentOS 7.x

konarshankar07:

This pull request (PR) provides information for installing RabbitMQ in CentOS 7.x

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/install-gde/prereq/install-rabbitmq.html#install-rabbitmq
Fixed https://github.com/magento/devdocs/issues/2370


devops-devdocs:

konarshankar07:
Hello @rogyar ,
Thanks for the feedback, I have tested this in CentOS 7 and `erlang` version is compatible but I'll try to install CentoOS 6 and update the documentation according to the latest version.
Thanks

konarshankar07:
Hello @rogyar ,
Any Updates?

dobooth:
I do not think we can take this PR. This is saying that 2.2 (and 2.3) support RabbitMQ 3.7, which is not true. While it is excellent news that you got Magento working with the latest version of Rabbit, this PR would say that Magento the company supports these versions, which is not true.
We have not officially tested v2.2 with RabbitMQ 3.7 so we cannot update the docs to say we do.

PR #5347, opened 2019-09-10
fix external mutagen link

hguthrie:
A quick fix in the Cloud guide Docker launch topic. Fixed the link to the Mutagen installation.
There were two instances of this link, but only one required the fix.


hguthrie:
running tests

PR #5372, opened 2019-09-12
MAGEDOC-4011: Added 2.3.3 release

jeff-matthews:

This pull request (PR) adds the 2.3.3 Commerce release to the compatibility page.

## Affected DevDocs pages

- https://devdocs.magento.com/release


jeff-matthews:
@alena-marchenko, is this correct?

jeff-matthews:
Thanks @hickskenh. I spoke to Olena and this page appears to be correct. However, we need to look into rolling off older versions of Magento to prevent the table from growing to an unreasonable size.

One proposal was to remove older versions after a critical security patch is released.

Another option is to somehow archive a version of the table for older versions of Magento. 

I'm going to merge this for now and we'll revisit with Product later.

jeff-matthews:
running tests

PR #5405, opened 2019-09-17
remove information that vcl export is optional

jonashrem:
The documentation states, that the export of the vcl file is optional but when you simple install varnish from your Linux distribution, it won't cache anything for magento unless the vcl file shipped with magento.

This pull request (PR) removed the information that vcl exporting is optional

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/config-guide/varnish/config-varnish-magento.html


devops-devdocs:

rogyar:
Can confirm that it's true. With the default Varnish VCL file, the full page cache for all pages will be bypassed. Magento VCL file with rules should be exported and applied.

keharper:
Thanks @jonashrem . We'll merge this in our next batch.

PR #5409, opened 2019-09-17
fixed repository name and more

hguthrie:
This PR is a bug fix for the bitbucket integration topic. It is clear in the topic that a user should "Be sure to use the name of your Bitbucket repository and not the URL. The integration fails if you use a URL."

However, there was an errant ".git" that threw people off. Removed this, but then I had to fix some markdown... you can see this fix on line 149.


dobooth:
running tests

PR #5425, opened 2019-09-19
GraphQL: Add fields and example showing cart rule discounts

keharper:

This pull request (PR) adds attributes and an example to the `cart` query, per internal ticket MC-18402.

## Affected DevDocs pages

- guides/v2.3/graphql/queries/cart.html

whatsnew
Added support for cart rule discounts in the GraphQL [Cart output object](https://devdocs.magento.com/guides/v2.3/graphql/queries/cart.html#CartItemPrices).

### DIFF
--- a/guides/v2.3/graphql/queries/cart.md
+++ b/guides/v2.3/graphql/queries/cart.md
@@ -17,13 +17,15 @@ Cart functionality is defined in the `Quote` module. A Quote represents the cont
 
 `{cart(cart_id: String!) {Cart}}`
 
-## Example usage
+## Sample queries
+
+### Cart ready for checkout
 
 The following query shows the status of a cart that is ready to be converted into an order.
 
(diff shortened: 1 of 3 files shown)


guides/v2.3/graphql/queries/cart.md
```diff
@@ -401,6 +403,117 @@ query {
 }
 ```
 
+### Cart discounts
+
+In this query, the **Buy 3 tee shirts and get the 4th free** cart price rule from the sample data is active. This rule was modified slightly to add the label `3T1free`.  (If a cart price rule does not have a label, Magento returns a default label of `Discount`.) A custom rule in which the customer saves 10% on the order by applying a discount code is also in effect.
```
jeff-matthews: tee shirt or t-shirt? the following paragraph is "t-shirt"

PR #5427, opened 2019-09-20
Snippet provided has sintax errors

mrodespin:
I modified the snipped because if you copy paste to the Magento Fastly Custom VCL Snippets configuration it will give you a sintax error due the \ in the code.


devops-devdocs:

PR #5446, opened 2019-09-21
Added missing git ignore file

konarshankar07:

This pull request (PR) added the missing git ignore file

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.2/config-guide/prod/config-reference-gitignore.html
- https://devdocs.magento.com/guides/v2.3/config-guide/prod/config-reference-gitignore.html


devops-devdocs:

konarshankar07:
Hello @rogyar ,
As per magento gitignore, `app/etc/config.php` is added [here](https://raw.githubusercontent.com/magento/magento2/2.2/.gitignore).

hostep:
@konarshankar07, the `.gitignore` file in the magento2 git repo is different than the one you get when setting up a new project using composer. The one and only difference (at the time of writing) is that particular `app/etc/config.php` line.
It makes sense for projects to not ignore that file, but for the magento2 repo, it should probably be ignored.

I also don't really like [that particular page](https://devdocs.magento.com/guides/v2.3/config-guide/prod/config-reference-gitignore.html), why would you need a different `.gitignore` file between different environments? Can't we just delete that page as a whole because it's just really weird, or maybe it's missing a lot of context about what it wants to convey?

konarshankar07:
Hello @hostep ,
You are right either we have to delete the page or we can update the content.I personally think we just need to update the content for gitignore file cause its really helpful for someone trying to find the gitignore file for the project
Thanks

hostep:
I'm not sure if people are actually using that kind of workflow where they have different `.gitignore` files for development & production (build servers in the last case I assume).

But I would not add `app/etc/config.php` to the `.gitignore` file, it's meant to be included in the project's git repository so it can be shared with developers and deployed to servers.

I don't know what the correct way is to move this PR forward here, if somebody else could chime in, that would be great! 🙂

konarshankar07:
Hello @hostep and @rogyar ,
We need to remove the multiple `.gitignore` as per @hostep suggested and add single `.gitignore` file which can be used for both Magento cloud and Magento CE and remove link of github `.gitignore` which remove the confusion of different `.gitignore` file or add some context why Magento have added `app/etc/config.php` in github `.gitignore` file. Let me know your thoughts
Thanks

hostep:
Hi

Does it make sense to keep a copy of the `.gitignore` file contents on devdocs do you think?
Because setting up a project using composer also gives you that default `.gitignore` file already?

Although thinking about it some more, it might make sense to keep a copy of the most recent version on devdocs, since the file is not automatically upgraded if you upgrade to a newer Magento version using composer as far as I'm aware.
We would then need to make sure to keep it up2date here on devdocs every time a new Magento version is released. Not that this file changes that often, but it's something to keep in mind...

So my conclusion: remove the confusing differences between "development" & "production" `.gitignore` file (unless somebody can come up with a good reason to keep it). And keep a copy of the `.gitignore` contents on devdocs from the file which gets generated by running `composer create-project --repository-url=https://repo.magento.com/ magento/project-community-edition`.

As far as Cloud goes, there is already a good reference on https://github.com/magento/magento-cloud/blob/master/.gitignore
Maybe we should remove the link `You can review the recommended files for your file in the .gitignore reference.` from [this page](https://devdocs.magento.com/guides/v2.3/cloud/reference/git-integration.html#gitignore) as it's currently completely different?
And then maybe also explicitly add a link from [here](https://devdocs.magento.com/guides/v2.3/config-guide/prod/config-reference-gitignore.html) to [here](https://github.com/magento/magento-cloud/blob/master/.gitignore) for Cloud users?

Thoughts?

konarshankar07:
Hello @hostep,
Sounds good. I'll make the first draft and let's see how it goes?
Thanks

hguthrie:
I'll be happy to chime in.

The Cloud guide briefly mentions ignoring files in the [Project structure topic](https://devdocs.magento.com/guides/v2.2/cloud/project/project-start.html?itm_source=devdocs&itm_medium=search_page&itm_campaign=federated_search&itm_term=gitignore#ignoring-files). To prevent constant maintenance on the DevDocs side, we provide a link to the `.gitignore` file in the development repo. We even provide an example of applying a specific version of the file to your local workstation.

I believe this topic would be much less maintenance for DevDocs if there was a link to the latest `.gitgnore` file for that version. You can still call out any nuances for ignoring certain entities if you find the need.

guides/v2.2/config-guide/prod/config-reference-gitignore.md
```diff
@@ -7,155 +7,9 @@ functional_areas:
   - Setup
 ---
 
-This reference shows suggested `.gitignore` files to use in a development system and in a build or production system.
+We include a base .gitignore file with the Magento Community Edition project repository. See the latest [link to the latest Magento `.gitignore`](https://raw.githubusercontent.com/magento/magento2/2.2/.gitignore).
```
hguthrie:
```suggestion
We include a base `.gitignore` file with the {{site.data.var.ce}} project repository. See [the latest Magento `.gitignore`](https://raw.githubusercontent.com/magento/magento2/2.2/.gitignore) file. 
```

guides/v2.3/config-guide/prod/config-reference-gitignore.md
```diff
@@ -7,170 +7,9 @@ functional_areas:
...
-/var/*
-!/var/.htaccess
-/vendor/*
-!/vendor/.htaccess
-# /generated/*
-!/generated/.htaccess
-.DS_Store</pre>
-
-{% endcollapsible %}
+We include a base .gitignore file with the Magento Community Edition project repository. See the latest [link to the latest Magento `.gitignore`](https://raw.githubusercontent.com/magento/magento2/2.3/.gitignore).
```
hguthrie:
```suggestion
We include a base `.gitignore` file with the {{site.data.var.ce}} project repository. See [the latest Magento `.gitignore`](https://raw.githubusercontent.com/magento/magento2/2.3/.gitignore) file. 
```

hguthrie:
running tests

hguthrie:
@dobooth passes local checks, ready to go

PR #5457, opened 2019-09-23
Fix Cloud procedures for Docker and integrations

hguthrie:
This replaces level 4 headers used for procedures with the new procedure class. The procedure class shifts the title to the right to align with the step numbering.

Some procedures have a single step, or a single step with response. I need feedback on how to handle these cases. You can find a sample of single step items in Health Notifications: /guides/v2.2/integrations/health-notifications.html

Also, line 41 in guides/v2.2/cloud/docker/docker-development.md would also be a single step, but I used an unordered list so that I could indent the items.

### DIFF
--- a/guides/v2.2/cloud/docker/docker-config.md
+++ b/guides/v2.2/cloud/docker/docker-config.md
@@ -78,7 +78,7 @@ For example, the following command starts the Docker configuration generator for
     ```
 
     {: .bs-callout-tip }
-To change the `magento2.docker` hostname for your project, you must update the host in three files: `.docker/config.php`, `docker-compose.yml`, and `/etc/hosts`
+    To change the `magento2.docker` hostname for your project, you must update the host in three files: `.docker/config.php`, `docker-compose.yml`, and `/etc/hosts`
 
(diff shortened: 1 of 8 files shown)


jeff-matthews:
In the case of single step procedures, especially the examples in the [Health Notifications](https://devdocs.magento.com/guides/v2.2/cloud/integrations/health-notifications.html) topic, I would expect the code block to be indented as well.

hguthrie:
> In the case of single step procedures, especially the examples in the [Health Notifications](https://devdocs.magento.com/guides/v2.2/cloud/integrations/health-notifications.html) topic, I would expect the code block to be indented as well.

We also have single steps that include multiple lines of explanation or sample responses. Another, more time-consuming solution would be to improve the statement and not rely on procedural emphasis. The following example USED TO BE a procedure that began with "To list blah blah":

Screen Shot 2019-09-23 at 14 09 32

hguthrie:
Decided, if it does not have multiple steps, then work it into the content appropriately without depending on a title. In other words, stop being lazy. 😬

Issue #5252, opened 2019-08-22
Markdown linting: Fenced code blocks should have a language specified (MD040)


dshevtsov:
# Bug report

## Description

The Markdown linter reports about 1421 issues when the [rule 'MD040']( https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md040---fenced-code-blocks-should-have-a-language-specified) is enabled.

### Steps to reproduce

1. In the `_checks/styles/style-rules-prod` change `exclude_rule 'MD040'` to `rule 'MD040'`.
2. In your terminal, run `rake test:md`.

## Test your updates

1. Follow "Steps to reproduce" locally to get list of errors to fix.
2. Preview the HTML locally to ensure your changes did not break formatting in the resulting HTML (run `rake:preview`). For more information, refer to the [README](https://github.com/magento/devdocs#to-build-locally).

**Note**-->Do NOT use [special keywords](https://help.github.com/en/articles/closing-issues-using-keywords#about-issue-references) in your commit message.

## Expected result

All the reported issues are fixed.
The `rake test:md` command returns `No issues found`.

jeff-matthews:
This rule flags indented code blocks as errors; not just fenced code blocks with no language specified. See the following example:

<img width="1031" alt="Screen Shot 2019-09-19 at 10 32 28 AM" src="https://user-images.githubusercontent.com/13662379/65258814-30971480-dac9-11e9-8488-3ddb657c0ead.png">

That's going to make it more difficult to enable this rule, especially when we have a separate issue for enforcing the fenced code block style (#5253).

shrielenee:
@danidelcar This will affect OMS Docs significantly when/if this is merged to our repo. I think I brought this up to you recently when doing a review.

jeff-matthews:
There are still a lot of errors to fix, but there are two that are in the MRG:

```
guides/v2.3/mrg/ce/InstantPurchase.md:35: MD040 Fenced code blocks should have a language specified
guides/v2.3/mrg/ce/InstantPurchase.md:43: MD040 Fenced code blocks should have a language specified
```

jeff-matthews:
PR #5478 brings the error count down to ~`265`.

jeff-matthews:
After #5495, #5497, and #5498 are merged, we can turn on the MD040 rule in production.

PR #5458, opened 2019-09-23
Fixed MD040 errors in _includes/config/

jeff-matthews:

This pull request (PR) fixes all MD040 errors in the `_includes/config/` directory.

The MD040 rule requires that a language be specified for fenced code blocks. However, it also identifies indented code blocks, so you may see changes in this PR that replace indentation with fences.

This is one of many PRs related to #5252.

### DIFF
--- a/_includes/config/secure-ws-apache_step1.md
+++ b/_includes/config/secure-ws-apache_step1.md
@@ -6,9 +6,12 @@ First, see if you have the Apache `htpasswd` utility is installed as follows:
 
 1.	Enter the following command to determine if `htpasswd` is already installed:
 
-		which htpasswd
+    ```bash
+    which htpasswd
+    ```
 
-	If a path displays, it is installed; if the command returns no output, `htpasswd` is not installed.
+    If a path displays, it is installed; if the command returns no output, `htpasswd` is not installed.
+
 2.	If necessary, install `htpasswd`:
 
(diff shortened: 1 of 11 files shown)


_includes/config/secure-ws-apache_step1.md
```diff
@@ -18,8 +21,10 @@ First, see if you have the Apache `htpasswd` utility is installed as follows:
 
 Enter the following commands as a user with `root` privileges:
 
-	mkdir -p /usr/local/apache/password
-	htpasswd -c /usr/local/apache/password/.<password file name> <username>
+```bash
+mkdir -p /usr/local/apache/password
+htpasswd -c /usr/local/apache/password/.<password file name> <username>
```
keharper: Separate lines here.

Larger question. Maybe the rendering for `bash` can be adjusted so that every line within the code block gets a `$`?

dobooth: I was wondering the same thing just this morning.

dshevtsov: 1. Each command executes different Unix tools, so each one expects some return and should be described separately.
2. When I _Copy_ the script, I want control it, and run step-by-step, not all at once. I also may want to edit each command with my data.

When you're running more than one command at a time, you should run it as a Shell script actually.

PR #5476, opened 2019-09-25
Fix procedures in Cloud setup and live topics

hguthrie:
This replaces level 4 headers used for procedures with the new procedure class. The procedure class shifts the title to the right to align with the step numbering.

This PR addresses the topics in the Setup and Live directories, and other remaining procedure issues.

### DIFF
--- a/guides/v2.2/cloud/live/go-live-checklist.md
+++ b/guides/v2.2/cloud/live/go-live-checklist.md
@@ -86,5 +86,5 @@ You can also test using the following 3rd party options:
 * [Blackfire]({{ page.baseurl }}/cloud/project/project-integrate-blackfire.html) (provided): Helps track through the issues New Relic finds and helps you dig deeper into the issue for specifics. Blackfire profiles the environment and helps locate bottlenecks in depth: process, method call, query, load, and so on.
 * [WebPageTest](https://www.webpagetest.org/) and [Pingdom](https://www.pingdom.com/): Real-time analysis of your site pages load time with different origin locations. Pingdom may cost a fee. WebPageTest is a free tool.
 
-#### Next step:
+**Next step:**
 [Launch steps]({{ page.baseurl }}/cloud/live/launch-steps.html)
--- a/guides/v2.2/cloud/live/stage-prod-migrate.md
+++ b/guides/v2.2/cloud/live/stage-prod-migrate.md
@@ -6,8 +6,7 @@ functional_areas:
   - Deploy
 ---
 
-#### Previous step:
-
+**Previous step:**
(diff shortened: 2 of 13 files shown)


guides/v2.2/cloud/live/stage-prod-migrate.md
```diff
@@ -144,15 +143,14 @@ This command uses the following options:
 
 For additional options, see the [rsync man page](http://linux.die.net/man/1/rsync).
 
-#### To migrate static files from your local machine:
-
-Use the rsync command to copy the `pub/media` directory from your local Magento server to staging or production:
+To migrate static files from your local machine, use the `rsync` command to copy the `pub/media` directory from your local Magento server to staging or production:
```
jeff-matthews: I'm assuming that since you expanded the description that you're intentionally omitting the procedure class.

guides/v2.2/cloud/live/go-live-checklist.md
```diff
@@ -86,5 +86,5 @@ You can also test using the following 3rd party options:
 * [Blackfire]({{ page.baseurl }}/cloud/project/project-integrate-blackfire.html) (provided): Helps track through the issues New Relic finds and helps you dig deeper into the issue for specifics. Blackfire profiles the environment and helps locate bottlenecks in depth: process, method call, query, load, and so on.
 * [WebPageTest](https://www.webpagetest.org/) and [Pingdom](https://www.pingdom.com/): Real-time analysis of your site pages load time with different origin locations. Pingdom may cost a fee. WebPageTest is a free tool.
 
-#### Next step:
+**Next step:**
```
jeff-matthews: It's a shame you have to do this manually. Out of scope for this PR, but you might want to consider leveraging the layout for tutorials to solve for the "previous" and "next" on page navigation.

guides/v2.2/cloud/live/stage-prod-migrate.md
meker12: Just to share and confirm my understanding: @hguthrie and discussed the procedure label, and we established a convention that one step instructions didn't require the procedure heading.  Instead, we use an introductory sentence ("To do this thing:") with no special formatting.

guides/v2.2/cloud/live/go-live-checklist.md
hguthrie: Yes, you're right. I think these topics will go away, though. I didn't want them to break the MD001 effort. Cloud doesn't have many of these left.

guides/v2.2/cloud/live/stage-prod-migrate.md
hguthrie: Yes, in the case of single steps, we decide to incorporate them better into the content. It's quick and dirty. LOL

PR #5494, opened 2019-09-26
[site] Move source of the Module Reference Guide topics to _data

dshevtsov:

This pull request reimplements the Module Reference Guide topics generation. It removes the topics from a source filesystem, and instead, implicitly generates the topics from files at `_data/codebase/mrg`.

## Affected DevDocs pages

- module-specific topics at the [Module Reference Guide](https://devdocs.magento.com/guides/v2.3/mrg/intro.html)

### Reasons

1. General concept is to keep all autogenerated content in `_data`, when possible and reasonable.
2. Direct contributions are not allowed. No changes are allowed to the source files, because the original content is generated from the actual magento codebase.
3. Markdown linting. Markdown style that is specific to devdocs is very strict and should not be enforced to codebase.

### Internal references

Staging build: 881


_data/toc/module-reference-guide-2_3.yml
```diff
@@ -1,7 +1,8 @@
-label: Magento Reference Guide
+label: Module Reference Guide
 pages:
-  - label: Introduction
+  - label: INTRODUCTION
```
jeff-matthews: We don't want this appearing as all caps in the HTML.

dshevtsov: fixed

dshevtsov:
running tests

PR #5498, opened 2019-09-26
Fixed MD040 errors in cloud topics

jeff-matthews:

This pull request (PR) fixes most MD040 errors in the Cloud Guide.

The MD040 rule requires that a language be specified for fenced code blocks. However, it also identifies indented code blocks, so you may see changes in this PR that replace indentation with fences.

This PR doesn't address YAML code blocks because I'm not sure exactly how to fix those without breaking the code samples when users copy/paste. I know that the spacing is important, so deferring on those for now. 

@meker12 & @hguthrie, please provide guidance on fixing the YAML code blocks. 

The only cloud topic that has unfenced YAML code blocks is `guides/v2.2/cloud/howtos/debug.md`.

This is one of many PRs related to #5252.


guides/v2.2/cloud/live/sens-data-initial.md
```diff
@@ -120,6 +130,7 @@ The following snippet from `config.php` shows an example of changing the default
 
      ... more ...
 ```
+
```
hguthrie: Does no-copy still work when you add a line? I didn’t realize...

jeff-matthews: Agh... no it does not. I thought it might since the collapsible tag works with a space, but I got lazy and neglected to test first. Thank you for catching. This will br a problem for rule MD031 (blanks around fences),

jeff-matthews:
running tests

guides/v2.2/cloud/live/sens-data-initial.md
hguthrie: We need at least one way to turn off the copy ability. I'm open to other solutions.

jeff-matthews: I agree. Rule MD031 is relatively unimportant, so we may need to ignore it.

PR #5517, opened 2019-09-30
Removed mcrypt from extension and noted use of sodium

### LINKED ISSUE #5490: Feedback on page: /guides/v2.3/performance-best-practices/software.html
# General issue

## Description:

mcrypt php extensions isn't available on php 7.2, it was replaced to sodium

dobooth:

This pull request (PR) fixes #5490. mcrypt has been removed from PHP.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/performance-best-practices/software.html


guides/v2.3/performance-best-practices/software.md
```diff
@@ -50,7 +50,6 @@ We recommend limiting the list of active PHP extensions to those that are requir
 * `php-gd`
 * `php-intl`
 * `php-mbstring`
-* `php-mcrypt`
```
osrecio: you don't should add `php-sodium` ?

hostep: Strictly speaking it's not required, Magento works without php-sodium, but it will be less secure since the password hashing will not use the argon2id algorithm.

However it should be encouraged to install it I think.

dobooth: For PHP 7.2, mcrypt has been removed entirely. It now ships with and uses sodium by default. We can't use or recommend EOL software.

osrecio: Nope, we're talking about add `php-sodium`. I mean. You're deleting `php-mcrypt` that's ok. But we should add `php-sodium`. I've explained myself correctly?

hostep: Yeah, for security's sake, I would add it as well 👍

Since this is a performance topic, imagine that somebody would first have `php-sodium` installed, than saw this list of extensions and starts thinking: "hmm, sodium is not required, let me remove it, my shop might run a tiny bit faster" and by doing so, suddenly all the passwords in his database would no longer work.

dobooth: Please correct me if I am wrong, but we are removing mcrypt from a list of extensions/modules that need to be specifically enabled to function. Now that Sodium is a built-in module, it is enabled by default and therefore no longer needs to be enabled specifically in php.ini.

hostep: It depends on the operating system and package manager if sodium is included by default in php or not. For example on the package manager I'm personally using (MacPorts), `php-sodium` needs to be installed as an extra extension, it's not part of php itself.
I think Ubuntu also works like that, but not 100% sure. And probably some others as well.

osrecio: According to php documentarion for php 7.2 is not needed: https://www.php.net/manual/en/sodium.installation.php.

But this is a list of extensions in PHP, and you need to ensure that you have sodium.

Tomorow I will test in Centos 7.

dobooth:
running tests

dobooth:
Thanks for the feedback, gents. I appreciate it.

dobooth:
running tests

PR #5534, opened 2019-10-01
Updated Magento Testing Strategy

nathanjosiah:

Implementation of Acceptance Criteria from https://github.com/magento/architecture/pull/277 to describe a consolidated and authoritative Definition of Done for all of Magento Commerce and Community based on the approved testing strategy. This includes changes to testing requirements as well as definitions and examples for various testing aspects and terminology.

This strategy has been reviewed with many layers and members of management, architecture, QA, and developers.

**Important note before reading on** The P0/P1 test scenarios will be published to the devdocs in the near future and will serve as the authoritative list for both internal and external developers. Also, we currently have a communications gap between community members trying to deliver code and the product owners that own the affected code. We need to change our culture to bridge this gap and create some channel of communication instead of compromising the quality of our product. This will be key in fulfilling this strategy.

### Problem

Magento test coverage is sporadic and somewhat arbitrary and our testing strategy is unconventional, not very effective, and inconsistent/conflicting across 4 sets of testing requirements.

Despite running up to 30,000 tests per pull request, the release team still requires a costly manual regression cycle to verify that the results of the automated test suite is valid. This instability and lack of confidence in our build suite also contributes to a painful pull request process for many developers and leads to many reruns of builds to achieve a stable result. To even further the point, tests that can't pass via rerunning are ultimately skipped to be fixed at a later time.

### Solution

Implement a new incremental testing strategy that focuses on layered coverage with each level containing decreasing levels of granularity and quantity of tests.

This will allow us to focus our test coverage on test coverage that is easier to write, cheaper and easier to maintain, inherently more stable to run, and provides a more comprehensive and consistent approach to coverage and quality in Magento.

Eventually this will allow us to rely on our automated builds as they will represent comprehensive coverage at appropriate scopes and allow us to release more quickly.

There are many initiatives in progress and planned to improve other aspects of quality but they rely on this being implemented first.

### Caveats

**Test extensibility** 

The issue of extensibility for MFTF vs non-MFTF tests ultimately comes down to the issue of declarative vs imperative testing frameworks. 

It's important to understand that, as described within this PR, MFTF tests only provide system-level behavioral coverage of the final packaged application. These tests indirectly provide test coverage to whatever implementations are loaded in the system at the time the tests are run. 

For these reasons, MFTF tests can be though of more as behavioral contracts that ensure that regardless of what loads, certain consumer workflows can be followed. And also for these reasons, it makes sense that these are declarative because the packages that are loaded may introduce new user-level features that should be tested as part of the user-level behavior flows.

By contrast,

Other test types are realistically only testing lower-level concrete implementations. Lower-level tests are only able to interact with code at a low level. Even if the lowest level of interaction is a high-level abstraction such as a controller, that controller is hardcoded into that test. But this is also the nature of the code that these tests are testing. 

But, this is fine. There needs to be coverage of the low level components that invoke them more thoroughly than the ui can invoke them. For many reasons this is true but even just consider the simple case where some lower-level component isn't even used indirectly from the UI but is expected to be used from an extension. 

**Service Isolation**

Another concern that has been made regarding this strategy is what this will mean for the upcoming impact of the service isolation initiative. The design and exact impact of this initiative are unclear. The details surrounding how tests can and will be written are unclear. There are still many details to be decided for this initiative but one thing is very clear, we need to be able to test that our components still work as they are refactored. Having a high-level test suite that passes is great and will be helpful as well. But we will need more comprehensive coverage than that as we massively refactor modules and classes.

"But what about controller tests?" I think the more important question is "what about controllers?". 
 Due to the heavy amounts of logic and mixed application layers and simply poor design decisions especially in more legacy code, this code doesn't have a place in the new service isolation infrastructure and will need to be rewritten into services or simply thrown away. So what about the controller tests? Well the answer is pretty much the same. The exact fate of these is yet to be determined as the implementation of the API Gateway replacement for controllers is still under heavy discussion. It's more likely the tests will be more compatible with the new infrastructure than the controllers themselves but that remains to be seen. Either way, it is important that the scope of the new tests within this new testing strategy align as close to the "Narrow form integration tests" definition in this PR as possible in order to maintain forward compatibility in respect to service isolation.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/contributor-guide/contributing_dod.html


devops-devdocs:

guides/v2.2/contributor-guide/contributing_dod.md
```diff
@@ -57,74 +57,113 @@ Any backward-incompatible changes must also be recorded in the accompanying docu
...
 ### Functional Tests
 
-Must cover new or changed application behavior (functional).
-Added/updated functionality should be covered by a functional autotest(s) related to previous sprint commitment in scope of current sprint.
+UI functional tests are inherently unstable irrespective of platform or testing framework. For this reason along with reasons of maintainability and quality assurance delivery time, these tests should be prioritized below all other tests types.
+
+Functional test policy: 
+* Web API endpoints must have functional test coverage. These tests should ensure that the endpoints behave in accordance to their service contracts regardless of the actual concrete implementation that may be loaded.
+* For UI-related code, there are some circumstances where it is not possible to cover changes with unit or integration tests. In these cases the appropriate type of coverage would be a functional test but this is a last resort only. 
+  However, aside from those cases UI functional tests should only be used to cover P0/P1 testing scenarios. If there are no scenarios available to make this decision, a product owner must be consulted for approval of new tests.
```
joanhe: I don't think it is good idea to restrict adding tests. We want to increase test coverage. If developer and/or QA determine function test is good type of test to ensure coverage of product change. We shouldn't prevent the test to be added by extra approval process or just because the scenarios doesn't have high enough priority.

nathanjosiah: Hey @joanhe, I know this has already been discussed offline but I'll comment here for the sake of history:

This goes against one of the main goals of this initiative. Adding additional more expensive coverage for unnecessary flows creates more of the problem we are trying to solve. If there is a scenario new or old that isn't classified as a P0/P1, it can be discussed with a product owner to get it elevated if it is important enough to deserve a dedicated test.

guides/v2.2/contributor-guide/contributing_dod.md
```diff
@@ -57,74 +57,115 @@ Any backward-incompatible changes must also be recorded in the accompanying docu
...
+  For example, `Magento\Framework\SomeClassFilter` may contain a `FilterPool` that comes with default `FilterInterface`'s from `Magento\Framework`. Each of these implementations would have their own coverage pursuant to this document. However, there should also be some basic assertions within the test coverage for `Magento\Framework\SomeClassFilter` that ensure each of the default filters are loaded correctly.     
+  This shouldn't be explicitly coverage such as `$filter->isLoaded('someDefaultFilter')` 
+
+See: [Running Integration Tests][3].
 
 ### Functional Tests
 
-Must cover new or changed application behavior (functional).
-Added/updated functionality should be covered by a functional autotest(s) related to previous sprint commitment in scope of current sprint.
+UI functional tests are inherently unstable irrespective of platform or testing framework. For this reason along with reasons of maintainability and quality assurance delivery time, these tests should be prioritized below all other tests types.
```
irenelagno: I think it is reasonable to use functional tests to cover user behavior scenarios. When we are talking about the user story, it makes sense to cover S0, S1 scenario with UI tests. About maintainability, it can be improved by refactoring test-related entity, removing code duplication.

guides/v2.2/contributor-guide/contributing_dod.md
```diff
@@ -57,74 +57,115 @@ Any backward-incompatible changes must also be recorded in the accompanying docu
...
+For the purpose of Magento testing, there are essentially two broad categories of integration tests: narrow-form and broad-form.
+
+* Narrow-form integration tests focus on testing something mostly in isolation an can be as simple as a unit test as described in this guide except without mocked dependencies.
+* Broad-form integration tests focus on testing something with less isolation and will involve multiple components explicitly collaborating. (compared to narrow-form where the dependencies are implicit collaborating)
+
+Integration test policy:
+
+* Methods and classes must have blackbox or greybox test coverage making sure to include all variations for normal use as well as corner-cases. 
+* SPI's must have test coverage. Often times Magento has extension points that may only be utilized by extension developers. Use TestModule's to implement those SPI extension points and verify they are used correctly. 
+* As mentioned above, classes marked with `@api` must contain coverage. Integration tests must be used cover these classes unless a unit test is more appropriate.
```
irenelagno: in this case, could you add the clarification that we should use js unit tests for @api marked js files

guides/v2.2/contributor-guide/contributing_dod.md
```diff
@@ -57,74 +57,115 @@ Any backward-incompatible changes must also be recorded in the accompanying docu
...
+* Consumers of default SPI's implementations should have at least basic coverage that ensures the default implementations of the SPI are correctly configured and loaded.  
+  
+  For example, `Magento\Framework\SomeClassFilter` may contain a `FilterPool` that comes with default `FilterInterface`'s from `Magento\Framework`. Each of these implementations would have their own coverage pursuant to this document. However, there should also be some basic assertions within the test coverage for `Magento\Framework\SomeClassFilter` that ensure each of the default filters are loaded correctly.
```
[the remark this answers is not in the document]
nathanjosiah: Maybe this is a bad example or could be worded better. However, I don't think the logic changes here as it scales. We still want to verify behavior of known collaborators with 10,000 filters. We would load them all anyway in the production system so if it's too slow for our test it's going to be too slow for production. 

Also the tests would be written in their respective modules. ModuleA defines the FilterPool and ModuleB defines FilterA (which of course then has a dependency on ModuleA inherently). FilterA would have isolated tests but we would have a test inside of ModuleB that loads FilterA inside of a ModuleA's FilterPool. Since this is how it would work in production, we would want to verify that it actually behaves the way we expect.

guides/v2.2/contributor-guide/contributing_dod.md
nathanjosiah: Yes, it is still required to cover the P0/P1 scenarios with functional test coverage. UI test coverage in the case of user behavior. These scenarios have to be prioritized and defined by the product owner who should have knowledge of which behavioral flows are important enough to deserve dedicated system-level blackbox tests.

guides/v2.2/contributor-guide/contributing_dod.md
nathanjosiah: Good idea.

guides/v2.2/contributor-guide/contributing_dod.md
```diff
@@ -57,74 +57,115 @@ Any backward-incompatible changes must also be recorded in the accompanying docu
...
+UI functional tests are inherently unstable irrespective of platform or testing framework. For this reason along with reasons of maintainability and quality assurance delivery time, these tests should be prioritized below all other tests types.
+
+Functional test policy: 
+* Web API endpoints must have functional test coverage via api-functional tests. These tests should ensure that the endpoints behave in accordance to their service contracts regardless of the actual concrete implementation that may be loaded.
+* For UI-related code, there are some circumstances where it is not possible to cover changes with unit or integration tests. In these cases the appropriate type of coverage would be a functional test but this is a last resort only. 
+  However, aside from those cases UI functional tests should only be used to cover P0/P1 testing scenarios. If there are no scenarios available to make this decision, a product owner must be consulted for approval of new tests.
+  
+  If there is a scenario new or old that isn't classified as a P0/P1, it can be discussed with a product owner to get it elevated if it is important enough to deserve a dedicated test.
```
xmav: Might not work well with community contributions as it might be problematic to discuss it with product owner
@okorshenko Could you please provide feedback on this ?

okorshenko: Hi @xmav, you are right. Community members don't have access to test cases and their status. Also, it is very hard to contact PO for them.  There should be an exemption for community members.

nathanjosiah: @xmav @okorshenko We will be publishing our list of P0/P1 scenarios to the devdocs as the authoritative list for internal and external developers. As for how to contact the PO, we need to create a culture where this is not true. It doesn't make sense to separate the developers that contribute from the owners of the code. If there is a communications gap we should fix it and not compromise our quality.

guides/v2.2/contributor-guide/contributing_dod.md
```diff
@@ -57,74 +57,115 @@ Any backward-incompatible changes must also be recorded in the accompanying docu
...
-Continuous integration enforces execution of all tests and author is accountable for broken builds.
+Be aware that while high-level tests may provide coverage to code, it is only indirect coverage. Tests with more direct usage of the changed code will likely still needs to be written to ensure that regardless of the which components are actually used in the black box, the individual components still have coverage.  
+
+High level testing policy:
+ 
+* Before committing code changes, author must ensure successful execution of all tests by running all tests or at least those which might be affected by code changes.
+  Continuous integration enforces execution of all tests and author is accountable for broken builds.
+* Any code that is added or changed must have explicit tests to validate the behavior. To clarify, this policy is intended to prove that the integrity of the code is verified regardless of which implementations are loaded in a black-box environment. This may be done through unit or integration testing as appropriate.
```
danmooney2: May be confusing to hyphenate blackbox

guides/v2.2/contributor-guide/contributing_dod.md
```diff
@@ -57,74 +57,115 @@ Any backward-incompatible changes must also be recorded in the accompanying docu
...
+* SPI's must have test coverage. Often times Magento has extension points that may only be utilized by extension developers. Use TestModule's to implement those SPI extension points and verify they are used correctly. 
+* As mentioned above, classes marked with `@api` must contain coverage. Integration tests must be used cover these classes unless a unit test is more appropriate. This includes JS modules and components. 
+* Consumers of default SPI's implementations should have at least basic coverage that ensures the default implementations of the SPI are correctly configured and loaded.  
+  
+  For example, `Magento\Framework\SomeClassFilter` may contain a `FilterPool` that comes with default `FilterInterface`'s from `Magento\Framework`. Each of these implementations would have their own coverage pursuant to this document. However, there should also be some basic assertions within the test coverage for `Magento\Framework\SomeClassFilter` that ensure each of the default filters are loaded correctly.     
+  This shouldn't be explicitly coverage such as `$filter->isLoaded('someDefaultFilter')`
```
danmooney2: Confusing to read

guides/v2.2/contributor-guide/contributing_dod.md
nathanjosiah: I meant to go back and fix this. Thank you!

guides/v2.2/contributor-guide/contributing_dod.md
```diff
@@ -57,74 +57,115 @@ Any backward-incompatible changes must also be recorded in the accompanying docu
...
+  However, aside from those cases UI functional tests should only be used to cover P0/P1 testing scenarios. If there are no scenarios available to make this decision, a product owner must be consulted for approval of new tests.
+  
+  If there is a scenario new or old that isn't classified as a P0/P1, it can be discussed with a product owner to get it elevated if it is important enough to deserve a dedicated test.
 
 See [Functional Tests][2].
 
 ### Unit Tests
 
-Cover classes and methods that:
+There are a small number of use cases for unit tests in Magento. The nature of our code and backwards compatibility practices make it increasingly hard to write and maintain high-quality unit tests.
```
buskamuza: I don't think it is related to backward compatibility.
Is this explanation necessary in DoD document? IMO, such important document should be as short as possible, so everybody can easily read and remember main points. I'd say, it should only have main points. Explanations (that should be read only once or maybe couple times) can be moved to separate documents and references to those can be put here.

guides/v2.2/contributor-guide/contributing_dod.md
```diff
@@ -57,74 +57,115 @@ Any backward-incompatible changes must also be recorded in the accompanying docu
...
+  - This can be hard to predict but a good example of when it would be unlikely is a simple utility class that performs some standalone operation. By contrast, a helper would be very likely to have dependencies by nature. 
 
 Examples:
 
-* Utils (like `\Magento\Framework\Math\Random`)
-* Simple classes that can also be used (at least initiated) independently (like `\Magento\Framework\Api\SortOrder`)
-* Algorithms that perform calculation, parsing
+* Utility classes (like `\Magento\Framework\Math\Random`).
+* Simple classes that can also be used independently (like `\Magento\Framework\Api\SortOrder`).
+* Simple logic within a Data Transfer Object.
```
buskamuza: Are there examples? Most likely, DTOs should not have any logic.

guides/v2.2/contributor-guide/contributing_dod.md
```diff
@@ -6,7 +6,7 @@ title: Magento Definition of Done
 ## Overview
 
 The "Definition of Done" (DoD) is a collective term for a set of acceptance criteria that is applied to any changes in the product code base.
-The requirements that result from our DoD are applied by default to all user stories and to each task within this user story.
+The requirements that result from our DoD are applied by default to all user stories and to each task within this user story as well as for bug fixes.
```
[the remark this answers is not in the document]
nathanjosiah: This is a good point. I will update it to include more general wording.

guides/v2.2/contributor-guide/contributing_dod.md
nathanjosiah: There are already dedicated documents to each type of test but the DoD needs to explain the critical concepts that define when something is considered to have enough coverage. I can remove this particular note as I'm not sure what I meant by this.

guides/v2.2/contributor-guide/contributing_dod.md
nathanjosiah: To be honest I was having a hard time with giving good examples for unit tests in Magento and any concrete examples would likely reference undesirable design. I can just remove this example if it would be better unless you have more to add?

guides/v2.2/contributor-guide/contributing_dod.md
```diff
@@ -57,74 +57,115 @@ Any backward-incompatible changes must also be recorded in the accompanying docu
...
 When choosing how to cover your changes, pick the most lightweight (execution time wise) test type that will provide sufficient coverage.
 If you encounter an existing test that insufficiently covers your changes, you can delete that test but you must write a proper test to replace it. For example, a method that interacts with database has a unit test. You can replace it with an integration test.
 
-Before committing code changes, author must ensure successful execution of all tests by running all tests or at least those which might be affected by code changes.
-Continuous integration enforces execution of all tests and author is accountable for broken builds.
+Be aware that while high-level tests may provide coverage to code, it is only indirect coverage. Tests with more direct usage of the changed code will likely still needs to be written to ensure that regardless of the which components are actually used in the black box, the individual components still have coverage.  
+
+High level testing policy:
+ 
+* Before committing code changes, author must ensure successful execution of all tests by running all tests or at least those which might be affected by code changes.
```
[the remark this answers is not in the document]
buskamuza: I'd just remove this item.

guides/v2.2/contributor-guide/contributing_dod.md
buskamuza: I'm ok with removing this one.

nathanjosiah:
@dobooth @shrielenee Who can run the required builds for this PR?

shrielenee:
@nathanjosiah We just need an approving review, then the builds will automatically run. Looks like @dobooth is on this as far as reviewing go... Don let me know if you need assistance.

guides/v2.2/contributor-guide/contributing_dod.md
nathanjosiah: > Even if tests are unreliable, it should be still more efficient to run them during regression than doing manual testing.

More efficient in what way though? Not seeing a real failed test right away and waiting for code to go through a full PR process then waiting for a pagerduty team to pick it up and investigate the issue, understand what happened, try and understand who broke it, then either fix the test or try to fix the behavior, seems like a way less efficient process then just having a reliable test suite.


> To optimize every day work, functional tests can be divided into groups (which I thought we already do) and only selected scenarios run on every PR.

For functional tests, yes this would be desirable and there is a separate initiative that hasn't been started to selectively run functional tests based on what was actually changed.

dobooth:
running tests

PR #5551, opened 2019-10-02
Clarified password management explanation

### LINKED ISSUE #5472: Clarify encryption library for PHP 7.2
The page states "added support for Argon2ID13, which is used by default in PHP 7.2 or if you have installed the Libsodium extension for PHP."

PHP 7.2 uses bcrypt by default, and I'm pretty sure even when compiled with Argon2 and Sodium, only Argon2i is supported by PHP 7.2. Support was added in 7.3 for Argon2id.

[https://www.php.net/manual/en/function.password-hash.php#refsect1-function.password-hash-changelog](https://www.php.net/manual/en/function.password-hash.php#refsect1-function.password-hash-changelog)

Edit:
It appears, PHP 7.2 can use Argon2ID, but it must be compiled with a more recent version of Sodium than I had tested with.

dobooth:

This pull request (PR) clarifies the password management section.

Fixes #5472. 

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/architecture/security_intro.html


guides/v2.3/architecture/security_intro.md
```diff
@@ -8,7 +8,7 @@ menu_order:
 
 ## Enhanced password management
 
-Magento has strengthened the hashing algorithms (SHA-256) used in password management. In addition, we added support for Argon2ID13, which is used by default in PHP 7.2 or if you have installed the Libsodium extension for PHP.
+Magento has strengthened the hashing algorithms (SHA-256) used in password management. In addition, we added support for Argon2ID13, which is available by default in PHP 7.2.
```
hostep: Support for Argon2ID13 got added to libsodium version 1.0.13 (source: https://download.libsodium.org/doc/password_hashing/the_argon2i_function#key-derivation)

So if PHP's sodium extension is using a lower version of libsodium, Argon2ID13 won't be available and Magento won't use it.

Therefore I would change the phrase to something like this:
> Magento has strengthened the hashing algorithms (SHA-256) used in password management. In addition, we added support for Argon2ID13, which is available in PHP through its sodium extension, but libsodium itself needs to be at least version 1.0.13 for it to become available.

I wouldn't mention the PHP version, since that is irrelevant, the functionality is brought in through the sodium extension, I've seen people on [stackoverflow](https://stackoverflow.com/a/57681879/145829) claiming they got the sodium extension working with PHP 7.1, so ...

@Tsquare17: is it clearer like this?

dobooth: I like that clarification. Thanks for the suggestion.

guides/v2.3/architecture/security_intro.md
```diff
@@ -8,7 +8,7 @@ menu_order:
 
 ## Enhanced password management
 
-Magento has strengthened the hashing algorithms (SHA-256) used in password management. In addition, we added support for Argon2ID13, which is used by default in PHP 7.2 or if you have installed the Libsodium extension for PHP.
+Magento has strengthened the hashing algorithms (SHA-256) used in password management. In addition, we added support for Argon2ID13, which is available in PHP through its sodium extension, but libsodium itself needs to be at least version 1.0.13 for it to become available.
```
hguthrie: itself? Is it aware?

guides/v2.3/architecture/security_intro.md
```diff
@@ -8,7 +8,7 @@ menu_order:
 
 ## Enhanced password management
 
-Magento has strengthened the hashing algorithms (SHA-256) used in password management. In addition, we added support for Argon2ID13, which is used by default in PHP 7.2 or if you have installed the Libsodium extension for PHP.
+Magento has strengthened the hashing algorithms (SHA-256) used in password management. Magento now supports Argon2ID13 through the PHP sodium extension. Sodium version 1.0.13 or higher is required for this functionality.
```
hostep: Hi @hguthrie & @dobooth 

Be aware that there is a difference between the php sodium extension and the libsodium library.
The php sodium extension has a dependency on libsodium, so they both need to be installed.

In this case, the libsodium library needs to be at least version 1.0.13, the php sodium extension version doesn't really matter and probably just follows the php version numbering.

Hope this helps :)

hguthrie: Yes, @hostep ! Thank you. 

```suggestion
Magento has strengthened the hashing algorithms (SHA-256) used in password management. Magento now supports Argon2ID13 through the PHP sodium extension, which requires the libsodium library version 1.0.13 or higher.
```

hostep: Looks good, thanks!

Issue #5468, opened 2019-09-24
Prerequisite task in "Create a configurable product using bulk APIs" tutorial doesn't work


webmaster-sfs:
When I get to the last step (Use the bin/magento queue:consumers:start async.operations.all command to enable bulk endpoint processing.), I get an error:

$ bin/magento queue:consumers:start async.operations.all
In ConnectionTypeResolver.php line 43:

I'm on Magento CE 2.3.1 

Affected topic: /guides/v2.3/rest/tutorials/bulk-configurable-product/config-product-intro.html

#5867  fixes this issue

keharper:
@nuzil Is there any system configuration that needs to occur befoe you run `bin/magento queue:consumers:start async.operations.all`?

brunoroeder:
@keharper and @webmaster-sfs You should configure the cron_consumers_runner before: https://devdocs.magento.com/guides/v2.3/config-guide/mq/manage-message-queues.html

meker12:
This was fixed in #5867 fixes this issue.  Must ensure that RabbitMQ is configured before running the command provided.

PR #5569, opened 2019-10-04
Add manage message queues instructions

brunoroeder:

Inserted infos about Manage message queues. This is a suggestion base on  #5468 'issue'.

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.3/rest/bulk-endpoints.html

### DIFF
--- a/guides/v2.2/rest/bulk-endpoints.md
+++ b/guides/v2.2/rest/bulk-endpoints.md
@@ -11,7 +11,7 @@ functional_areas:
 Bulk API endpoints differ from other REST endpoints in that they combine multiple calls of the same type into an array and execute them as a single request. The endpoint handler splits the array into individual entities and writes them as separate messages to the message queue.
 
 {:.bs-callout .bs-callout-tip}
-Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing, after configure the message queues'. For more details see: [
+Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing, after configure the message queues. For more details see: [
 Manage message queues
 ]({{ page.baseurl }}config-guide/mq/manage-message-queues.html).
 
--- a/guides/v2.2/rest/bulk-endpoints.md
+++ b/guides/v2.2/rest/bulk-endpoints.md
@@ -11,9 +11,7 @@ functional_areas:
 Bulk API endpoints differ from other REST endpoints in that they combine multiple calls of the same type into an array and execute them as a single request. The endpoint handler splits the array into individual entities and writes them as separate messages to the message queue.
 
 {:.bs-callout .bs-callout-tip}
-Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing, after configure the message queues. For more details see: [
-Manage message queues
-]({{ page.baseurl }}config-guide/mq/manage-message-queues.html).
+Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing, after configure the message queues. For more details see: [Manage message queues]({{ page.baseurl }}config-guide/mq/manage-message-queues.html).
 
 ### Routes
 
--- a/guides/v2.2/rest/bulk-endpoints.md
+++ b/guides/v2.2/rest/bulk-endpoints.md
@@ -11,7 +11,9 @@ functional_areas:
 Bulk API endpoints differ from other REST endpoints in that they combine multiple calls of the same type into an array and execute them as a single request. The endpoint handler splits the array into individual entities and writes them as separate messages to the message queue.
 
 {:.bs-callout .bs-callout-tip}
-Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing.
+Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing, after configuring the message queues'. For more details see: [
+Manage message queues
+]({{ page.baseurl }}config-guide/mq/manage-message-queues.html).
 
 ### Routes
 
--- a/guides/v2.2/rest/bulk-endpoints.md
+++ b/guides/v2.2/rest/bulk-endpoints.md
@@ -11,7 +11,7 @@ functional_areas:
 Bulk API endpoints differ from other REST endpoints in that they combine multiple calls of the same type into an array and execute them as a single request. The endpoint handler splits the array into individual entities and writes them as separate messages to the message queue.
 
 {:.bs-callout .bs-callout-tip}
-Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing, after configuring the message queues'. For more details see: [
+Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing, after configure the message queues'. For more details see: [
 Manage message queues
 ]({{ page.baseurl }}config-guide/mq/manage-message-queues.html).
 
--- a/guides/v2.3/rest/bulk-endpoints.md
+++ b/guides/v2.3/rest/bulk-endpoints.md
@@ -9,10 +9,14 @@ functional_areas:
 
 Bulk API endpoints differ from other REST endpoints in that they combine multiple calls of the same type into an array and execute them as a single request. The endpoint handler splits the array into individual entities and writes them as separate messages to the message queue.
 
+{: .bs-callout .bs-callout-warning}
+By default, Magento uses RabbitMQ (AMQP) for processing message queues. Before using Bulk operations, you should see [RabbitMQ]({{ page.baseurl }}/install-gde/prereq/install-rabbitmq.html).
+
+{:.bs-callout .bs-callout-info}
+Alternatively you can use CRON for processing message queues without RabbitMQ installed, see [Manage message queues]({{ page.baseurl }}/config-guide/mq/manage-message-queues.html).
+
 {:.bs-callout .bs-callout-tip}
-Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing, after configure the message queues. For more details see: [
-Manage message queues
-]({{ page.baseurl }}config-guide/mq/manage-message-queues.html).
+Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing.
 
 ### Routes
 
--- a/guides/v2.3/rest/bulk-endpoints.md
+++ b/guides/v2.3/rest/bulk-endpoints.md
@@ -10,7 +10,9 @@ functional_areas:
 Bulk API endpoints differ from other REST endpoints in that they combine multiple calls of the same type into an array and execute them as a single request. The endpoint handler splits the array into individual entities and writes them as separate messages to the message queue.
 
 {:.bs-callout .bs-callout-tip}
-Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing.
+Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing, after configuring the message queues'. For more details see: [
+Manage message queues
+]({{ page.baseurl }}config-guide/mq/manage-message-queues.html).
 
 ### Routes
 
--- a/guides/v2.3/rest/bulk-endpoints.md
+++ b/guides/v2.3/rest/bulk-endpoints.md
@@ -10,7 +10,7 @@ functional_areas:
 Bulk API endpoints differ from other REST endpoints in that they combine multiple calls of the same type into an array and execute them as a single request. The endpoint handler splits the array into individual entities and writes them as separate messages to the message queue.
 
 {:.bs-callout .bs-callout-tip}
-Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing, after configure the message queues'. For more details see: [
+Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing, after configure the message queues. For more details see: [
 Manage message queues
 ]({{ page.baseurl }}config-guide/mq/manage-message-queues.html).
 
--- a/guides/v2.3/rest/bulk-endpoints.md
+++ b/guides/v2.3/rest/bulk-endpoints.md
@@ -10,7 +10,7 @@ functional_areas:
 Bulk API endpoints differ from other REST endpoints in that they combine multiple calls of the same type into an array and execute them as a single request. The endpoint handler splits the array into individual entities and writes them as separate messages to the message queue.
 
 {:.bs-callout .bs-callout-tip}
-Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing, after configuring the message queues'. For more details see: [
+Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing, after configure the message queues'. For more details see: [
 Manage message queues
 ]({{ page.baseurl }}config-guide/mq/manage-message-queues.html).
 


devops-devdocs:

guides/v2.2/rest/bulk-endpoints.md
```diff
@@ -11,7 +11,9 @@ functional_areas:
 Bulk API endpoints differ from other REST endpoints in that they combine multiple calls of the same type into an array and execute them as a single request. The endpoint handler splits the array into individual entities and writes them as separate messages to the message queue.
 
 {:.bs-callout .bs-callout-tip}
-Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing.
+Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing, after configure the message queues. For more details see: [
```
rogyar: Basically, the [mentioned](https://devdocs.magento.com/guides/v2.3/config-guide/mq/manage-message-queues.html) article describes how to configure message queues using CRON consumers runner which is not default MQ processor in Magento. By default, Magento uses RabbitMQ (AMQP) for processing message queues and its configuration process is slightly different. So, I would mention the [following](https://devdocs.magento.com/guides/v2.3/install-gde/prereq/install-rabbitmq.html) article first of all and then the [alternative]((https://devdocs.magento.com/guides/v2.3/config-guide/mq/manage-message-queues.html)) way of configuring the message queues. 

Thank you!

guides/v2.3/rest/bulk-endpoints.md
```diff
@@ -9,6 +9,12 @@ functional_areas:
 
 Bulk API endpoints differ from other REST endpoints in that they combine multiple calls of the same type into an array and execute them as a single request. The endpoint handler splits the array into individual entities and writes them as separate messages to the message queue.
 
+{: .bs-callout .bs-callout-warning}
+By default, Magento uses RabbitMQ (AMQP) for processing message queues. Before using Bulk operations, you should see [RabbitMQ]({{ page.baseurl }}/install-gde/prereq/install-rabbitmq.html).
+
+{:.bs-callout .bs-callout-info}
+Alternatively you can use CRON for processing message queues without RabbitMQ installed, see [Manage message queues]({{ page.baseurl }}/config-guide/mq/manage-message-queues.html).
```
hostep: Correct me if I'm wrong, but I'm almost 100% convinced that cron should always be used, even if you use RabbitMQ, otherwise there won't be any consumers running and messages will only end up in the queue but will not get processed.
There seems to be a lot of misinformation out there around this fact.

The exception is if you use something like [supervisor to manage consumers](https://devdocs.magento.com/guides/v2.3/config-guide/mq/manage-message-queues.html#process-management), but that has nothing to do with RabbitMQ vs MysqlMQ...

guides/v2.2/rest/bulk-endpoints.md
hostep: I don't agree with this comment, Magento both uses MysqlMQ and RabbitMQ out of the box, it depends on which queue and how it's defined in the xml configuration files for that particular queue.
Cron's don't really have anything to do with this, they are used to spawn message queue consumers, doesn't matter for which broker (RabbitMQ or MysqlMQ).

As for bulk operations, I think Magento defaults to the RabbitMQ at the moment, so setting up RabbitMQ is probably a requirement before bulk operations can work.

guides/v2.3/rest/bulk-endpoints.md
brunoroeder: Hi @hostep base on @rogyar [previous](https://github.com/magento/devdocs/pull/5569#discussion_r331595732) explanation and first paragraph of [
Manage message queues
](https://devdocs.magento.com/guides/v2.3/config-guide/mq/manage-message-queues.html#process-management) - "If you don’t want to implement the RabbitMQ solution, you can manage message queues with cron jobs (or an external process manager) and the command line to ensure that consumers are retrieving messages." - Is not necessary have the job 'consumers_runner' enabled if you use RabbitMQ to manage queues.

hostep: Well, that's what I mean by
> There seems to be a lot of misinformation out there around this fact.

😉 

Anyway, just to make sure I know what I'm talking about, I just tested this again to be very sure, here are some graphs from the RabbitMQ management interface:
Screenshot 2019-10-07 at 22 11 30

Between 22:06 and 22:10, I send in 3 REST calls to the bulk endpoints. No cronjobs or message queue consumers were running at that time. Notice that messages are added to the queue but stay in there and aren't getting processed.

At around 22:11, I executed a `bin/magento cron:run` which spawned some message queue consumer processes in the background. Notice that suddenly all messages are getting processed and the queue is empty again.

So using RabbitMQ by itself is not enough, you need consumers running within Magento so they can process the pending messages in the queue. Those consumers either get spawned using Magento's cronjob, you can start them manually using `bin/magento queue:consumers:start ...`, or you can use something like supervisor to manage them.

Is this enough information which you guys can use to fix all incorrect information on devdocs around this? Thanks! 🙂

brunoroeder: Thanks @hostep. I'll investigate it. Probably the misinformation is caused by de description of the [Manage message queues](https://devdocs.magento.com/guides/v2.3/config-guide/mq/manage-message-queues.html#process-management) that explains:

> If you don’t want to implement the RabbitMQ solution, you can manage message queues with cron jobs ...

But at [RabbitMQ](https://devdocs.magento.com/guides/v2.3/install-gde/prereq/install-rabbitmq.html) explains that Crons are necessary to run consumers:

> After you have connected Magento and RabbitMQ, you must start the message queue consumers. See Configure message queues for details.

Probably the mistake is in the [Manage message queues](https://devdocs.magento.com/guides/v2.3/config-guide/mq/manage-message-queues.html#process-management) description that cause this confusion :(

BaDos: Hi there!

Consumers can use RabbitMQ or the DB for messages. It depends on the implementation of the consumer.
But to run consumers you need to run the command `bin/magento queue:consumers:start ...`
This command can be run by the cron job or some worker (for example, you can use workers on Magento Cloud Pro instances)

dobooth:
Thanks for the work here. Is everyone in agreement with this now? Ready to merge?

keharper:
@meker12 Are you working on this too?

guides/v2.3/rest/bulk-endpoints.md
meker12: No.  I still need to review the PR, and there is also a confusion between running the consumer to process the messages, which is covered by the start consumers  and configuring the message queues.

@brunoroeder Check this topic:  https://devdocs.magento.com/guides/v2.3/extension-dev-guide/message-queues/implement-bulk.html

I also think that this link would be a better one for linking to how to start consumers:  https://devdocs.magento.com/guides/v2.3/config-guide/cli/config-cli-subcommands-queue.html

guides/v2.2/rest/bulk-endpoints.md
```diff
@@ -11,7 +11,7 @@ functional_areas:
 Bulk API endpoints differ from other REST endpoints in that they combine multiple calls of the same type into an array and execute them as a single request. The endpoint handler splits the array into individual entities and writes them as separate messages to the message queue.
 
 {:.bs-callout .bs-callout-tip}
-Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing.
+Use the `bin/magento queue:consumers:start async.operations.all` command to enable bulk endpoint processing, after configure the message queues. For more details see: [Manage message queues]({{ page.baseurl }}config-guide/mq/manage-message-queues.html).
```
meker12: I don't think that the manage message queues is the best place to point.

Elsewhere in the REST API Guide:
https://devdocs.magento.com/guides/v2.3/rest/asynchronous-web-endpoints.html
```Use the bin/magento queue:consumers:start async.operations.all command to start the consumer that handles asynchronous and bulk API messages.```

Perhaps it's enough to include this note in the bulk-endpoints topic with an additional link to the [Start message queue consumers]https://devdocs.magento.com/guides/v2.3/config-guide/cli/config-cli-subcommands-queue.html) reference topic?

The Extensions Guide provides the details for implementing the [Bulk Operations](https://devdocs.magento.com/guides/v2.3/extension-dev-guide/message-queues/implement-bulk.html).

There is also a tutorial about bulk endpoints: https://devdocs.magento.com/guides/v2.3/rest/tutorials/bulk-configurable-product/config-product-intro.html

It seems like they can try the command, if it fails, then they could list the consumers.  If the `async.operations.all` consumer isn't available, then you would have to figure out how to enable it. See [Bulk operations] (https://devdocs.magento.com/guides/v2.3/extension-dev-guide/message-queues/bulk-operations.html) in the Extension Guide.

File: guides/v2.3/rest/bulk-endpoints.md [hunk 4]
```diff
@@ -9,6 +9,12 @@ functional_areas:
 
 Bulk API endpoints differ from other REST endpoints in that they combine multiple calls of the same type into an array and execute them as a single request. The endpoint handler splits the array into individual entities and writes them as separate messages to the message queue.
 
+{: .bs-callout .bs-callout-warning}
```
meker12: I don't think this warning is needed.  If the command doesn't work, it's likely because Bulk Operations haven't been configured, not because of RabbitMQ or cron.

guides/v2.3/rest/bulk-endpoints.md
```diff
@@ -9,6 +9,12 @@ functional_areas:
...
+By default, Magento uses RabbitMQ (AMQP) for processing message queues. Before using Bulk operations, you should see [RabbitMQ]({{ page.baseurl }}/install-gde/prereq/install-rabbitmq.html).
+
+{:.bs-callout .bs-callout-info}
```
meker12: Based on review comments, this info note isn't needed and adds to the confusion.

guides/v2.3/rest/bulk-endpoints.md
```diff
@@ -9,6 +9,12 @@ functional_areas:
 
 Bulk API endpoints differ from other REST endpoints in that they combine multiple calls of the same type into an array and execute them as a single request. The endpoint handler splits the array into individual entities and writes them as separate messages to the message queue.
 
+{: .bs-callout .bs-callout-warning}
+By default, Magento uses RabbitMQ (AMQP) for processing message queues. Before using Bulk operations, you should see [RabbitMQ]({{ page.baseurl }}/install-gde/prereq/install-rabbitmq.html).
```
meker12: I don't think this note is needed either, but I'm not sure.  I guess it depends on what error you get if you run the command to start the `async.operations.all` consumer.

guides/v2.3/rest/bulk-endpoints.md
meker12: @brunoroeder   I provided updated comments in the PR.   Short story:  Edits to note referring to start consumer content, and remove note about RabbitMQ.   Let me know if you have any questions. Thx.

guides/v2.3/rest/bulk-endpoints.md
nuzil: Sorry, but I would disagree. I would align this note in a next way

```suggestion
Bulk API require to have RabbitMQ (AMQP) configured for processing message queues. Before using Bulk operations, you should see [RabbitMQ]({{ page.baseurl }}/install-gde/prereq/install-rabbitmq.html).
```

guides/v2.3/rest/bulk-endpoints.md
nuzil: Hi @hostep , @meker12 
> Correct me if I'm wrong, but I'm almost 100% convinced that cron should always be used

Its not now. CRON really still CAN be used. But you can also disable your consumer to run by using cron. And them align your server to run consumer as a service, without basically letting Magento know that its running.

So Cronjob is really optional on this step, same us direct running on consumer process. This decision is up to developers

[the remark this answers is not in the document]
nuzil: So I would leave it

hostep: Hi @nuzil, thanks for the feedback!

I agree with your comment, sorry that was a bad wording from me.
But I don't agree with the current proposal as it seems to suggest that you need to run RabbitMQ **or** need to run cronjobs for the messages to get processed.

This is not the case. You need a place to store the messages (RabbitMQ or MysqlMQ) **and** consumer processes to do something with those messages. You need both, one of them by itself is not going to work.

For bulk api, you always need RabbitMQ (at the moment of writing) and using cronjobs as an alternative makes no sense. The cronjob which is being referred to here, is the `consumer_runner` one which spawns these consumer processes.

You are correct that cronjobs aren't necessary to spawn consumer processes, you can manually run them or have some other system manage them. But you do need consumer processes running to have the messages being processed.
So they are **optional**, but are not an **alternative** as what is being suggested.

I already tried to explain it a bit better previously a bit higher up, let me quote myself:

> So using RabbitMQ by itself is not enough, you need consumers running within Magento so they can process the pending messages in the queue. Those consumers either get spawned using Magento's cronjob, you can start them manually using bin/magento queue:consumers:start ..., or you can use something like supervisor to manage them.


I'm not seeing ghosts right? I did test this many times now already ... 🙂

nuzil: Right, I guess I skip this part of your message :-)
No you are totally correct described situation 👍 , with this amount of discussions can be wrongly to define what is the last point

meker12: > Hi @nuzil, thanks for the feedback!
> 
> I agree with your comment, sorry that was a bad wording from me.
> But I don't agree with the current proposal as it seems to suggest that you need to run RabbitMQ **or** need to run cronjobs for the messages to get processed.
> 
> This is not the case. You need a place to store the messages (RabbitMQ or MysqlMQ) **and** consumer processes to do something with those messages. You need both, one of them by itself is not going to work.
> 
> For bulk api, you always need RabbitMQ (at the moment of writing) and using cronjobs as an alternative makes no sense. The cronjob which is being referred to here, is the `consumer_runner` one which spawns these consumer processes.
> 
> You are correct that cronjobs aren't necessary to spawn consumer processes, you can manually run them or have some other system manage them. But you do need consumer processes running to have the messages being processed.
> So they are **optional**, but are not an **alternative** as what is being suggested.
> 
> I already tried to explain it a bit better previously a bit higher up, let me quote myself:
> 
> > So using RabbitMQ by itself is not enough, you need consumers running within Magento so they can process the pending messages in the queue. Those consumers either get spawned using Magento's cronjob, you can start them manually using bin/magento queue:consumers:start ..., or you can use something like supervisor to manage them.
> 
> I'm not seeing ghosts right? I did test this many times now already ... 🙂

guides/v2.3/rest/bulk-endpoints.md
nuzil: Sorry maybe I was confusing.
So we are here talking about BULK API. And BULK API do not support DB for messages. Thats just not implemented at all and there are no option currently make it works, so we should not touch DB messages when we are talking about BULK or Async

Also this link https://devdocs.magento.com/guides/v2.3/extension-dev-guide/message-queues/implement-bulk.html#configmq is not really correct in this case. Cause again if we are talking about Bulk, this part with *xml configuration is already done but implementation itself.

So to run BULK you need only:
- Magento
- RabbitMQ configured and running
- Magento connected to use rabbitMQ (env.php file)
- Running consumer: 
* via CRON
* bin/magento command executed directly
* bin/magento command added and running as a service on environment with any comfortable for integrators way)

meker12: Based on this conversation--Bulk API is configured to use RabbitMQ for message processing by default.  So customers should configure it to before using the Bulk API to process messages.

guides/v2.2/rest/bulk-endpoints.md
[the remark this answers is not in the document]
meker12: I edited the section above to include the information about RabbitMQ.  The intent is to avoid having too many separate notes in the topic.

guides/v2.3/rest/bulk-endpoints.md
meker12: Closing this conversation, which clarifies that Bulk API has its own default configuration for processing message queues, so you do not have to follow the process to configure message queues described in the Extension Guide.

guides/v2.3/rest/bulk-endpoints.md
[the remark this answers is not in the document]
meker12: @hostep  and @nuzil   Looking at an issue opened in Magento code repo: https://github.com/magento/magento2/issues/25142#issuecomment-544003657

In that issue, it looks like customer gets an error if they don't use RabbitMQ with Bulk API.   Adding the note about using alternatives for processing message seems to add confusion. Even if it is possible to use something besides RabbitMQ to process messages with the Bulk API, isn't the advantage of the Bulk API is that it is pre-configured with the functionality that you need? 

My suggestion is to delete the "Alternatively, you can use ... " note from this topic for now.  And to provide information about the custom configuration that does not use RabbitMQ as a separate topic, perhaps a KB article or blog.  That is assuming that the best practice and easiest way to process message queues using the Bulk API is to use the default configuration.

Additionally, after we close this PR, it might be helpful to update the [Bulk API tutorial](https://devdocs.magento.com/guides/v2.3/rest/tutorials/bulk-configurable-product/config-product-intro.html) to include RabbitMQ configuration in the _Before you begin_ section.

guides/v2.2/rest/bulk-endpoints.md
```diff
@@ -11,7 +11,7 @@ functional_areas:
 Bulk API endpoints differ from other REST endpoints in that they combine multiple calls of the same type into an array and execute them as a single request. The endpoint handler splits the array into individual entities and writes them as separate messages to the message queue.
 
 {:.bs-callout .bs-callout-tip}
```
meker12:
```suggestion
{:.bs-callout .bs-callout-info}
```

guides/v2.3/rest/bulk-endpoints.md
meker12: Suggest incorporating this information in the first note instead of having a second note immediately following the first one.

meker12:
This PR getting complicated, and I haven't heard back from @brunoroeder.  Created #5867 incorporating review feedback and suggested changes. Closing this one.

PR #5595, opened 2019-10-06
Added note about mcrypt and PHP < 7.2

jonashrem:
as php 7.2 depracted the mcrypt extension the recommendation here should only apply for 7.1.

While it is still possible to compile mcrypt for php 7.2 this is not recomandated.

Instead, for 7.2 magento the mcrypt_compat php extension is included which supplies the mcrypt functionality.

adding a note for php 7.2 and mcrypt

## Affected DevDocs pages

 - https://devdocs.magento.com/guides/v2.3/install-gde/system-requirements-tech.html

- https://github.com/magento/magento2/blob/2.3-develop/composer.json#L47


devops-devdocs:

PR #5598, opened 2019-10-06
 Add the recommendation for using IP address over a hostname for database config

jonashrem:
hostname can cause an additional delay on each connect depending on your dns setup as the hostname needs to be resolved on each request.

Using the IP address should be recommended because of this.

This pull request (PR) adds an recomandation for using IP adresses over hostnames 

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/install-gde/install/web/install-web_2-db.html


devops-devdocs:

Issue #5248, opened 2019-08-22
Markdown linting: Ordered list item prefix (MD029)


dshevtsov:
# Bug report

## Description

The Markdown linter reports about 3867 issues when the [rule 'MD029'](https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md029---ordered-list-item-prefix) is enabled.

### Instructions

1. Add a comment to this issue requesting the `devdocs-admins` to assign the issue to you.
1. Checkout the `MD029-integration` branch.
1. In the `_checks/styles/style-rules-prod` change `exclude_rule 'MD029'` to `rule 'MD029'`.
1. In your terminal, run `rake test:md`.
1. Review the list of errors.
1. Add a comment to this issue claiming a directory that you want to work on (should include both v2.2 and v2.3 directories when applicable).
1. Fix errors.

## Test your updates

1. Follow "Steps to reproduce" locally to get list of errors to fix.
2. Preview the HTML locally to ensure your changes did not break formatting in the resulting HTML (run `rake:preview`). For more information, refer to the [README](https://github.com/magento/devdocs#to-build-locally).

**Note**-->Do NOT use [special keywords](https://help.github.com/en/articles/closing-issues-using-keywords#about-issue-references) in your commit message or PR comments.

## Expected result

All the reported issues are fixed.
The `rake test:md` command returns `No issues found`.

## Additional information

- **Scope:** It makes sense to fix the following issues at the same time as MD029:
   - MD030: Spaces after list markers

- See the [Markdown linting issues](https://github.com/magento/devdocs/wiki/Markdown-linting-issues) wiki page for detailed guidelines.

jeff-matthews:
I'm working on the following directories:

- `guides/v2.2/rest/`
- `guides/v2.2/test/`
- `guides/v2.2/ui_comp_guide/`

jeff-matthews:
I'm working on the following directories:

- `release-notes/`
- `performance-best-practices/`

jeff-matthews:
I'm working on the following directories:

- `pattern-library/`
- `payments-integrations/`

jeff-matthews:
I'm working on the following directories:

- `marketplace/`
- `migration/`
- `mtf/`

jeff-matthews:
I'm working on the following directories:

- `inventory/`
- `javascript-dev-guide`

jeff-matthews:
I'm working on the following directories:

- `howdoi/`

[...]

PR #5668, opened 2019-10-11
Fixed ordered list prefix (MD029) - Group 3

jeff-matthews:

This pull request (PR) fixes all MD029 errors in the following guides:

- Pattern Library
- Payments Integrations

The MD)29 rule requires all ordered list prefixes to be consistent. We've chosen "one" as our prefix. The scope of #5248 also permits—but does not require—resolving errors related to the following rules:

- MD030: Spaces after list markers

This is one of three PRs related to #5248.

### DIFF
--- a/guides/v2.2/pattern-library/containers/slideouts-modals-overlays/slideouts-modals-overalys.md
+++ b/guides/v2.2/pattern-library/containers/slideouts-modals-overlays/slideouts-modals-overalys.md
@@ -32,12 +32,12 @@ _Example of panel animation_
 
 The active panel should include the following elements:
 
-* Content container (Panel)
-* A Page title
-* Close control "×" (ESC-key via keyboard)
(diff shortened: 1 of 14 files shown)


guides/v2.2/pattern-library/containers/slideouts-modals-overlays/slideouts-modals-overalys.md
```diff
@@ -133,12 +133,12 @@ _Example of a Confirmation Modal_
 
 The active modal should include the following elements:
 
-* Content container (Modal Window)
+* ccContent container (Modal Window)
```
keharper: Verify that you meant to add "c" before almost every list item and "cc" before this one.

Also, this topic has a typo "overalys", but fixing that is out of scope.

jeff-matthews: Yikes, thanks for catching that...

PR #5682, opened 2019-10-14
Fixed ordered list prefix (MD029) - Group 11

jeff-matthews:

This pull request (PR) fixes some MD029 errors in the following guides:

- Configuration

The MD029 rule requires all ordered list prefixes to be consistent. We've chosen "one" as our prefix. The scope of #5248 also permits—but does not require—resolving errors related to the following rules:

- MD030: Spaces after list markers

This is one of three PRs related to #5248.


jeff-matthews:
Good call @keharper.

PR #5781, opened 2019-10-17
[cicd] Update the test Action

dshevtsov:

This pull request changes the **test** GH Action to install only required gem instead of the entire bundle. This test does not require any other gem and makes the test to pass/fail faster.


dshevtsov:
running tests

PR #5783, opened 2019-10-17
JS variable declaration let and const over var

drpayyne:

This pull request (PR) is a **preliminary PR to discuss** about the preferred scoping of variable in Javscript in Magento. I'm ready to work on this more after discussion. ES6 recommends usage of only `let` and `const`, unless `var` is _necessary_. Looking for a discussion on this, regarding Javascript in Magento

## Affected DevDocs pages
- https://devdocs.magento.com/guides/v2.3/coding-standards/code-standard-javascript.html
- https://devdocs.magento.com/guides/v2.2/coding-standards/code-standard-javascript.html

### DIFF
--- a/guides/v2.2/coding-standards/code-standard-javascript.md
+++ b/guides/v2.2/coding-standards/code-standard-javascript.md
@@ -354,7 +354,7 @@ Remember that `let` and `const` are block scoped!
 let foo = 'bar',
     num = 1,
     arr = [1, 2, 3];
-    
+
 const identifier = 'CUSTOM_IDENTIFIER';
 ```
 
--- a/guides/v2.2/coding-standards/code-standard-javascript.md
+++ b/guides/v2.2/coding-standards/code-standard-javascript.md
@@ -345,14 +345,17 @@ Modifying other built-ins like `Function.prototype` is less dangerous but leads 
 
 ### Variable declarations
 
-Declare a variable with `var` wherever possible to avoid overwriting existing global values.
+Declare a variable with `let` as the default scope assignment for a variable. Use `const` when the variable will not be reassigned. Never use `var` unless you need to function scope a variable.
 
-Using only one var per scope promotes readability.
+{: .bs-callout-info }
+Remember that `let` and `const` are block scoped!
 
 ```javascript
-var foo = 'bar',
+let foo = 'bar',
     num = 1,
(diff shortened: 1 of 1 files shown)


drpayyne:
@jeff-matthews Looking for discussion on this topic! Thank you.

devops-devdocs:

jeff-matthews:
Thanks @drpayyne.

Needs review by architects @DrewML & @buskamuza.

rogyar:
I came to the same idea a few times however still don't bring this approach to life. In my opinion, recommending `let` might be a little bit inconsistent. We are talking about ES6 here, but the Magento JS base is about ES5 most of the time. But, I believe it's a good point of discussion

drpayyne:
Thank you, @rogyar. A general question to all, what's the reason DevDocs shouldn't say the following?
> Magento follows only ES5, but the developer **may** use ES6 scope assignments if they choose to do so, but with caution regarding IE11.

DrewML:
> ES6 recommends usage of only let and const, unless var is necessary

To the best of my knowledge there is no recommendations in the [ECMAScript 2015 specification](https://www.ecma-international.org/ecma-262/6.0/) for this.

> IE11 doesn't support `let` entirely. It has some issues while using in `for` loops. Use with caution.

I wouldn't be super comfortable with this unfortunately. That's a _major_ footgun, and unless we're transpiling to get identical semantics between browsers. Most folks don't typically do a full regression test in IE11.

It's worth noting that, although we enforce our coding standards in core, folks are free to use any features of the web platform they want. As we support IE11 and don't have a transpilation process in place, I wouldn't be comfortable advising people to switch over to `let`/`const`.

This is solved for the long-term in `pwa-studio`, which has Babel in place for compilation to ES5.

DrewML:
> A general question to all, what's the reason DevDocs shouldn't say the following?

It's a bit of a slippery slope unfortunately. There are 100s of web platform features we probably could use in core but don't. Unless the note will really improve the quality of stores, it'll end up leading to more confusion for folks trying to ship.

Those that are familiar with ES2015 will already be aware of `let`/`const`.

drpayyne:
@DrewML Thanks a lot for your views and opinions regarding this! Really appreciate it. I'm much more clear on this now.

PR #5824, opened 2019-10-23
Fixed the wrong path to cron.php file

serhiyzhovnir:

This pull request (PR) is to fix the path to the `cron.php` file.
The real path to this file is `pub/cron.php`.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/architecture/archi_perspectives/components/modules/mod_and_areas.html
- https://devdocs.magento.com/guides/v2.2/architecture/archi_perspectives/components/modules/mod_and_areas.html

-  https://github.com/magento/magento2/blob/2.2/pub/cron.php
- https://github.com/magento/magento2/blob/2.3/pub/cron.php


devops-devdocs:

PR #5855, opened 2019-10-27
Update events-and-observers.md

rodrigowebjump:

Add explicit information about observer node merging making understand easier for new Magento developers

## Affected DevDocs pages
https://devdocs.magento.com/guides/v2.3/extension-dev-guide/events-and-observers.html


devops-devdocs:

guides/v2.2/extension-dev-guide/events-and-observers.md
```diff
@@ -141,6 +141,7 @@ Below is an example of how to assign observers to watch certain events:
 In the preceding example, we assign the observer `MyObserver` to the custom event `my_module_event_before` and `AnotherObserver` to `my_module_event_after`.
 
 Observer names must be unique per event definition. This means that you cannot have two observers with the same name in the same event definition. In the example, both observers have the name `myObserverName`. This is acceptable because each of those observers belong to different event definitions.
+If you declare an observer with a name already in use in the same event, so Magento will merge these declarations nodes respecting the module load order as defined in app/etc/config.php file. This is very useful as you can see below.
```
rogyar: Hi @rodrigowebjump. Thank you for the clarification. I believe we might have some confusion here. In case if you declare an observer with a name that is already used in the same event, the system will **override** (not merge) an existing observer with the same name (see [schema converter](https://github.com/magento/magento2/blob/e60d3a2d8ff3964510ecaa0f7617fa9e482236c4/lib/internal/Magento/Framework/Event/Config/Converter.php#L41) for the details).
Merging means combining two values, but overriding means replacing one value by another. According to your assumption, both observers with the same name for the same event will coexist. 

Could you also clarify, please, the exact place mentioned by the following sentence:

> This is very useful as you can see below.

I'm asking because below we have an example about disabling an observer. 

Thank you!

rodrigowebjump: Hi @rogyar I am with you on this point.
I will try to rephrase to be more explicit that the merging is meaning the declarations and not that both observers will coexist.

rogyar: Awesome, thank you

PR #5879, opened 2019-10-30
Add warning for SCD symlink strategy

nuzil:

Added additional note, cause variable management is different between Cloud interface and files.

Error from cloud logs: 

- Environment configuration is not valid
     The variable STATIC_CONTENT_SYMLINK has wrong value: "true" and will be ignored, use only disabled or enabled

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.2/cloud/env/variables-deploy.html
https://devdocs.magento.com/guides/v2.3/cloud/env/variables-deploy.html


devops-devdocs:

hguthrie:
Hi, thanks for taking the time to contribute.

The environment variable reference in the Cloud guide is specific to the `.magento.env.yaml` configuration file, so references to the Cloud interface are not used in this section. The value for that environment variable should be either true or false. If you find that this in not the case, then you may need to submit a bug.

You can see the ece-tools reference for this specific variable at: https://github.com/magento/ece-tools/blob/develop/dist/.magento.env.yaml

hguthrie:
I am closing this PR because further research shows that this variable is deprecating soon:
https://github.com/magento/ece-tools/blob/develop/src/Config/Stage/DeployInterface.php

hguthrie:
@nuzil  I wanted to follow-up with you on this and let you know that I have brought your concerns up with our Cloud devs in case there is some improvement they can make.

PR #5886, opened 2019-10-30
magento/devdocs#: Add revokeCustomerToken mutation API errors

atwixfirster:

This pull request (PR) adds information about API errors of `revokeCustomerToken` mutation.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/graphql/mutations/revoke-customer-token.html

-  [The current customer isn't authorized.](https://github.com/magento/magento2/blob/2.3.3/app/code/Magento/CustomerGraphQl/Model/Resolver/RevokeCustomerToken.php#L48)

Thank you!


devops-devdocs:

guides/v2.3/graphql/mutations/revoke-customer-token.md
```diff
@@ -43,6 +43,12 @@ Attribute |  Data Type | Description
 --- | --- | ---
 `result` | Boolean! | Returns `true` if the token was successfully revoked
 
+## Errors
+
+Error | Description
+--- | ---
+`The current customer isn't authorized.` | The customer's token does not exist in the `oauth_token` table.
```
rogyar: This error appears in case if the customer is not currently logged in. If the customer token does not exist in the table, a localized exception with the message `This customer has no tokens` will be thrown. 

Please, take a look at the [following](https://github.com/magento/magento2/blob/7fc549404f7986cb85de783ea909e14a24f1d445/app/code/Magento/CustomerGraphQl/Model/Resolver/RevokeCustomerToken.php#L47) condition.

atwixfirster: @rogyar , thank you for notice but let me please clarify that technically I am correct.

If you set debug point to `line 177` in `app/code/Magento/Webapi/Model/Authorization/TokenUserContext.php`

```
$token = $this->tokenFactory->create()->loadByToken($bearerToken);
```

you will detect that `app/code/Magento/Integration/Model/Oauth/Token.php` verifies token by the next SQL query:

```
SELECT `oauth_token`.* FROM `oauth_token` WHERE (`oauth_token`.`token`='p0o183shvg7a4hfygz1h5erxkya4ywdh')
```

So, if token does not exist then Magento\CustomerGraphQl\Model\Resolver\RevokeCustomerToken returns `The current customer isn't authorized.` error.

Thank you!

rogyar: In that case, we have two different errors with the same message. If you take a look at the [link](https://github.com/magento/magento2/blob/7fc549404f7986cb85de783ea909e14a24f1d445/app/code/Magento/CustomerGraphQl/Model/Resolver/RevokeCustomerToken.php#L48) I've sent above, you will see that we have "The current customer isn't authorized" localized exception thrown by the resolver in case if the customer information is not present in the context (we have a guest user, boom!)

So, we could clarify the explanation like 

"The current customer is not currently logged in. The customer token is expired or does not exist in the database" 

(I realize that expired tokens are being removed from the database, this explanation just reduces the confusion).


Thank you!

atwixfirster: **implemented**

Thanks, @rogyar

atwixfirster:
Thanks, @keharper

PR #5891, opened 2019-10-30
Auth.md

dewimagento:

Updates the auth.md file:

- Call the token a "User Session Token" throughout.
- Add callout at top to notify that the login method will change after the phase 1 trial period.
- Add section to list and explain the usage of the two base URLs.
- Move app id and secret into its own section.
- Add "Parameters" section for the GET session token endpoint.
- Converted parameters into a table.
- Fixed 'apps' path element to 'app' throughout.
- Converted response value list to table.

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.3/marketplace/eqp/auth.html

### DIFF
--- a/marketplace/eqp/v1/auth.md
+++ b/marketplace/eqp/v1/auth.md
@@ -2,51 +2,81 @@
 group: marketplace-api
 title: Authentication
 ---
+{: .bs-callout-info }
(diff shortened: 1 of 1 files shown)


devops-devdocs:

PR #5896, opened 2019-10-31
Update files.md

dewimagento:

(Handy and mostly unrelated tip:
I find the following bookmarklet handy for very wide code-reviews: 
```
javascript:document.body.classList.toggle('full-width');
```
- putting that in your bookmarks menu or bookmark bar and clicking it while, reviewing git's diff page, makes the git diff take the whole width of the browser.]

This PR updates the files.md file:

- Add table of params for the Get files endpoint.
- Clarify the explanation of the multipart form upload.
- Remove all references to the DELETE file endpoint, no longer available to users (which I only finally got confirmation about just now, yay).

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.3/marketplace/eqp/files.html


devops-devdocs:

PR #5903, opened 2019-11-01
Fixed symbolic link creation command

drpayyne:

This pull request (PR) fixes the command to create symbolic link since using file paths in the existing way doesn't work. Absolute paths or relative paths must be used. References: [one](https://superuser.com/a/511919) and [two](https://stackoverflow.com/a/8602079/12236466)

## Affected DevDocs pages

-  https://github.com/magento/devdocs/blob/master/.github/CONTRIBUTING.md


devops-devdocs:

.github/CONTRIBUTING.md
```diff
@@ -127,14 +127,14 @@ We use symbolic links for topics that are the same across versions of Magento. A
...
+For example, if you created a new file for v.2.2: `guides/v2.2/install-gde/new-doc-topic.md` and its content is the same for v.2.3, use the command below to create the symbolic link.
 
 ```bash
-cd <DEVDOCS_REPOSITORY_ROOT_DIR>
+$ cd <DEVDOCS_REPOSITORY_ROOT_DIR>/guides/v2.3/install-gde
 ```
 
 ```bash
-ln -s guides/v2.2/install-gde/new-doc-topic.md guides/v2.3/install-gde/new-doc-topic.md
+$ ln -s ../../v2.2/install-gde/new-doc-topic.md new-doc-topic.md
```
rogyar: Hi @drpayyne. I agree with everything else except this one. Paths that include links like `../../` almost in all cases look confusing from the readability perspective. The existing version is more readable since it shows the exact source and destination of the symlink without breaking the eyes.

drpayyne: @rogyar But the method given in the docs wasn't working for me and upon searching I came across the two references linked in the PR description. And pretty much every symlink in the repository is linked relatively. Hence, my suggestion. Thank you.

rogyar: I mean, from the technical point of view the following commands

```bash
cd <DEVDOCS_REPOSITORY_ROOT_DIR> ; ln -s guides/v2.2/install-gde/new-doc-topic.md guides/v2.3/install-gde/new-doc-topic.md

cd <DEVDOCS_REPOSITORY_ROOT_DIR>/guides/v2.3/install-gde ; ln -s ../../v2.2/install-gde/new-doc-topic.md new-doc-topic.md
```
are identical. However, the second command use backslashed paths that break readability. You can compare the following two simple examples:

```bash
cd [root_dir] ; cp file.php some/destination/directory
cd [root_dir]/some/destination/directory ; cp ../../file.php .
```

At first glance, it's pretty easy to guess what happens in the case of `cp file.php some/destination/directory`. But it's relatively hard to guess what is the source and the destination of `cp ../../file.php .`.

So, I would recommend using an approach that is self-explanatory instead of using workarounds in the documentation. 
Your assumption will work, for sure. But it looks much more tricky that the existing example that technically is identical. 

Sorry about the long read, just want to explain what I mean. It's not that easy in cases like this one :)

drpayyne: @rogyar Thank you for your explanation. Helped me understand this well. Although this command `cd <DEVDOCS_REPOSITORY_ROOT_DIR> ; ln -s guides/v2.2/install-gde/new-doc-topic.md guides/v2.3/install-gde/new-doc-topic.md` does not work for me. :thinking: Only making the paths relative, works for me. Is that expected?

rogyar: I've just tried to find out why it might not work and realized that without passing an absolute path to the <source> argument that might not work. 

In that case, your example will do the trick.

drpayyne: @rogyar Oh, alright. Thanks for the info!

dobooth:
Hey @drpayyne: Note that when using the ```bash code block, our CSS will add the $ but won't select it when copying the line.

dobooth:
running tests

drpayyne:
@dobooth Thanks for the piece of info! I didn't know that. :sweat_smile:

PR #5909, opened 2019-11-01
Users.md

dewimagento:

This PR updates the users.md file:

- Remove all references to the DELETE file endpoint, no longer available to users (which I only finally got confirmation about just now, yay).

ToDo: Still more work needed on the reports section there, but doesn't need to be done for day 1.

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.3/marketplace/eqp/users.html

### DIFF
--- a/marketplace/eqp/v1/users.md
+++ b/marketplace/eqp/v1/users.md
@@ -66,7 +66,7 @@ curl -X GET \
         "payment_type": 1,
         "payment_info": {"paypal_email" : "cnorris@example.com"},
         "taxpayer_type": 2,
-        "tax_review_status": 3;
+        "tax_review_status": 3,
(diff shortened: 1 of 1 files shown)


devops-devdocs:

marketplace/eqp/v1/users.md
```diff
@@ -38,17 +40,17 @@ curl -X GET \
 ```json
 {
         "mage_id": "MAG123456789",
-        "first_name": “Chuck”,
-        "last_name": “Norris”,
+        "first_name": "Chuck",
+        "last_name": "Norris",
```
keharper: Let's change the example so that we're not using a celebrity name. Also, the value of the `email` attribute is missing the first quotation mark.

PR #5985, opened 2019-11-12
[site] Fix bug in rake build:stage

dshevtsov:

This pull request fixes the `rake build:stage` implementation by replacing a string bash syntax with ruby syntax.

Green build on staging: 1057


dshevtsov:
running tests

PR #6008, opened 2019-11-15
[site] Add 'rake index' and 'rake multirepo:reinit'

dshevtsov:

This pull request:

1. Adds `rake index` that runs Algolia indexing (internal use only)
1. Adds `rake multirepo:reindex` that removes existing sources listed in `Docfile.yml` and runs `rake init` to recreate the file tree
1. Adds exception handling to the `rake multirepo:` tasks when a subrepository failed to checkout


dshevtsov:
> why it's set up this way

It's done on purpose to decrease a risk of mixing up `rake init` with `rake reinit`. The command is destructive and can make undesirable changes when run by mistake by a person who contributes to a subrepository.

jeff-matthews:
Thanks for explaining!

dshevtsov:
running tests

PR #6133, opened 2019-12-05
Corrected the Docker compose override file name

shiftedreality:

Fixes the default Docker override filename according to https://docs.docker.com/compose/extends/#multiple-compose-files


shiftedreality:
@drpayyne @hguthrie, please review

devops-devdocs:

PR #6143, opened 2019-12-06
Stop indexing MTF docs

jeff-matthews:

This pull request (PR) updates the `robots.txt` file to prevent search engines from indexing the MTF docs.

The Magento Test Framework (MTF) has been deprecated in favor of the Magento Functional Test Framework (MFTF). Although we will still publish MTF docs, we don't want users finding them via Google (or any other search engine) search because we would rather that they use MFTF. Excluding MTF docs from Google search will help drive users to MFTF.

## Affected DevDocs pages

All MTF docs


jeff-matthews:
running tests

PR #6207, opened 2019-12-14
Update documentation for the Text UI component

serhiyzhovnir:

This pull request (PR) updates the documentation for the Text UI component.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/ui_comp_guide/components/ui-text.html
-  https://devdocs.magento.com/guides/v2.2/ui_comp_guide/components/ui-text.html

-  https://github.com/magento/magento2/blob/2.3/app/code/Magento/Ui/view/base/web/js/form/element/text.js
-  https://github.com/magento/magento2/blob/2.2/app/code/Magento/Ui/view/base/web/js/form/element/text.js
-  https://github.com/magento/magento2/blob/2.3/app/code/Magento/Ui/view/base/web/templates/form/element/text.html
-  https://github.com/magento/magento2/blob/2.2/app/code/Magento/Ui/view/base/web/templates/form/element/text.html


devops-devdocs:

keharper:
@serhiyzhovnir I want to verify that the addition of `elementTmpl` is correct. It doesn't appear in either of the codebase files that you link to.

serhiyzhovnir:
Hi @keharper 
Thanks for asking.
Yeah, you are right, the `elementTmpl ` option is missed in the https://github.com/magento/magento2/blob/2.2/app/code/Magento/Ui/view/base/web/js/form/element/text.js.
But if we want to show the field in the UI form, we need to use the *ui/form/field* and configure the template for the element via `elementTmpl ` option. 
The code reference where the field template renders the element template: https://github.com/magento/magento2/blob/2.3/app/code/Magento/Ui/view/base/web/templates/form/field.html#L21
The element template for this component is already implemented here: https://github.com/magento/magento2/blob/2.3/app/code/Magento/Ui/view/base/web/templates/form/element/text.html
In this PR I've added an example, how to configure the text component and add it to the UI form.
Let me know if you have any other questions.
Thank you!

PR #6244, opened 2019-12-18
Add support for DELETE async calls

nuzil:

DELETE support for Async/Bulk calls was added from 2.3.2, so I adding it to the list

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.3/rest/bulk-endpoints.html
https://devdocs.magento.com/guides/v2.3/rest/asynchronous-web-endpoints.html


devops-devdocs:

keharper:
@nuzil Would you add at least one example showing the usage of a DELETE call?

nuzil:
@keharper 
Good point, will do

nuzil:
@keharper 
I added only for BULK requests, cause for Async there are nothing special you have to think about

PR #6246, opened 2019-12-18
Added additional step in add custom field document

chandru-199:

This pull request (PR) : Added additional step with cli commands. And step number is missing in step1 so it has been added.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/howdoi/checkout/checkout_custom_checkbox.html


devops-devdocs:

chandru-199:
@jeff-matthews This PR also have technical changes. But technical label is not added . Am I right?

jeff-matthews:
In the spirit of the holiday season, I'll add the technical label. But it's on the cusp of what we would normally consider technical.

PR #6283, opened 2019-12-19
Cloud Suite release note updates for Cloud components package

meker12:

- Added release notes topic for magento-cloud-patches with 1.0.0 and 1.0.1 release updates
- Updated titles on release notes topics for consistency

## Affected DevDocs pages

- https://devdocs.magento.com/cloud/release-notes/mcc-release-notes.html
- https://devdocs.magento.com/cloud/release-notes/mcd-release-notes.html
- https://devdocs.magento.com/cloud/release-notes/cloud-tools-release-notes.html
- https://devdocs.magento.com/cloud/release-notes/mcp-release-notes.html

whatsnew
Added [Magento Cloud Components release notes ](https://devdocs.magento.com/cloud/release-notes/mcc-release-notes.html) topic to the _Cloud Guide_.


src/cloud/cdn/cloud-vcl-custom-snippets.md
```diff
@@ -2,8 +2,8 @@
 group: cloud-guide
 title: Custom Fastly VCL snippets
 redirect_from:
-   - /guides/v2.2/cloud/configure/cloud-vcl-custom-snippets.html
-   - /guides/v2.3/cloud/configure/cloud-vcl-custom-snippets.html
+   - /cloud/configure/cloud-vcl-custom-snippets.html
```
hguthrie: This seems strange if that file never existed at that address. You could probably remove these redirects completely, no? Would that work?

@jeff-matthews  What do you think about this? Would the global redirect cover for these older redirects?

jeff-matthews: The global AWS redirect takes you to https://devdocs.magento.com/cloud/configure/cloud-vcl-custom-snippets.html, which returns a 404. I agree it might seem strange, but stripping the version info from the path could be the solution to get these file-based cloud redirects working again.

meker12: ah--this is a test. I forgot that it was in this PR. Had a conversation with Dima. None of the redirects are working because the global redirect is applied first. Dima said that if we strip the version from the path, then the redirects should work.

PR #6248, opened 2019-12-18
Replace symlinks to guides/v2.2 with real files

dshevtsov:
## Purpose of this pull request

This pull request replaces all symlinks to files in `guides/v2.2` with real files. This is a required step in preparation of the 2.2 content transfer to a separate branch.

This PR also removes 2.3 redirects from 2.2 guides, and removes 2.2 redirects from 2.3 guides.

After this PR is merged, all new version-specific PRs should not affect the `guides/v2.2` files.

_Internal preview: 1257_

## Affected DevDocs pages

-  the content was not affected

PR #6284, opened 2019-12-20
Removed symlink info from README

jeff-matthews:

This pull request (PR) removes info about symlinks for doc versioning from the README (per #6248). It also notifies contributors that we accept updates to 2.3 docs only.


jeff-matthews:
running tests

Issue #6311, opened 2019-12-30
Wrong Inventory (MSI) repo url in the links


milindsingh:
# General issue
https://devdocs.magento.com/guides/v2.3/inventory/inventory-api-reference.html

## Description:
All urls are 404 in the inventory API reference page.

## Possible solutions:
Add a new variable and update the blob urls.

PR #6312, opened 2019-12-30
Inventory (msi) blob url variable added and all urls updated

milindsingh:

This pull request (PR) is for issue fix https://github.com/magento/devdocs/issues/6311

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/inventory/inventory-api-reference.html


devops-devdocs:

devops-devdocs:

_config.yml
```diff
@@ -191,6 +191,9 @@ check_links: true
 # Magento 2 repository variable
 mage2bloburl: https://github.com/magento/magento2/blob
 
+# Magento 2 inventory project (a.k.a MSI) release branch url variable
+inventory_blob_url: https://github.com/magento/inventory/blob/1.1-develop
```
rogyar: Hi @milindsingh. Thank you for your collaboration. 
I'm not sure we should use a develop branch instead of a release branch as a reference. Perhaps it's better to use [1.1.3](https://github.com/magento/inventory/tree/1.1.3) brach instead? @jeff-matthews, would you help us with the correct decision here, please? Thank you!

jeff-matthews: Thanks at @rogyar. I agree that it is better to use a stable release branch instead because a develop branch will likely include unreleased (and therefore undocumented) code.

@keharper, what say you?

Also, the name of the new inventory variable uses underscores, which is inconsistent with other variable names. That's very minor and the underscores help with readability, so I'm inclined to accept the naming convention. I just wanted to note the difference.

milindsingh: Thanks, @rogyar I have updated the correct release branch. (and tested on my local, can be merged if okay)

@jeff-matthews The whole config is using the `snake case` variables which is good in term of readability except the few variables (which are inconsistent). Maybe it was something very minor and got ignored. 

Also, all the reference URLs in doc open on the same tab upon click, which while reading makes users go back again to the doc and continue reading. 
Is opening reference links on new tab have any other disadvantage?

jeff-matthews: @milindsingh, we have stopped using the attribute to open external links in a new tab because we believe that users should be able to choose whether a link opens in a new tab or in the same window. 

Functionally, the only other disadvantage I can think of is the potential security vulnerability in [reverse tabnabbing](https://www.owasp.org/index.php/Reverse_Tabnabbing). We can mitigate that using additional attributes, but it starts making the source markdown cluttered and hard to read for maintainers/contributors.

dshevtsov:
I'm not sure about the need to introduce a global variable if it is used only in one topic.

You can create a local variable in the topic:
```
{% assign  msi_blob_url = 'https://github.com/magento/inventory/blob/1.1.3' %}
```
And use it in text without `site` like:
```
{{ msi_blob_url }}
```

milindsingh:
> I'm not sure about the need to introduce a global variable if it is used only in one topic.
> 
> You can create a local variable in the topic:
> 
> ```
> {% assign  msi_blob_url = 'https://github.com/magento/inventory/blob/1.1.3' %}
> ```
> 
> And use it in text without `site` like:
> 
> ```
> {{ msi_blob_url }}
> ```

Sure I can update that.
I initially thought that MSI repo link might be needed in different topics related to catalog or shipping where inventory source might be involved.

milindsingh:
Hi @dshevtsov 
I have updated the variable from global to the topic only and tested.
It ready to merge now.

dobooth:
running tests

PR #6359, opened 2020-01-08
Prepare 2.4-develop branch for contribution 

jeff-matthews:

This pull request (PR) adds the necessary configuration and `src/guides/v2.4/` files to start contributing to 2.4.0 development.

See DOC-64 ticket for implementation details and link to staging build.

Key items to consider:

- We are not using symlinks, so ongoing changes to 2.3 files will need to be manually forward ported to corresponding files in the 2.4 directory to ensure 2.4 contains the latest downstream doc updates. Should we reconsider using symlinks until we get closer to implementing branch-based versioning for 2.3 and 2.4?

- The MRG generator plugin is hardcoded to 2.3, so this branch is not generating 2.4 MRG topics at the moment. The link checker is ignoring those files, but we'll need to include them when we refactor the plugin to ensure test coverage.


jeff-matthews:
Which option do you think is more efficient in terms of time & effort?

hguthrie:
When you say manual forward port...I can't escape thinking about network routing. Are you talking about cherry-picking changes?

dshevtsov:
@hguthrie no, cherry-picking won't work in this case.
We need to sync guides/v2.4 content with guides/v2.3 when both directories are in the same branch (2.4.0-develop).

dshevtsov:
@jeff-matthews in terms of efforts and size of the repository symlinks seem to be more efficient.
@jcalcaben what do you think?

jcalcaben:
Instead of cherry-picking why not use patches?

https://stackoverflow.com/questions/16526321/how-to-apply-a-git-patch-to-a-file-with-a-different-name-and-path

dobooth:
Is this something that can be `rake`ed? `rake patch filename.md`?

dshevtsov:
> 1. Port changes all-at-once (monthly or once before the release or before the devdocs reconfiguration) using diff-tools.

This option considered patching. There are several ways of what and how to patch in terms of use of git history or real files. We'll need more research on discovering which would work best for us.

> Is this something that can be `rake`ed? `rake patch filename.md`?

It depends on the final scenario we'll come up with.

jeff-matthews:
I expect there to be a large volume of changes on a regular basis due to community contributions. Considering the anticipated volume, is patching really something we want to fool with? It's yet another maintenance task that'll likely become the responsibility of 1 or 2 people unless it can **reliably** automated.

dshevtsov:
That's why I suggested to port changes all-at-once periodically, not for each PR or commit.

The scenario has to implement the following logic:
1. get all the changes made for some period of time in 2.3 -> 2.3-patch (result of diff between two commits [commit with last sync and latest commit] from master branch)
1. get all the changes made for some period of time in 2.4 -> 2.4-patch (result of diff between two commits [commit with last sync and latest commit] from master branch)
1. merge the patches and apply the resulting patch to the 2.4 files

jeff-matthews:
That all sounds practical and possible, but I worry about the risk of something going wrong. Especially since this is intended to be a short-term solution; not something we'll need to get really good at and do on a regular, permanent basis.

I guess it's a question of level of confidence whether patching will go smoothly. That makes me think we need to set up a test scenario before deciding.

jeff-matthews:
Since there is no urgent need to set this up right away, I'm going to close this and wait until after the 2.3.4 release (1/28), when we can implement a more efficient branching strategy that won't require manual maintenance in the same way that was proposed here.

The more efficient branching strategy will leverage existing automation to sync 2.4-develop with earlier branches and only require manual intervention in the event of a merge conflict.

PR #6367, opened 2020-01-08
Remove New Relic and Blackfire section

meker12:

Remove New Relic and Blackfire information from Cloud Guide per Product Manager request.

## Affected DevDocs pages

https://devdocs.magento.com/cloud/project/new-relic.html#new-relic-and-blackfire


meker12:
> It's not pictured and may be out of scope for this quick request, but there is also a mention of Blackfire methods on line 151. Does that need removal?

I removed that reference. Thanks for catching it.

meker12:
running tests

PR #6372, opened 2020-01-10
Added an example to "Set body attributes" section

david-ziffity:

This pull request (PR) will add an example to "Set body attributes" section.

## Affected DevDocs pages
https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/layouts/xml-manage.html

### DIFF
--- a/src/guides/v2.3/frontend-dev-guide/layouts/xml-manage.md
+++ b/src/guides/v2.3/frontend-dev-guide/layouts/xml-manage.md
@@ -253,6 +253,21 @@ To set attributes for the HTML `body` tag use the `<attribute>` instruction.
 {:.bs-callout-warning}
 It is not recommended to set the `body` id in layout files that have a wider impact (`e.g. default.xml`).
 
+
+**Example:** Add inline style to the `body` tag.
+
+```xml
+<page>
+    <body>
+        <attribute name="style" value="opacity:0;"/>
+    </body>
+</page>
+```
+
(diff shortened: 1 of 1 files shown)


devops-devdocs:

devops-devdocs:

src/guides/v2.3/frontend-dev-guide/layouts/xml-manage.md
```diff
@@ -253,6 +253,21 @@ To set attributes for the HTML `body` tag use the `<attribute>` instruction.
...
+```xml
+<page>
+    <body>
+        <attribute name="style" value="opacity:0;"/>
+    </body>
+</page>
+```
+
+{:.bs-callout-warning}
+It is not recommended to write inline styles, it will impact on page performance and css precedence.
```
rogyar: Hi @david-ziffity. Thank you for your collaboration. I'm not sure about the following assumption

```
 it will impact on page performance
```

The inline styles ideally don't affect the performance in a negative way (rather in a positive, they are faster in terms of DOM processing). It's still not recommended to use them as a general styling approach.

david-ziffity: @rogyar Thanks for the clarification. I thought the inline css wont get cached so it will take time to render on every page load. I will change the warning text as "It is not recommended to write inline styles as a general styling approach and will impact on css precedence."

PR #6380, opened 2020-01-11
remove double quote to make code perfect usable

niravkrish:
Resolve code error.

I have resolved core error earlier code was not useful to run and use directly in our project or example.I have make it usable.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/ui_comp_guide/concepts/knockout-bindings.html#outerclick

-  ...

### DIFF
--- a/src/guides/v2.3/ui_comp_guide/concepts/knockout-bindings.md
+++ b/src/guides/v2.3/ui_comp_guide/concepts/knockout-bindings.md
@@ -346,7 +346,7 @@ Callback that is invoked when user clicks outside of the element.
 
 ```html
 <div id="target" outerClick="function () {
-    console.log('Clicked outside of the target node.');
+    console.log('Clicked outside of the &quot;target&quot; node.');
     }">
 </div>
 ```
--- a/src/guides/v2.3/ui_comp_guide/concepts/knockout-bindings.md
+++ b/src/guides/v2.3/ui_comp_guide/concepts/knockout-bindings.md
@@ -346,7 +346,7 @@ Callback that is invoked when user clicks outside of the element.
 
 ```html
 <div id="target" outerClick="function () {
-    console.log('Clicked outside of the "target" node.');
+    console.log('Clicked outside of the target node.');
     }">
 </div>
 ```


devops-devdocs:

devops-devdocs:

src/guides/v2.3/ui_comp_guide/concepts/knockout-bindings.md
```diff
@@ -346,7 +346,7 @@ Callback that is invoked when user clicks outside of the element.
 
 ```html
 <div id="target" outerClick="function () {
-    console.log('Clicked outside of the "target" node.');
```
rogyar: Hi @niravkrish. I believe that the "target" with quotes is a reference to the DOM element with `id="target"`. So, by removing the quotes we might reduce the clarity of the statement inside `console.log` invocation. I propose either leave the quotes or add an additional clarification in the following manner

```javascript 
    console.log('Clicked outside of the node with id="target"');
```

Thank you.

rogyar: Hi @niravkrish. Thank you for the change. I'm not sure it will work in case of `console.log()`

image

niravkrish: - I have tested it and we can definitely get the exact result.
- you have tested it in console but it will work in html.
- you can test it in html on-click element. 
- Please see below screenshot and example

<a href='https://codepen.io/mage2nirav/pen/gObQYMy'>See the example code in codepen</a>

screenshot-codepen io-2020 01 19-17_47_51

Thank You
Nirav Patel

niravkrish:
> We can probably use back ticks here, since it is a console message: `target`.

- quote; in the html will work properly you can check the screenshot as well as demo which i have provided in above comment.

[See the example code in codepen](https://codepen.io/mage2nirav/pen/gObQYMy)

**Note:** Make sure to get double quote result please visit browser console not the codepen console.

Thank you

src/guides/v2.3/ui_comp_guide/concepts/knockout-bindings.md
rogyar: You are right. If we put this message into an HTML node it will be shown in the console as expected. Thank you for the clarification.

PR #6395, opened 2020-01-15
magento/devdocs#: Extend “Customize the 503 error page” topic name

atwixfirster:

This pull request (PR) has appeared from the question in cloud channel:

and adds a "Maintenance" world into "Customize the 503 error page" section of the [Customize error pages](https://devdocs.magento.com/cloud/cdn/cloud-fastly-custom-response.html) page.

## Affected DevDocs pages

-  https://devdocs.magento.com/cloud/cdn/cloud-fastly-custom-response.html

-  ["Set Error / Maintenance page guide" by Fastly](https://github.com/fastly/fastly-magento2/blob/master/Documentation/Guides/ERROR-MAINTENANCE-PAGE.md)

### DIFF
--- a/src/cloud/cdn/cloud-fastly-custom-response.md
+++ b/src/cloud/cdn/cloud-fastly-custom-response.md
@@ -1,6 +1,6 @@
 ---
 group: cloud-guide
-title: Customize error pages
+title: Customize error and maintenance pages
 redirect_from:
    - /cloud/configure/fastly-vcl-extend-timeout.html
 functional_areas:
--- a/src/_data/toc/cloud-guide.yml
+++ b/src/_data/toc/cloud-guide.yml
@@ -205,7 +205,7 @@ pages:
           url: /cloud/cdn/configure-fastly.html
           versionless: true
           children:
-            - label: Customize error pages
+            - label: Customize error and maintenance pages
               url: /cloud/cdn/cloud-fastly-custom-response.html
               versionless: true
 


devops-devdocs:

devops-devdocs:

src/cloud/cdn/cloud-fastly-custom-response.md
```diff
@@ -30,7 +30,7 @@ The HTML code for the custom page must meet the following requirements:
 -  Specify all CSS inline in the HTML source.
 -  Bundle images in the HTML page using base64 so that they display even if Fastly is offline. See [Data URIs on the css-tricks site](https://css-tricks.com/data-uris/).
 
-## Customize the 503 error page
```
rogyar: There's at least one link in this document that goes to `(#customize-the-503-error-page)` (line 23). We need to make sure that we don't introduce any broken links with this change. 

Thank you

atwixfirster: Golden eye, @rogyar !

**fixed** and **verified** via local preview.

Thank you!

meker12:
@atwixfirster  Problem is that the page provides instructions for two types of error pages -- the WAF error page for 403 errors, and the 503 error pages.  So changing the title doesn't work.  However, we can probably put a more specific header in the page itself so that the content would show up in the search.

atwixfirster:
> @atwixfirster Problem is that the page provides instructions for two types of error pages -- the WAF error page for 403 errors, and the 503 error pages. So changing the title doesn't work. However, we can probably put a more specific header in the page itself so that the content would show up in the search.

Hi, @meker12!

Tutorial in the `Customize the 503 error page` contains the same instruction like this one - ["Set Error / Maintenance page guide" by Fastly](https://github.com/fastly/fastly-magento2/blob/master/Documentation/Guides/ERROR-MAINTENANCE-PAGE.md).

`Customize the WAF error page` section describes how to edit WAF page only.

meker12:
I think we should leave the content of the page as is, with no changes to the headers because a 503 error page can be for a timeout or maintenance as indicated in the bulleted explanation.

A simpler solution is to change the page title to "Customize error and maintenance pages".

meker12:
> > @atwixfirster Problem is that the page provides instructions for two types of error pages -- the WAF error page for 403 errors, and the 503 error pages. So changing the title doesn't work. However, we can probably put a more specific header in the page itself so that the content would show up in the search.
> 
> Hi, @meker12!
> 
> Tutorial in the `Customize the 503 error page` contains the same instruction like this one - ["Set Error / Maintenance page guide" by Fastly](https://github.com/fastly/fastly-magento2/blob/master/Documentation/Guides/ERROR-MAINTENANCE-PAGE.md).
> 
> `Customize the WAF error page` section describes how to edit WAF page only.

See my alternative suggestion below.

atwixfirster:
>A simpler solution is to change the page title to "Customize error and maintenance pages".

@meker12 , I've implemented your proposition. Could you please review?

Many thanks!

meker12:
running tests

Issue #6325, opened 2020-01-03
Configuring the lock provider should be done with 'setup:config:set' instead of 'setup:store-config:set'


hostep:
# Bug report

## Description

In the [configure the lock provider](https://devdocs.magento.com/guides/v2.3/install-gde/install/cli/install-cli-subcommands-lock.html#instgde-cli-lockconfig) section, it says to use `setup:store-config:set`, but that is incorrect. It should be done using `setup:config:set`.

### Steps to reproduce

1. Run `bin/magento setup:store-config:set --help`
1. Run `bin/magento setup:config:set --help`
1. Verify which one of those command have the options which start with `--lock-`

## Expected result

Seeing the command `setup:config:set` being used

## Possible solutions

Changing the command `setup:store-config:set` to `setup:config:set`

karyna-tsymbal-atwix:
Indeed, looks like store-config:set is deprecated since 2.2.0, and config:set should be used.
I'm preparing PR with fix.

karyna-tsymbal-atwix:
PR https://github.com/magento/devdocs/pull/6402

hostep:
This was fixed in #6402, thanks @karyna-tsymbal-atwix!

Closing issue.

PR #6402, opened 2020-01-15
Issue #6325 Change command which has to be used to configure a lock

karyna-tsymbal-atwix:
Change command which has to be used to configure a lock provider (the command that currently specified in docs doesn't have options to configure a lock and it's deprecated).

## Affected DevDocs pages
https://devdocs.magento.com/guides/v2.3/install-gde/install/cli/install-cli-subcommands-lock.html#instgde-cli-lockconfig


devops-devdocs:

devops-devdocs:

PR #6486, opened 2020-01-28
Update composer.lock for 2.3.4

dobooth:
## Purpose of this pull request

This pull request (PR) updates our composer.lock data files for 2.3.4., including ope source, EE and Cloud.

PR #6490, opened 2020-01-28
Updated composer_lock data after 2.3.4 release

dshevtsov:

This pull request (PR) updates data that are used to generate content about Magento packages (dependencies) and PHP requirements.


jeff-matthews:
@dshevtsov, is this a duplicate of #6486?

dshevtsov:
@jeff-matthews I didn't know about that PR. The purpose is the same but I cannot verify since I didn't create it.

jeff-matthews:
Ok thanks @dshevtsov. 

@dobooth, please confirm whether we can close this as a duplicate.

dshevtsov:
I don't think that we have to close it since it shows a big number of differences.

jeff-matthews:
Good point. It would be helpful to understand why.

dshevtsov:
I checked just one file in #6486
The `content hash` does not correspond to the [one tagged by 2.3.4 for Open Source](https://github.com/magento/magento2/blob/2.3.4/composer.lock#L7)

dshevtsov:
running tests

dshevtsov:
> Good point. It would be helpful to understand why.

The updates were done before the actual release, that's why the substituted files were different.

PR #6506, opened 2020-01-30
Update the Magento CLI reference data (2.2.11)

dshevtsov:

This pull request (PR) updates data for the Magento CLI reference after the 2.2.11 release.

## Additional information

The original format of the generated data is JSON, but we had to convert it to YAML because of parsing issues caused by inappropriate JSON. JSON seems to work fine now that is why the format has been changed in this PR.

whatnsew
Updated the Magento CLI reference ([Commerce](https://devdocs.magento.com/guides/v2.2/reference/cli/magento-commerce.html) | [Open Source](https://devdocs.magento.com/guides/v2.2/reference/cli/magento.html))


dshevtsov:
running tests

PR #6531, opened 2020-02-02
Added example for preference

ajithkumar-maragathavel:

This PR will provide an example of how to use the preference node for overriding methods from the core class.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/extension-dev-guide/build/di-xml-file.html


devops-devdocs:

devops-devdocs:

src/guides/v2.3/extension-dev-guide/build/di-xml-file.md
```diff
@@ -267,6 +267,67 @@ This mapping is in `app/etc/di.xml`, so the object manager injects the `Magento\
 
 This mapping is in `app/code/Magento/Backend/etc/adminhtml/di.xml`, so the object manager injects the `Magento\Backend\Model\Url` implementation class wherever there is a request for the `Magento\Core\Model\UrlInterface` in the [admin](https://glossary.magento.com/admin) area.
 
+### Override method using Preference
+
+If you want to override a method from core file, you can utilize preference to acheive it.
```
rogyar: Please, clarify the method scopes that may be overridden 

```suggestion
If you want to override a public or protected method from core class, you can utilize `preference` to achieve it.
```

src/guides/v2.3/extension-dev-guide/build/di-xml-file.md
```diff
@@ -267,6 +267,67 @@ This mapping is in `app/etc/di.xml`, so the object manager injects the `Magento\
...
+    protected function isVisible(Order $order)
+    {
+        # Write your custom logic here.
+        return !in_array(
+            $order->getStatus(),
+            $this->_orderConfig->getInvisibleOnFrontStatuses()
+        );
+    }
+}
+```
```
rogyar: Please, add the following callout here:

```
Overriding entire methods is not recommended approach and should be avoided whenever it's possible. This approach brings potential conflicts to the system and reduces the system upgradability. Use other extensibility options like event observers and plugins instead. 
```

It's a good thing to have information about the overriding but we should not advocate it somehow. 

Thank you!

ajithkumar-maragathavel: @rogyar I will update the docs. That is a great suggestion. Thanks.

src/guides/v2.3/extension-dev-guide/build/di-xml-file.md
ajithkumar-maragathavel: @rogyar Updated the contents as per your suggestion.

src/guides/v2.3/extension-dev-guide/build/di-xml-file.md
```diff
@@ -267,6 +267,70 @@ This mapping is in `app/etc/di.xml`, so the object manager injects the `Magento\
 
 This mapping is in `app/code/Magento/Backend/etc/adminhtml/di.xml`, so the object manager injects the `Magento\Backend\Model\Url` implementation class wherever there is a request for the `Magento\Core\Model\UrlInterface` in the [admin](https://glossary.magento.com/admin) area.
 
+### Override method using Preference
+
+If you want to override a public or protected method from core class, you can utilize `preference` to achieve it.
```
rogyar:
```suggestion
If you want to override a public or protected method from a core class, you can utilize the `preference` node from `di.xml` to achieve it.
```

src/guides/v2.3/extension-dev-guide/build/di-xml-file.md
```diff
@@ -267,6 +267,70 @@ This mapping is in `app/etc/di.xml`, so the object manager injects the `Magento\
...
+Here is an example to overide a method from core file.
```
rogyar:
```suggestion
Here is an example of overriding a method from the core file.
```

src/guides/v2.3/extension-dev-guide/build/di-xml-file.md
```diff
@@ -267,6 +267,70 @@ This mapping is in `app/etc/di.xml`, so the object manager injects the `Magento\
...
+
+```xml
+<!-- app/code/<VendorName/<ModuleName>/etc/di.xml -->
```
rogyar: I would suggest using `OrangeCompany/OverrideExample` as a module/vendor name. See [other](https://devdocs.magento.com/guides/v2.3/extension-dev-guide/routing.html#before-and-after-parameters) documentation for more examples

rogyar: It applies to all your examples

src/guides/v2.3/extension-dev-guide/build/di-xml-file.md
```diff
@@ -267,6 +267,70 @@ This mapping is in `app/etc/di.xml`, so the object manager injects the `Magento\
...
+        # Write your custom logic here.
+        return !in_array(
+            $order->getStatus(),
+            $this->_orderConfig->getInvisibleOnFrontStatuses()
+        );
+    }
+}
+```
+
+`Note:`
```
rogyar: Please use `{:.bs-callout-info}` directive for the callouts. You can see an example of a callout in the [following](https://devdocs.magento.com/guides/v2.3/get-started/authentication/gs-authentication-session.html) topic (the blue section with the "i" icon)

src/guides/v2.3/extension-dev-guide/build/di-xml-file.md
ajithkumar-maragathavel: @rogyar Updated the content as per your suggestion. Thanks for the review.

src/guides/v2.3/extension-dev-guide/build/di-xml-file.md
ajithkumar-maragathavel: @rogyar Updated the content and commit the changes. Thanks for the suggestion.

src/guides/v2.3/extension-dev-guide/build/di-xml-file.md
ajithkumar-maragathavel: @rogyar Thanks for providing an example link. Hereafter will follow the same. Content Updated.

src/guides/v2.3/extension-dev-guide/build/di-xml-file.md
ajithkumar-maragathavel: @rogyar Fixed this one too.

dobooth:
Hi @ajithkumar-maragathavel and @rogyar. Thanks for the work on this. We are a bit confused here why we are telling them how to do this and then saying that it is not the recommended way. If it is not recommended, then perhaps it should not be in Devdocs. Is there a good reason?

ajithkumar-maragathavel:
> Hi @ajithkumar-maragathavel and @rogyar. Thanks for the work on this. We are a bit confused here why we are telling them how to do this and then saying that it is not the recommended way. If it is not recommended, then perhaps it should not be in Devdocs. Is there a good reason?

@dobooth  We know It's not a recommended way of overriding methods. But when we are doing customization on Magento for many projects, we end up like there is no other way than adding **preference**  for the particular class to achieve our requirements. 
That's why I would like to have an example of this topic. I believe this example will help the developer's on how to use preference for method overriding.

I appreciate your thoughts on this. @dobooth 

CC: @rogyar

rogyar:
Hi @dobooth. Thank you for asking.

This part of extensibility is like a "last chance" to extend some functionality that works when no other recommended extensibility method applies. Overriding using preferences is a part of Magento extensibility system, that's why I think it's a good idea to document it but leave the note that it's not recommended approach until you have nothing else to do for the extending the functionality.

dobooth:
running tests

PR #6542, opened 2020-02-04
Note added for the object property convertors behaviour

milindsingh:

This pull request (PR) is to add a note for Object methods  to property convertors behaveiour.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/extension-dev-guide/service-contracts/design-patterns.html

-  https://github.com/magento/magento2/blob/2.4-develop/lib/internal/Magento/Framework/Api/SimpleDataObjectConverter.php

## Issues
https://github.com/magento/magento2/issues/25490

### DIFF
--- a/src/guides/v2.3/extension-dev-guide/service-contracts/design-patterns.md
+++ b/src/guides/v2.3/extension-dev-guide/service-contracts/design-patterns.md
@@ -10,7 +10,7 @@ In the programming community, a _design pattern_ is a recommended way of writing
 
 Design patterns for service contracts tell you which types of interfaces to define, and how and where to define and implement those interfaces.
 
- {:.bs-callout-info}
+{:.bs-callout-info}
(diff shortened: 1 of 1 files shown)


devops-devdocs:

devops-devdocs:

rogyar:
I can confirm that the [corresponding](https://github.com/magento/magento2/blob/40d9c7e452eaae742f35ea587bbc6f8a2d3fc8d4/lib/internal/Magento/Framework/Api/SimpleDataObjectConverter.php#L175) method of the `SimpleDataObjectConverter` follows the strict conversion. So, the underscore will be added only in case of transition from lower case to upper case in a string. This rule does not apply if the transition between alphabet character and a digit happens (i.e. thisIsNumber1 -> this_is_number1)

PR #6569, opened 2020-02-05
Update release compatibility table

jeff-matthews:

This pull request (PR) improves the version compatibility content:

- Added 2.3.4 Commerce
- Added several Page Builder versions
- Fixed matrix to reflect correct compatibility for Inventory and Page Builder

## Affected DevDocs pages

-  https://devdocs.magento.com/release/

## Additional info

To help future contributors make accurate updates, consider the following when making updates:

- There is a 1:1 relationship between versions of Magento and Page Builder. Later versions of Page Builder are **not supported** on earlier versions of Magento. The same applies to VBEs.
- Later versions of Inventory are **compatible** with earlier versions of Magento (and vice versa).


src/_data/mde.yml
```diff
@@ -38,83 +42,122 @@ extensions:
...
           2.3.1: general availability
           2.3.2: general availability
           2.3.3: general availability
+          2.3.4: general availability
   -
     name: Magento Inventory
     versions:
       -
-        name: 1.6.1
+        name: 1.1.4
```
dobooth: This is correct?

jeff-matthews: Yes, according to the canonical source for all VBEs and MDEs on the internal wiki.

jeff-matthews: Confirmed that 1.6.1 doesn't exist. Also confirmed that later versions of Inventory usually work fine with earlier Magento versions and are considered "compatible" but not supported. I'll need to push some additional changes.

jeff-matthews:
Ok, i'm done pushing changes until the next review.

jeff-matthews:
running tests

PR #6577, opened 2020-02-06
Adding cli commands in custom payment method page

VinothKumar361:

Adding cli commands in custom payment method page

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/howdoi/checkout/checkout_payment.html

### DIFF
--- a/src/guides/v2.3/howdoi/checkout/checkout_payment.md
+++ b/src/guides/v2.3/howdoi/checkout/checkout_payment.md
@@ -313,21 +313,22 @@ In your custom module directory, create a new `<your_module_dir>/view/frontend/l
 
 These steps are required in production mode only, not while in development mode.
 
-1.Compile the code:
+1. Compile the code:
 
-```bash
-bin/magento setup:di:compile
-```
+   ```bash
+   bin/magento setup:di:compile
+   ```
 
 1. Deploy the static contents:
 
-```bash
-bin/magento setup:static-content:deploy
-```
+   ```bash
+   bin/magento setup:static-content:deploy
+   ```
+
 1. Clean the cache:
 
-```bash
-bin/magento cache:clean
-```
+   ```bash
+   bin/magento cache:clean
+   ```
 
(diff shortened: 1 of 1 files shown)


devops-devdocs:

devops-devdocs:

src/guides/v2.3/howdoi/checkout/checkout_payment.md
```diff
@@ -308,5 +308,24 @@ In your custom module directory, create a new `<your_module_dir>/view/frontend/l
     </body>
 </page>
 ```
+## Step 5:Run the CLI Commands {#compile-code-deploy-contents-and-clean-the-cache}
```
rogyar: Hi @VinothKumar361. Thank you for your collaboration. 

I'm not sure we need this information trough the documentation. The DevDocs is a resource with the Magento development-related documentation. The target audience is Magento developers. 

If you are a Magento developer, you are supposed to know the very basic development routine like cache cleaning, running your development environment, using IDE/browser before creating working solutions. So, posting this info over and over again on all pages (it applies for most of the code change operations though) might not be as useful as we want it to see. 

Also, if we are talking about Magento development, the developer's environment is supposed to be in the `development` mode which does not require DI compilation and static content deployment. On production environments, the mentioned operations should be a part of the deployment process.

dobooth: You make a good point, Rogyar. There have been a few PRs adding this same info to other topics. I would rather lean on the side of providing more information than necessary, if it can prevent confusion to users. Good feedback.

dobooth:
running tests

src/guides/v2.3/howdoi/checkout/checkout_payment.md
dshevtsov: From the other hand, providing more information than necessary leads to future search and maintenance issues.
If the information is relevant and must be present in several topics, consider using the `{% include %}` functionality to follow the docs-like DRY principle.

PR #6579, opened 2020-02-06
Remove unneeded data files

dshevtsov:

This pull request (PR) removes 2.2 data completely (as they are not used anymore) ~and YAML duplicates in 2.3 data~.

## Additional information

Adding the `.ruby-version` file with the Ruby version we support. The file is supported by the version managers such as **rvm**, **rbenv**, and **chruby**.


dshevtsov:
@jeff-matthews we did use YAML before the 2.3.4 update because the original JSON file caused parsing issues in the past. After 2.3.4 release I added the data in the original format, and the build didn't fail. But looks like I forgot to remove the existing YAML files.

dshevtsov:
Actually, I just recalled that I already removed them in #6546

dshevtsov:
@dobooth when do you plan to merge `small changes`?

dshevtsov:
I removed the commit that deletes the YAML files in this PR.

dshevtsov:
running tests

dshevtsov:
> Will it still be possible to build 2.2 docs if we remove these files? Do we care?

no

PR #6602, opened 2020-02-10
Added example response for sample query for grouped product data types

VinothKumar361:

Added example response for sample query for grouped product data types

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/graphql/product/grouped-product.html


devops-devdocs:

rogyar:
From a technical perspective, this PR is correct. There's no exact place in the code (since we are talking about an API response). However, [this](https://github.com/magento/devdocs/pull/6602) part of the test expects exactly the same response as described in the current PR.

PR #6629, opened 2020-02-12
magento/devdocs#: Extend setShippingAddressesOnCart example

atwixfirster:

This pull request (PR) extends example of `setShippingAddressesOnCart` mutation usage. Current example a bit confused because we specify `region` and `country_code` attributes but do not retrieve them in the response.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/graphql/tutorials/checkout/checkout-shipping-address.html

- [region](https://github.com/magento/magento2/blob/2.3.4/app/code/Magento/QuoteGraphQl/etc/schema.graphqls#L217)
- [country](https://github.com/magento/magento2/blob/2.3.4/app/code/Magento/QuoteGraphQl/etc/schema.graphqls#L219)

Thank you!

### DIFF
--- a/src/guides/v2.3/graphql/mutations/set-shipping-address.md
+++ b/src/guides/v2.3/graphql/mutations/set-shipping-address.md
@@ -48,8 +48,16 @@ mutation {
         company
         street
         city
+        region {
+          code
+          label
+        }
         postcode
         telephone
+        country {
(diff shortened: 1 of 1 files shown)


devops-devdocs:

PR #6635, opened 2020-02-12
#6530 Formatting and grammar fixes

### LINKED ISSUE #6530: DocBlock standard
# Bug report

Correction of Article and Spelling incorrect.

## Description

On this page
https://devdocs.magento.com/guides/v2.3/coding-standards/docblock-standard-general.html

Following these standard is optional for third-party Magento 
In **this sentences** or **these standard** should be these standards or it should be this standard.

Formatting according phpDocumentor standard
In this sentences to should be added after **the** according.

The documentation should be as short as possible, but it should not skip necessary details.

in this **"the"** needs to be added before "necessary"

Each Magento source code file must have a DocBlock header with short description of the file. After the short description, there can be a long description.

in this **"a"** needs to be added before "short description"

Any file in Magento source should have a header with license and copyright notice.

in this **"the"** needs to be added before "Magento"

Exceptions are files with format that doesn’t support comments, and so no meta information can be added.
in this **"a"** or **"the"** needs to be added before "format"

License notice and copyright MUST be declared in the very beginning of the file.
in this **"at"** needs to be added before instead of in.

description for the element should be declared as a separate DocBlock.
**in this **"a"** needs to be added before description.**

Have file header DocBlock separately, then inclusion construct, then a DocBlock for the element with duplicated short description
**in this **"a"** needs to be added before duplicated.**

In general, typed method signature must be preferred over PHPDoc annotations whenever possible.
in this **"a"** needs to be added before typed.

The comment must describe why method is implemented and not how. For example:
in this **"the"** needs to be added before **method**.

The declaration of all arguments (if any) using @param tag, 
in this "**the"** needs to be added before **@param**

For example: @return Config|null.
in this sentence its appears that the **colon** is unnecessary.

Have file header DocBlock separately, then inclusion construct, then a DocBlock for the element with duplicated short description.
in this **"a"** needs to be added before "**duplicated short description**"

If an exception is thrown by a method,  sexplain the cause or situation.
sexplain should be **explain** not sexplain.

In general method return type signatures should be prefered over
in this prefered should be **Preferred**.

return statement without a value,
it's should be **value**.

@author**-**,
consider removing unnecessary space

If a child class method
it's should be **child's**

In general such method overrides are a code smell and 
comman is missing after the In **general.**

or to the entire class/interface. If the @api tag is applied at the file level,
**to** should be removed it.

It will be retained to provide backward compatibility until next major component release.
in this **"the"** **next** to be added before typed.

Motivation behind the added @deprecated tag MUST be explained. 
**The** should be added before the **Motivation**

This tag declares variables that will emerge in next lines of code as follows:
**The** should be added before the **next lines**

Some IDEs understand a different notation, where type is specified after variable name. This notation is also valid:
**The** should be added before the **variable**

Specifically, this is possible when a PHP-file composed from multiple file includes
it should be **files** not file.

Any other valid DocBlock tags may be specified if author deems 
unnecessary **comma** before the specified. and it's should be **the author**.

In a given DocBlock, style of formatting must be consistent.
the is missing before **the style.**

Did you find what you were lookin
it's should be **looking** not lookin spelling mistake.

### Steps to reproduce

1. First Step
1. Second Step
1. Etc.

dobooth:

This pull request (PR) fixes a slew of small grammatical errors in the topic.

Fixes #6530 

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/coding-standards/docblock-standard-general.html


src/guides/v2.3/coding-standards/docblock-standard-general.md
```diff
@@ -349,29 +347,28 @@ Functions and methods should have:
    *  There is more than one possible input/output type.
 
       For example: `@return Config|null`.
-      The DockBlock needs to explain what situations return `null`.
+      The DockBlock needs to explain what situations will return `null`.
```
keharper:
```suggestion
      The DocBlock should explain what situations will return `null`.
```

keharper: Changed the verb here to match line 353.

dobooth:
running tests

Issue #6643, opened 2020-02-13
Verifying Recommendations


atwixfirster:
# Topic clarification/correction request

## Topic Link

[Verifying Recommendations](https://devdocs.magento.com/recommendations/testing.html)

## Description

I've a question about `environmentId` value on the next screenshot:

<img width="1296" alt="01" src="https://user-images.githubusercontent.com/13585327/74443652-375d6400-4e7c-11ea-87a9-8b1cb6eafaf5.png">

I am verifying context data on my Magento instance and I do not see it.

## Possible solutions

We should notice, that customers should install the latest `magento/module-data-services` version (now it's `3.0.6`) to have an `environmentId`:

<img width="1110" alt="03" src="https://user-images.githubusercontent.com/13585327/74444981-5230d800-4e7e-11ea-9e6d-833a958b11af.png">

## Additional information

The source of mentioned data is - `Magento\DataServices\Block\Context`::`getStorefrontInstanceContext()`:

<img width="901" alt="02" src="https://user-images.githubusercontent.com/13585327/74443676-3debdb80-4e7c-11ea-88fb-7e4f35c5153a.png">

I am using `magento/module-data-services`:`3.0.3` and there is no `environmentId` field in the mentioned module version.

But for `magento/module-data-services`:`3.0.6` all is fine:

<img width="901" alt="6644" src="https://user-images.githubusercontent.com/13585327/74447056-69bd9000-4e81-11ea-9740-23cd83495dfd.png">

CC: @erikmarr

jcalcaben:
@erikmarr is this related to what you are currently working on?

erikmarr:
Thanks @atwixfirster for your feedback! Because this content is specific to the EAP, we don't want to specifically mention the new version the user has to install in the docs. Rather, we will publish (on Slack) when we push new updates so users can run the following to update their install:
`composer update magento/product-recommendations --with-dependencies`
Admittedly, we have not yet made our EAP participants aware of any changes formally so your proactiveness is appreciated! We will improve our communication going forward so you can be made aware of these changes to avoid confusion.

PR #6644, opened 2020-02-13
magento/devdocs#6643: Verifying Recommendations

atwixfirster:

This pull request (PR) fixes https://github.com/magento/devdocs/issues/6643.

## Affected DevDocs pages

- https://devdocs.magento.com/recommendations/testing.html
- https://devdocs.magento.com/recommendations/troubleshooting.html

-  `magento/module-data-services`:`3.0.6`

Thank you!


devops-devdocs:

atwixfirster:
Good point, @rogyar !

I've updated the suggested document. Also, I discovered that `environmentId` has been added in the `3.0.5` `magento/module-data-services` module version.

Could you please review ?

Thanks !

src/recommendations/troubleshooting.md
```diff
@@ -23,6 +23,7 @@ For issues related to the Catalog SaaS Export module:
        "magento/module-catalog-url-rewrite-data-exporter"
        "magento/module-configurable-product-data-exporter"
        "magento/module-data-exporter"
+       "magento/module-data-services" (use the latest version. Required >=3.0.5)
```
rogyar: I'm still not sure about the following assumption

>  The `saas-export` metapackage installs the following modules
> ...
> "magento/module-data-services"
> ...

From my understanding  `saas-export` and `data-services` are separate packages that are part of `product-recommendations` metapackage (see p.2 of [installation guide](https://devdocs.magento.com/recommendations/install.html#install))

But I may be wrong.

atwixfirster: @rogyar , your notice is correct. When you run 

```
composer require magento/product-recommendations
```

composer will install

- magento/module-data-services
- magento/module-product-recommendations
- magento/module-product-recommendations-layout
- magento/product-recommendations

Modules `magento/module-data-services` and `magento/module-data-services` should be installed separately.

rogyar: Hi @atwixfirster.

Let me explain in more examples. Please read the very first item of the [list](https://devdocs.magento.com/recommendations/troubleshooting.html#catalog-saas-export-module) which says:

> The saas-export metapackage installs the **following modules, all of which must be enabled**
> ...

After our modification in this PR we bring the following info:

- magento/module-data-services is a part of `saas-export` 
- magento/module-data-services MUST be enabled

which might be not 100% correct.

erikmarr: @rossbrandon can I get your thoughts on this?

rossbrandon: `composer require magento/product-recommendations` will install the following packages:
`magento/services-connector`
`magento/module-services-id`
`magento/module-data-services`
`magento/saas-export`

`magento/product-recommendations` should be the entry point rather than installing the other packages separately. 

Once installed, `composer update magento/product-recommendations --with-dependencies` will update the packages that `magento/product-recommendations` depends on as well.

erikmarr: @atwixfirster, @rogyar thank you for your feedback! The goal of this troubleshooting topic is to discuss issues with the Catalog SaaS Export module which does not include the data-services module. As @rossbrandon mentioned above, the product-recommendations module will install and handle these dependencies. While there are parts of the modules, such as the environment ID, which are introduced in one module and used in another, we want the documentation to reflect the recommended install process :`composer require magento/product-recommendations`.

erikmarr:
Closing PR.

PR #6646, opened 2020-02-13
Markdown linting: Fixed MD046 errors

jeff-matthews:

This pull request (PR) fixes all "fixable" MD046 errors in the devdocs project.

There are still MD046 errors that we cannot fix. The mdl tool identifies code blocks inside nested lists as errors. For example:

```
1. Lorem ipsum

   - Lorem ipsum

      ```bash
      mkdir
      ```
```

We can fix these "errors" by removing the line between the bullet and the code block, but that makes the source less readable. Since our goal with enabling linting rules is to promote the readability of source markdown, that's not a viable option.

For now, we should merge the fixes for legitimate errors and report this issue to the mdl maintainers. Until it's fixed however, we won't be enabling this rule on production.


dobooth:
The irony...

_checks/styles/style-rules-prod
```diff
@@ -35,4 +35,4 @@ rule 'MD038'
 exclude_rule 'MD039'
 rule 'MD040'
 exclude_rule 'MD041'
-exclude_rule 'MD046'
+rule 'MD046', :style => :fenced
```
dshevtsov:
```suggestion
rule 'MD046', :style => :fenced # code block style
```

src/guides/v2.3/config-guide/memcache/memcache_centos.md
```diff
@@ -20,20 +20,33 @@ To install memcached on CentOS, perform the following tasks as a user with `root
...
+   ```bash
+   yum install -y memcached
+   ```
+
+   ```bash
+   yum install -y php-pecl-memcache
+   ```
+
+   {:.bs-callout-info}
+   The syntax of the preceding commands might depend on what package repositories you use. For example, if you use webtatic and PHP 5.6, enter <code>yum install -y php56w-pecl-memcache</code>. Use `yum search memcache|grep php` to find the appropriate package name.
```
dshevtsov:
```suggestion
   The syntax of the preceding commands might depend on what package repositories you use. For example, if you use webtatic and PHP 5.6, enter `yum install -y php56w-pecl-memcache`. Use `yum search memcache|grep php` to find the appropriate package name.
```

_checks/styles/style-rules-prod
jeff-matthews: I shouldn't have committed that change. We don't want to turn the rule on.

jeff-matthews:
running tests

jeff-matthews:
running tests

PR #6676, opened 2020-02-18
Remove obsolete content from Cloud Guide

meker12:

Removes obsolete content from the _Cloud Guide_.


dshevtsov:
The [Set up an account and project](https://devdocs.magento.com/cloud/before/before-workspace-cloud-account.html) topic doesn't contain any group. Should it be removed too?

meker12:
> The [Set up an account and project](https://devdocs.magento.com/cloud/before/before-workspace-cloud-account.html) topic doesn't contain any group. Should it be removed too?

You are correct. Removed that file as well.

meker12:
running tests

PR #6680, opened 2020-02-19
DEVDOCS: Improving the page "create-cache-type"

andrewbess:

This pull request (PR) improves the page "Create a cache type"

1. The cache type configuration has added and described;
2. The code block for the cache type model has fixed according to PSR-12 and Magento coding standards;
3. I think it would be more right to use class_name structure like "**_VendorName\ModuleName\Model\Cache\Type\CacheType_**" and most of the classes of Magento CE use it. 12 VS 3. The class "**_Magento\Eav\Model\Cache\Type_**" is contained in these 3 classes. And "**_Magento\Customer\Model\Cache\Type\Notification_**" looks like well on my mind. So, I changed it.

## Affected DevDocs pages

- [Create a cache type](https://devdocs.magento.com/guides/v2.3/extension-dev-guide/cache/partial-caching/create-cache-type.html)


devops-devdocs:

src/guides/v2.3/extension-dev-guide/cache/partial-caching/create-cache-type.md
```diff
@@ -7,18 +7,75 @@ A *cache type* enables you to specify what is cached and enables merchants to cl
...
+This topic discusses how to create a custom cache type.
+
+### Cache type configuration {#m2devgde-cache-type-configuration}
+
+Declare a new cache type in the `etc/cache.xml` file with the following attributes:
+
+| Attribute | Required? | Description |
+| --- | --- | --- |
+| `name` | Yes | A unique cache type ID |
+| `translate` | No | Parameters that would be translated during to render "Cache Management" page |
```
rogyar: We may add a reference to the paragraph above for clarity 

```suggestion
| `translate` | No | Parameters (described below) that will be translated on the "Cache Management" page |
```

src/guides/v2.3/extension-dev-guide/cache/partial-caching/create-cache-type.md
```diff
@@ -7,18 +7,75 @@ A *cache type* enables you to specify what is cached and enables merchants to cl
...
+| `instance` | Yes | The cache type model |
```
rogyar:
```suggestion
| `instance` | Yes | The cache type model class |
```

src/guides/v2.3/extension-dev-guide/cache/partial-caching/create-cache-type.md
```diff
@@ -7,18 +7,75 @@ A *cache type* enables you to specify what is cached and enables merchants to cl
...
 {
-  const TYPE_IDENTIFIER = '%cache_type_id%';
-  const CACHE_TAG = '%CACHE_TYPE_TAG%';
+    /**
+     * Cache type code unique among all cache types
+     */
+    const TYPE_IDENTIFIER = '%cache_type_id%';
+
+    /**
+     * Cache tag used to distinguish the cache type from all other cache
```
rogyar: We may clarify this part since the explanation sounds pretty similar to the `TYPE_IDENTIFIER`. I would say that the cache tag is used for the cache scope (for cleaning operations as a practical example). See the [TagScope](https://github.com/magento/magento2/blob/2.3/lib/internal/Magento/Framework/Cache/Frontend/Decorator/TagScope.php) class for more info.

src/guides/v2.3/extension-dev-guide/cache/partial-caching/create-cache-type.md
```diff
@@ -33,10 +90,10 @@ You must specify the following parameters:
 You can get more information about caching by looking at the code.
 We suggest you locate classes that extend [Magento\Framework\Cache\Frontend\Decorator\TagScope][tagscope].
 
-For example, look at [Magento\Eav\Model\Cache\Type][type] to understand more about the EAV cache type.
+For example, look at [Magento\Customer\Model\Cache\Type\Notification][type] to understand more about the customer notification cache type.
```
rogyar: Please, provide in the PR description more info on why exactly the Notification cache type is better for the example than the EAV cache type.

Thank you

src/guides/v2.3/extension-dev-guide/cache/partial-caching/create-cache-type.md
andrewbess: Hello @rogyar. Thank you for your remark. I fixed it. Please check.

src/guides/v2.3/extension-dev-guide/cache/partial-caching/create-cache-type.md
andrewbess: Hello @rogyar. Thank you for your remark. I fixed it. Please check.

src/guides/v2.3/extension-dev-guide/cache/partial-caching/create-cache-type.md
andrewbess: Hello @rogyar 
All 15 classes of Magento CE have this message in the PHPDoc. That's why I was pasting it here.

src/guides/v2.3/extension-dev-guide/cache/partial-caching/create-cache-type.md
andrewbess: Hello @rogyar 
I think it would be more right to use class_name structure like "**_VendorName\ModuleName\Model\Cache\Type\CacheType_**" and most of the classes of Magento CE use it. 12 VS 3. The class "**_Magento\Eav\Model\Cache\Type_**" is contained in these 3 classes. And "**_Magento\Customer\Model\Cache\Type\Notification_**" looks like well on my mind.

I changed the description of this PR.

src/guides/v2.3/extension-dev-guide/cache/partial-caching/create-cache-type.md
rogyar: But we always may make things better, aren't we? :) 

From my point of view, the current explanation is not very informative for new developers. Cache type identifier is more or less self-explanatory it's a sort of UID. But the cache tag brings an additional "why?". 

If we have the cache type identifier that distinguishes the cache type from other types why do we need the cache tag that does the same job? 

That's why I think it might be a good idea to clarify this point. I like the explanation [here](https://devdocs.magento.com/guides/v2.3/extension-dev-guide/cache/partial-caching/create-cache-type.html) better. Also, if we could add a word/two (literally) about cache cleaning and cache tag connection it would be very useful for new devs. 

What do you think?

andrewbess: Hello @rogyar 
Of course.
I fixed it.
Please check.
Thank you in advance.

dobooth:
running tests

PR #6754, opened 2020-02-29
Removing deprecated region_id from createCustomerAddress Mutation

VinothKumar361:

This pull request (PR) removes deprecated region_id from the createCustomerAddress Mutation.As you can see in this page(https://devdocs.magento.com/guides/v2.3/graphql/mutations/create-customer-address.html) the region_id is used in request output but not showed in response.Also it creates confusion,because  to create customer addres there is no need to know or want region_id.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/graphql/mutations/create-customer-address.html

-  https://github.com/magento/magento2/blob/a3c2af0498d21e3f689bc8dd690d7c633a025276/app/code/Magento/CustomerGraphQl/etc/schema.graphqls#L129

### DIFF
--- a/src/guides/v2.3/graphql/mutations/create-customer-address.md
+++ b/src/guides/v2.3/graphql/mutations/create-customer-address.md
@@ -37,7 +37,6 @@ mutation {
     id
     region {
       region
-      region_id
       region_code
     }
     country_code


devops-devdocs:

dobooth:
@keharper

PR #6763, opened 2020-03-02
Prepare master for 2.4 contributions

dshevtsov:

This pull request (PR) introduces the changes that are necessary for 2.4 contribution but are also applicable for the devdocs mainline (master) and are to be ported to both 2.3.5-develop and 2.4.0-develop branches:

- Resolved internal redirects: replaced the internal links that were redirected with the actual destinations, and removed old redirects
- Moved both the MFTF and the MRG reference data to versioned directory to be ready to use both 2.3 and 2.4 data
- Cleaned up the table of contents files. Removed customizations for 2.0, 2.1, and 2.2 since we don't build them anymore
- ~Removed the canonical links plugin to decrease the build time. We don't need it since we build only 2.3 version right now. It will be restored in the 2.4.0-develop branch though, unless we find some other option to replace it.~
- Remove the documentation deprecation notifications that we don't use anymore.
- Clean up the GitHub templates a little

## Expectations

- For a devdocs writer, a maintainer, or a contributor, this should not affect typical workflows. The build time can decrease a little bit though.
- For the MFTF and MRG writers, this requires to consider the changed paths in the DocOps tooling.
- For the UX developer, this requires to consider restoring or replacing the canonical links plugin in the 2.4-develop
- For a devdocs website user who uses links in their resources, this can break some old links that worked through redirects.

## Additional information

Internal ticket: DOC-64
Internal preview: 1526


jeff-matthews:
Shouldn't I be seeing a bunch of symlinks from 2.3 to a new 2.4 guides directory?

dshevtsov:
> Shouldn't I be seeing a bunch of symlinks from 2.3 to a new 2.4 guides directory or is that supposed to come later?

Later. The symlinks will be added in a PR to the 2.4.0-develop branch. This PR is a part of preparation for that PR.

dshevtsov:
> I don't think we can remove the canonical links plugin as long as we still have 2.2 content that's live on the site.

That 2.2 content is freezed. This plugin does not process that content, since 2.2 is not part of the jekyll build anymore.

jeff-matthews:
Yeah, but won't removing the plugin eliminate the canonical URL from every 2.3 page on a subsequent build? Do we know what impact that might have on our SEO? If we have already researched it and it's a low risk, then I'm all for it. I just don't want to assume it will be ok without doing a little research first.

dshevtsov:
deleted the commit that removes the plugin

dshevtsov:
running tests

keharper:
Why are redirects being deleted? I restructured graphql in 2.3.3, and it seems like we're asking for unnecessary redirects to the front page.

dshevtsov:
> Why are redirects being deleted? I restructured graphql in 2.3.3, and it seems like we're asking for unnecessary redirects to the front page.

I added explanation in the PR description. Which redirects do you want me to restore?

jeff-matthews:
@keharper, are you referring to these files specifically:

- `src/guides/v2.3/graphql/product/bundle-product.md`
- `src/guides/v2.3/graphql/queries/index.md`
- `src/guides/v2.3/graphql/queries/url-resolver.md`

dshevtsov:
Restored the graphql redirects

dshevtsov:
running tests

Issue #6773, opened 2020-03-03
Text UI Component Example incorrect


paul-gene:
https://devdocs.magento.com/guides/v2.3/ui_comp_guide/components/ui-text.html

The example on this page does not work.

PR #6785, opened 2020-03-04
magento/devdocs#6773: Text UI Component Example incorrect

atwixfirster:

This pull request (PR) provides a fix for https://github.com/magento/devdocs/issues/6773

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/ui_comp_guide/components/ui-text.html

-  ...


devops-devdocs:

dobooth:
running tests

paul-gene:
Thank you for fixing this. The entire UI Component section of the docs needs work, it's utter useless as is.

atwixfirster:
Hi, @paul-gene !

Thank you for your opinion.

> Thank you for fixing this. The entire UI Component section of the docs needs work, it's utter useless as is.

We will do our best to add a necessary information. So, we need your help here. Could you please add a couple of questions which you want to clarify for a particular  UI component or components.

Let's do the existing tutorial more closer to a reader. Thanks again!

paul-gene:
1. Why were the examples I highlighted added in the first place if they don't work? Is <text> even a legit component name? If yes, what's the correct way of using it?
2. How do I use the `<formElements>` node? I know it can be used to add configuration options, but it is not mentioned _anywhere_
3. Why do I have to use `formElement="input"` on an input field, when it's the default value? (try leaving it out, you'll get an error!)
4. dataType - what types are there? What does it relate to? Again, no mention of them in the docs.

For each component, I want to see a **valid** example of how to use the field both as a new entry for data, and to show how to populate with data from a dataprovider

atwixfirster:
Useful questions! 

Thank you, @paul-gene ! 👍

PR #6795, opened 2020-03-04
Replace link tags with site variables

jeff-matthews:

This pull request (PR) replaces the Jekyll `link` tag with the `{{ site.baseurl }}` variable.

We should not use the `link` tag in the devdocs project because it doesn't support the variables we use, such as `{{ page.baseurl }}`.

Furthermore, using the `link` tag is incompatible with our "fast local build" workflow using the `_config.local.yml` file.

The tag forces jekyll to validate internal links during build time. If you're building only certain directories specified in the `_config.local.yml` file, this could create problems because it forces you to build directories you wouldn't otherwise need just to prevent Jekyll from failing on links to directories you're excluding.

## Affected DevDocs pages

Topics in the _Marketplace_ and _Compliance_ sections.


jeff-matthews:
running tests

PR #6822, opened 2020-03-10
Added input attribute for delete payment token mutation

VinothKumar361:

This pull request (PR) adds the input attribute and update the non nullable attributes value in deletePaymentToken Mutation.There is no input attributes mentioned in docs but it actually require 
public_hash input. The non nullable objects which mentioned in (https://github.com/magento/devdocs/blob/master/src/guides/v2.3/graphql/queries/index.md -The exclamation point indicates the value is non-nullable) is also not updated
## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/graphql/mutations/delete-payment-token.html

- https://github.com/magento/magento2/blob/cac512ffe55578d41c33146c6fe03ef6bdf851ee/app/code/Magento/VaultGraphQl/etc/schema.graphqls#L5
https://github.com/magento/magento2/blob/cac512ffe55578d41c33146c6fe03ef6bdf851ee/app/code/Magento/VaultGraphQl/etc/schema.graphqls#L22
https://github.com/magento/magento2/blob/cac512ffe55578d41c33146c6fe03ef6bdf851ee/app/code/Magento/VaultGraphQl/etc/schema.graphqls#L23
https://github.com/magento/magento2/blob/cac512ffe55578d41c33146c6fe03ef6bdf851ee/app/code/Magento/VaultGraphQl/etc/schema.graphqls#L24
https://github.com/magento/magento2/blob/cac512ffe55578d41c33146c6fe03ef6bdf851ee/app/code/Magento/VaultGraphQl/etc/schema.graphqls#L9


devops-devdocs:

keharper:
running tests

PR #6825, opened 2020-03-10
Update jsonEncoder to jsonSerializer

BrentRobert:

This pull request changes the deprecated json encoder to the json serializer on this page.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/themes/theme-images.html

### DIFF
--- a/src/guides/v2.3/frontend-dev-guide/themes/theme-images.md
+++ b/src/guides/v2.3/frontend-dev-guide/themes/theme-images.md
@@ -172,7 +172,7 @@ Check the following example on getting the breakpoints variable by the [`Gallery
  */
 public function getBreakpoints()
 {
-    return $this->jsonEncoder->encode($this->getVar('breakpoints'));
+    return $this->jsonSerializer->serialize($this->getVar('breakpoints'));
 }
 ```
 


devops-devdocs:

src/guides/v2.3/frontend-dev-guide/themes/theme-images.md
```diff
@@ -172,7 +172,7 @@ Check the following example on getting the breakpoints variable by the [`Gallery
  */
 public function getBreakpoints()
```
rogyar: Hi @BrentRobert. In the following example, we have a piece of code from the actual codebase (`/app/code/Magento/Catalog/Block/Product/View/Gallery.php`) as explains the sentence before the example. 

Unfortunately, the mentioned piece of code still uses the `encode` method 

https://github.com/magento/magento2/blob/2.3/app/code/Magento/Catalog/Block/Product/View/Gallery.php#L127

So we cannot simply reference the existing method and change only the documentation.

BrentRobert: That's my bad, didn't see that.

PR #6828, opened 2020-03-10
Remove large downloadable binary files from the repo

jeff-matthews:

> IMPORTANT:
> This PR will fail tests until we upload the binaries to a the new location on S3 (production) https://devdocs.magento.com/download/.

This pull request (PR) moves large downloadable binary files to a new location on S3 (production) so that we don't need to store them in our git repo.

These files take up about 220 MB of storage and don't belong in our repo. Files include:

- Adobe Illustrator (.ai)
- Adobe Photoshop (.psd)
- Archives (.zip)
- PDFs

## Affected devdocs pages

Most of the impact is in the Pattern Library guide on devdocs.

## Additional info

We will leave the cloud launch checklist PDF in the repo while Magento Support works on moving it to a new location.


dshevtsov:
Files also include `.sketch`.

dshevtsov:
The links `https://devdocs.magento.com/download/` will be tested as internal by link checker since `https://devdocs.magento.com` is set as an internal domain in the link checker config.

To avoid link checking issues on those links, the `:file_ignore:` list at `_config.checks.yml` needs another item:
```
- !ruby/regexp /\/download\//
```

jeff-matthews:
As long as those files exist on our domain, I want to validate links to them.

dshevtsov:
> As long as those files exist on our domain, I want to validate links to them.

They can only be validated as external links, since they will not be present in the `_site` directory after each build.

jeff-matthews:
Technically, you're correct. However, since we control where they are stored on our subdomain, we are accountable for ensuring that the links work. Ignoring them effectively means we will never validate links because the external link testing pipeline won't validate them either.

dshevtsov:
> we are accountable for ensuring that the links work

I agree. I'm not saying that we shouldn't do this, I'm saying that it is impossible to test those links on each build, because they will not be internal anymore.

> because the external link testing pipeline won't validate them either.

I don't see a reason why it won't.

jeff-matthews:
Oh, I see what you're saying. I assumed that as long as the files exist at the location, then the link will resolve. I haven't tested that yet though.

The external link checker does not validate links outside the devdocs.magento.com subdomain. Why would it validate the links to the binaries at `https://devdocs.magento.com/download`?

dshevtsov:
> The external link checker does not validate links outside the devdocs.magento.com subdomain

It does, this is its main purpose.
Clarification of terminology:
- external link checking is run by `rake test:external_links`, it tests all absolute links within `_site` directory
- internal link checking is run by `rake test:html`, it validates HTML output and links to pages within the `_site`

jeff-matthews:
Sorry, what I meant to write was that it _only_ tests link outside the devdocs.magento.com subdomain. 

So the external link check runs on any `https://devdocs.magento.com` link as long as it uses an absolute path? That doesn't make a lot of sense. "External" should mean all other domains besides devdocs.magento.com.

dshevtsov:
> So the external link check runs on any `https://devdocs.magento.com` link as long as it uses an absolute path?

Yes.

> "External" should mean all other domains besides devdocs.magento.com.

This can be set in a separate link checking config. But with such setup, we won't be able to test links to files that exist in S3 bucket only and are not part of the build, such as downloads in this case.

dshevtsov:
Looks like some images were not even linked. For example, I've got empty search result for `RWD_icon_sprite.psd`.

dshevtsov:
By the way, the link checker can help to identify what downloadables has references from topics actually.
It finds 34 failures in this branch now. (The cause of failures is discussed in the above conversation.)

jeff-matthews:
I'd still like to test that to be sure. I'm sure you're right, but I'm curious...

jeff-matthews:
Confirmed with @meker12 that we can move the cloud launch readiness PDF along with the rest of the binaries.

jeff-matthews:
You were right. It fails even when the files exist in production. I'll add the exclude.

jeff-matthews:
running tests

jeff-matthews:
running tests

dshevtsov:
> To avoid link checking issues on those links, the `:file_ignore:` list at `_config.checks.yml` needs another item:

It should be `:url_ignore:`. Sorry for the mistake.

PR #6875, opened 2020-03-16
#6869: Fix typo in code sample

### LINKED ISSUE #6869: Feedback on page: /guides/v2.3/frontend-dev-guide/validations/custom-validation.html
# General issue

## Description:

This code doesn't work
## Possible solutions:

return value.split(' ').length == 5 ? true | false; 
should be replaced with
return value.split(' ').length == 5 ? true : false;

atwixfirster:

This pull request (PR) fixes #6869. Small code typo.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/validations/custom-validation.html


devops-devdocs:

rogyar:
Hi @dobooth. @atwixfirster optimized this example. The following line:

```javascript
return value.split(' ').length == 5
```
Will return either `true` or `false` depends on the condition. So, there's no reason to have one more check as a ternary operator (? true : false). May I provide another example of such optimization.

```javascript
return a == true ? | true : false;
```

equals

```javascript
return a == true;
```

dobooth:
Ah of course. I thought it was just a | vs : thing.

PR #6895, opened 2020-03-19
Add documentation for SCD_RUN_BALER env variable for Magento Cloud

meker12:

Add documentation for SCD_RUN_BALER env variable for Magento Cloud to enable javascript bundling during the build process.

## Affected DevDocs pages

https://github.com/magento/devdocs/blob/master/src/cloud/env/variables-global.html
https://github.com/magento/devdocs/blob/master/src/cloud/env/variables-build.html
https://github.com/magento/devdocs/blob/master/src/cloud/release-notes/ece-tools.html

https://github.com/magento/ece-tools/pull/660

### DIFF
--- a/src/cloud/env/variables-build.md
+++ b/src/cloud/env/variables-build.md
@@ -37,7 +37,7 @@ stage:
 ```
 
 {:.bs-callout-info}
-You must install and configure the Baler module before using this feature. Because Baler is currently in alpha release, enable this option only on Staging environments.
+You must install and configure the Baler module before using this feature. Because Baler is currently in alpha release, we do not recommend using it in Production environments.
 
 ### `ERROR_REPORT_DIR_NESTING_LEVEL`
 
--- a/src/cloud/env/variables-build.md
+++ b/src/cloud/env/variables-build.md
@@ -21,6 +21,24 @@ The following variables were removed in v2.2:
 -  `skip_di_clearing`
 -  `skip_di_compilation`
 
+### `SCD_USE_BALER`
+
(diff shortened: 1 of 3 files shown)


src/cloud/env/variables-build.md
```diff
@@ -21,6 +21,24 @@ The following variables were removed in v2.2:
...
+Set to `true` to run Baler after performing static content deployment.
+
+```yaml
+stage:
+  build:
+    SCD_USE_BALER: true
+```
+
+{:.bs-callout-info}
+You must install and configure the Baler module before using this feature. Because Baler is currently in alpha release, enable this option only on Staging environments.
```
bbatsche: Just to make it clearer that you can try it in integration as well, but also to I think it's better to say we _recommend_ as it's optional.
```suggestion
You must install and configure the Baler module before using this feature. Because Baler is currently in alpha release, we do not recommend using it in Production environments.
```

src/cloud/env/variables-global.md
```diff
@@ -16,6 +16,24 @@ stage:
...
+Set to `true` to run Baler after performing static content deployment.
+
+```yaml
+stage:
+  build:
+    SCD_USE_BALER: true
+```
+
+{:.bs-callout-info}
+You must install and configure the Baler module before using this feature. Because Baler is currently in alpha release, enable this option only on Staging environments.
```
bbatsche:
```suggestion
You must install and configure the Baler module before using this feature. Because Baler is currently in alpha release, we do not recommend using it in Production environments.
```

PR #6916, opened 2020-03-24
Added note regarding use of explicit_defaults_for_timestamps=on

davidalger:

Per https://github.com/magento/magento2/issues/19597#issuecomment-573510633 and https://github.com/magento/magento2/issues/19597#issuecomment-580361785 the setting `explicit_defaults_for_timestamp=on` in `my.cnf` is required for Declarative Schema to report everything as up-to-date allowing for zero-down deployments.

If this setting is not enabled, setup:db:status will continuously return the following;

```
Declarative Schema is not up to date
Run 'setup:upgrade' to update your DB schema and data.
```

This affects MySQL, MariaDB and PerconaDB.

## Affected DevDocs pages

* https://devdocs.magento.com/guides/v2.3/install-gde/prereq/mysql.html#instgde-prereq-mysql-config


devops-devdocs:

nuzil:
Hi @davidalger 
According to this note:
https://devdocs-beta.magento.com/guides/v2.3/release-notes/release-notes-2-3-4-open-source.html#platform-upgrades

Before Magento 2.3.4, when using declarative schema with MariaDB 10.2, Magento threw an error indicating that the schema was not up-to-date after running setup:upgrade. With this release, we have normalized the values returned by MariaDB, which allows system integrators to use declarative schema with both MySQL and MariaDB. 

Which means this issue is not anymore required to have Mysql Setting set to make it works.

davidalger:
@nuzil I would encourage you to either try this for yourself (either MySQL 5.7 or MariaDB 10.x, does not matter which) or study what this setting does per the referenced MySQL 5.7 docs before coming to a conclusion based on a wildly vague line in the release notes. 

Magento prior to 2.3.4 was straight up broken, and `setup:db:status` did not report clean before at all. It will now, but only if the database is configured correctly. I.e. the database must be working as Magento normalization expects, otherwise it's going to see a difference and report a dirty status.

This setting disables non-standard and deprecated behavior in MySQL, which Magento is incompatible with. The change log really should have made a note about this setting being required for the normalization to work as it relies on the new and standard behavior that can only be achieved by enabling `explicit_defaults_for_timestamp`.

cmacdonald-au:
This setting has been tested and shown to be required, as per to comments in the mentioned issue.

To summarise the "why" behind the change (this is in the issue, but repeating here for clarity)..

Without that setting in place, the declarative db schema management does not agree with the actual schema in place for TIMESTAMP fields. It continuously reports the schema as different (which it will be) and attempts to change it (which will have no effect _because_ this setting changes how the changes are interpreted by MySQL).

cmacdonald-au:
@davidalger thanks for doing this :)

dobooth:
running tests

Issue #6912, opened 2020-03-24
PHP Extension ext-sockets not in requirement list


dverkade:
# Bug report

## Description

Magento requires the following library: php-amqplib/php-amqplib this requirement is listed in the root composer.json of Magento here: https://github.com/magento/magento2/blob/2.4-develop/composer.json#L45

This library required the PHP extension "ext-sockets" since 2018. However this extension is not listed in the required PHP extension list in the documentation. 

The library requires ext-sockets since 2018 as listed here: https://github.com/php-amqplib/php-amqplib/blob/master/CHANGELOG.md#281---2018-11-13

### Steps to reproduce

Install Magento without the ext-sockets PHP extension. This will fail.

## Expected result

Have the required extension added to the documentation.

## Possible solutions

Add requirement to the list of required PHP extensions.

atwixfirster:
Fixed in https://github.com/magento/devdocs/pull/6922.

Thank you for reporting, @dverkade

PR #6922, opened 2020-03-25
magento/devdocs#6912: PHP Extension ext-sockets not in requirement list

atwixfirster:

This pull request (PR) fixes https://github.com/magento/devdocs/issues/6912

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/install-gde/system-requirements-tech.html

- [php-amqplib/php-amqplib](https://github.com/magento/magento2/blob/2.4-develop/composer.json#L74)
- [ext-sockets](https://github.com/php-amqplib/php-amqplib/blob/master/CHANGELOG.md#281---2018-11-13)

Thank you!


devops-devdocs:

hostep:
It isn't required (yet), the dependency on `php-amqplib/php-amqplib` (which is the cause of the `sockets` requirement) is highly confusing at the moment.

See https://github.com/magento/magento2/issues/27200 for more info.

I posed the question in the #appdesign slack channel, and @melnikovi was going to try to follow up, so we should probably wait until a decision was made there around this before marking it as required.

atwixfirster:
@hostep , I think we should follow a pretty simple logic here: if you can not install Magento without `ext-sockets` PHP extension then that extension is a required. :) And such point does not require an architectural approve ;)

Again, I'm may be wrong... 

Thank you!

hostep:
Hi @atwixfirster 

Well, you **can** install Magento without `ext-sockets`, because you can install `php-amqplib/php-amqplib` v2.7.x which has no requirement on `ext-sockets`.

But to do that, you need to run `composer update` instead of `composer install`, so it can downgrade `php-amqplib/php-amqplib` from v2.10.x to v2.7.x

Herein lies the problem in my opinion, Magento should choose which version of `php-amqplib/php-amqplib` they want, preferably the most recent.
And if that is decided, then we can see if `ext-sockets` becomes required or not.

atwixfirster:
> Well, you **can** install Magento without `ext-sockets`, because you can install `php-amqplib/php-amqplib` v2.7.x which has no requirement on `ext-sockets`.

@hostep , I know you can install anything without all required components but not we are talking about a regular user.

Regarding to this point PR is ok.

Thank you and have a wonderful Friday!

hostep:
A regular user can install Magento without `ext-sockets`: when you setup a new project using `composer create-project --repository-url=https://repo.magento.com/ magento/project-community-edition=2.3.4` it will work because composer will notice that you don't have `ext-sockets`  installed and will only install `php-amqplib/php-amqplib` v2.7.x
Composer will ignore the `composer.lock` which is [here in github](https://github.com/magento/magento2/blob/2.3.4/composer.lock#L1394) even though it says to install `php-amqplib/php-amqplib` v2.10.x

So at the moment the only people having issues are the people cloning the git repo from github and trying to install it like this, I don't think those are regular people 🙂 

Anyway, I fully agree with you that this has to become a required extension, but I want to have some architect first deciding what to do with the [`php-amqplib/php-amqplib` constraint defined in the `composer.json` file](https://github.com/magento/magento2/blob/2.3.4/composer.json#L45), one of both versions needs to be removed in my opinion.

atwixfirster:
> Anyway, I fully agree with you that this has to become a required extension, but I want to have some architect first deciding what to do with the [`php-amqplib/php-amqplib` constraint defined in the `composer.json` file](https://github.com/magento/magento2/blob/2.3.4/composer.json#L45), one of both versions needs to be removed in my opinion.

ok ok ok @hostep 

Let's wait for the approval from architect.

Thank you!

dobooth:
Gents, thanks for the discussion. A couple note: Let's be careful editing `composer.lock`. For each release, we copy the composer.lock file from the main code repo. It is used as a source of truth for supported versions for a particular release. We can add an info-note below the list to call out this version dependency.

atwixfirster:
> A couple notes: Let's be careful editing composer.lock. For each release, we copy the composer.lock file from the main code repo. It is used as a source of truth for supported versions for a particular release. We can add an info-note below the list to call out this version dependency.

Thank you for notice, @dobooth ! It's a very useful.

atwixfirster:
> Let's remove this change and add an info box below the list to explain the issue here.

**fixed**

Will look like:



@hostep , could you please bless a proposed change?

Thank you!

hostep:
@atwixfirster: I would wait until an architect gives feedback, if it gets decided that `ext-sockets` becomes required, the composer requirements will change and it will get added to the `composer.lock` file probably. And then everything will happen automatically, no?

@dobooth: can you try to pull some strings and get some architect to get involved here?

dobooth:
I think this is a good compromise for now. We don't want to change composer.lock because that should match the released version. It is used a data file throughout Devdocs. I think the warning will suffice. I will check with the architects and see if they want to clarify this.

dobooth:
running tests

PR #6940, opened 2020-03-27
Layout: set block visibility using ACL resource example

vasilii-b:

This pull request (PR) adds an example of how to change a layout's block visibility dynamically using the ACL resource.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/layouts/xml-manage.html#ref_config_block

- https://github.com/magento/magento2/blob/2.3.4/app/code/Magento/Backend/view/adminhtml/layout/default.xml#L31
-  https://github.com/magento/magento2/blob/2.3.4/app/code/Magento/AdminNotification/view/adminhtml/layout/default.xml#L19
- https://github.com/magento/magento2/blob/2.3.4/lib/internal/Magento/Framework/View/Layout/etc/elements.xsd#L209


devops-devdocs:

dobooth:
running tests

PR #6980, opened 2020-04-03
Fix the API GET request functionality

### LINKED ISSUE #1721: REST usage notes need to be re-validated
The REST usage notes in  http://devdocs.magento.com/guides/v2.2/rest/notes.html states

> a user could potentially specify one ID in the URL and another in the request body. Whenever this type of conflict occurs, **Magento uses the value specified in the URL**.
For example:

> The REST URL to update a customer is <route url="/V1/customers/:id" method="PUT">. If you specify an ID value of 1 in the URL (http://magento.loc/customer/1), and a body of { "id": 2, "attr": "value" }, the customer with ID of 1 will be modified.

> This applies to all REST APIs where a parameter is passed in the URL. **Anything specified in the request body with the same parameter name as the URL will be ignored**.

However, I have tested in 2.0.x, 2.1.x and 2.2.2 and it looks like it is the other way around. **Magento is actually ignoring the parameter in the URL and using the value specified in the payload.**

Using the same example of modifying a customer.
These are some of my current users:
customers_backend

If I wanted to add middle name to customer with Id =4 but by accident in my request I type customer Id =5. Per documentation, customer id =5 should be modified, however, it is the customer with id =4 (specified in the payload) the one that gets modified.

request_and_response

updated_customers_backend

Please note that in order to update a customer, email, firstname, lastname and website id is required.
I cannot modify a customer with just a payload like the one mentioned in the documentation
`{ "id": 2, "attr": "value" }`

I also tested updating a product using /V1/products/{sku} and got the same results.

dobooth:

This pull request (PR) changes the description of how URL params vs. request body works.

Fixes #1721 

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/rest/notes.html

### DIFF
--- a/src/guides/v2.3/rest/notes.md
+++ b/src/guides/v2.3/rest/notes.md
@@ -16,7 +16,7 @@ For example:
 
 The REST URL to update a customer is `PUT /V1/customers/:id`. If you specify an ID value of `1` in the URL (`http://<host>/rest/<store_code>/V1/customers/1`), and a body of `{ "customer": { "id": 2, "attr": "value" }}`, the customer with ID of `2` will be modified.
 
-This applies to all REST APIs where a parameter is passed in the URL. Any value specified in the URL with the same parameter name as in the request body will be ignored.
+This applies to all REST APIs where a parameter is passed in the URL. Any value specified in the URL with the same parameter name as in the request body is ignored.
 
 {:.ref-header}
 Related topics
--- a/src/guides/v2.3/rest/notes.md
+++ b/src/guides/v2.3/rest/notes.md
@@ -1,23 +1,22 @@
 ---
 group: rest-api
-subgroup: A_rest
 title: REST usage notes
-menu_title: REST usage notes
-menu_order: 7
 functional_areas:
   - Integration
 ---
 
-## Parameters in URLs
+This topic describes issues you might encounter when using REST endpoints.
 
-In a REST call, a user could potentially specify one ID in the [URL](https://glossary.magento.com/url) and another in the request body.
+## Duplicate parameters in an REST call
+
+In a REST call, a user could potentially specify one `id` in the [URL](https://glossary.magento.com/url) and another in the request body.
 Whenever this type of conflict occurs, Magento uses the value specified in the request body.
 
 For example:
 
 The REST URL to update a customer is `<route url="/V1/customers/:id" method="PUT">`. If you specify an ID value of `1` in the URL (`http://<host>/rest/<store_code>/V1/customers/1`), and a body of `{ "customer": { "id": 2, "attr": "value" }}`, the customer with ID of `2` will be modified.
 
-This applies to all REST APIs where a parameter is passed in the URL. Anything specified in the the URL with the same parameter name as the request body will be ignored.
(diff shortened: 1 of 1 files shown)


src/guides/v2.3/rest/notes.md
```diff
@@ -1,22 +1,22 @@
...
-subgroup: A_rest
 title: REST usage notes
-menu_title: REST usage notes
-menu_order: 7
 functional_areas:
   - Integration
 ---
 
-## Parameters in URLs
+This topic describes issues you might encounter when using REST endpoints.
```
hguthrie, line 8: Why say "this topic"? I tend to treat content as if it could fit anywhere, even like an include; never know in the future. Why not something like "You might encounter the following issues when using REST endpoints."

Also, this topic is so short. Are you expecting more issues to mention?

src/guides/v2.3/rest/notes.md
```diff
@@ -1,22 +1,22 @@
...
 
-In a REST call, a user could potentially specify one ID in the [URL](https://glossary.magento.com/url) and another in the request body. Whenever this type of conflict occurs, Magento uses the value specified in the URL.
+## Duplicate parameters in an REST call
+
+In a REST call, a user could potentially specify one `id` in the [URL](https://glossary.magento.com/url) and another in the request body.
```
hguthrie, line 12: "could potentially" seems redundant

src/guides/v2.3/rest/notes.md
```diff
@@ -1,22 +1,22 @@
...
+Whenever this type of conflict occurs, Magento uses the value specified in the request body.
 
 For example:
 
-The REST URL to update a customer is `<route url="/V1/customers/:id" method="PUT">`.  If you specify an ID value of `1` in the URL (`http://<host>/rest/<store_code>/V1/customers/1`), and a body of `{ "customer": { "id": 2, "attr": "value" }}`, the customer with ID of `1` will be modified.
+The REST URL to update a customer is `<route url="/V1/customers/:id" method="PUT">`. If you specify an ID value of `1` in the URL (`http://<host>/rest/<store_code>/V1/customers/1`), and a body of `{ "customer": { "id": 2, "attr": "value" }}`, the customer with ID of `2` will be modified.
 
-This applies to all REST APIs where a parameter is passed in the URL. Anything specified in the request body with the same parameter name as the URL will be ignored.
+This applies to all REST APIs where a parameter is passed in the URL. Anything specified in the URL with the same parameter name as in the request body will be ignored.
```
hguthrie, line 19:
```suggestion
This applies to all REST APIs where a parameter is passed in the URL. Anything specified in the URL with the same parameter name as in the request body is ignored.
```

src/guides/v2.3/rest/notes.md
keharper: This topic exists because there was (and continues to be) no other place to put this info. But yes, at one point I expected this page to contain more content.

keharper:
Confirmed this is still the behavior. The entity specified by the `id` in the payload is updated, not the entity in the route.

keharper:
running tests

PR #6979, opened 2020-04-03
Update setup:install command to include Elasticsearch flags

keharper:
## Purpose of this pull request

This pull request (PR) updates the setup:install command to include new flags for installing Elasticsearch.
Related internal ticket: MC-20710

## Affected DevDocs page

-  https://devdocs.magento.com/guides/v2.4/install-gde/install/cli/install-cli-install.html

## Links to Magento source code

-  ...

whatsnew
Added several new flags for including Elasticsearch when installing Magento from the command line. See [Install the Magento software](https://devdocs.magento.com/guides/v2.4/install-gde/install/cli/install-cli-install.html) for details.

PR #7039, opened 2020-04-15
Update installation wizard for Elasticsearch

keharper:

In 2.4, Elasticsearch is required for all installations, because MySQL is no longer supported as the catalog search engine.

This pull request (PR) makes the following categories of changes:

* I updated the installation Wizard documentation to include a new step for configuring Elasticsearch.

* I moved ES installation info from the Config Guide to the Install Guide. Not all ES topics are relevant for installation, so that content will remain in the Config Guide.

* I updated diagrams and screenshots relating to the wizard and installation process.

* While updating the Wizard documentation, I discovered that outdated files regarding hosted installation are still present in the filesystem. I deleted those files. Many of these files contained references to `include` files that were also used in the Wizard documentation. Since this content is no longer shared, I moved it to be inside the Wizard topics.

Note: I didn't delete the include files, because they are still referenced in the 2.3 documentation. Updating 2.3 files is outside the scope of this PR.

Related PR: #6979 

## Affected DevDocs pages

-  Many


keharper:
running tests

Issue #6411, opened 2020-01-16
Feedback on page: /cloud/project/project-upgrade.html


meker12:
# General issue

Most merchants on the Magento Cloud platform do not need to submit a ticket to remove mcrypt - only if they're not currently on the Wings interface

Remove the following note from the [Upgrade ](https://devdocs.magento.com/cloud/project/project-upgrade.html) topic: 

If you use PHP version 7.2, you must remove the mcrypt extension from the extensions section of the .magento.app.yaml file. For Pro projects, you need to create a support ticket to completely disable the mcrypt extension.
(edited)

Add note about PHP version upgrades:

When you upgrade the php version, you must submit a support ticket to update the New Relic and Blackfire extensions.

meker12:
Information about New Relic has not been added. Will close ticket when that task has been completed.

PR #7040, opened 2020-04-15
Cloud: remove statement to call support

hguthrie:
The `mcrypt` extension has been removed from the Pro environments. We no longer need to provide instruction to call support, as seen in the [Complete the upgrade](https://devdocs.magento.com/cloud/project/project-upgrade.html#complete-the-upgrade) section.

I noticed this was part of the request in #6411 , but I did not address anything related to other extensions. This was a quick plea that I received.


dobooth:
running tests

Issue #7029, opened 2020-04-12
for text column in db_schema the attribute 'length' is not allowed.


ioweb-gr:
# Bug report

https://devdocs.magento.com/guides/v2.3/extension-dev-guide/declarative-schema/db-schema.html
In the docs it suggests that we can use length attribute for text column nodes however it doesn't seem to be the case as magento 2 throws an error.

Element 'column', attribute 'length': The attribute 'length' is not allowed.

## Description

Currently in the docs it mentions that we can use length attribute to define the text column size however Magento 2 won't allow it failing to create the column. 

e.g. from the docs

```

<schema xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"                  xsi:noNamespaceSchemaLocation="urn:magento:framework:Setup/Declaration/Schema/etc/schema.xsd"> 
    <table name="declarative_table">
         <column xsi:type="int" name="id_column" padding="10" unsigned="true" nullable="false" comment="Entity Id"/>         
<column xsi:type="int" name="severity" padding="10" unsigned="true" nullable="false" comment="Severity code"/>
 -       <column xsi:type="varchar" name="title" nullable="false" length="255" comment="Title"/>
 +       <column xsi:type="text" name="title" nullable="false" length="255" comment="Title"/>         
<column xsi:type="timestamp" name="time_occurred" padding="10" comment="Time of event"/>         
<constraint xsi:type="primary" referenceId="PRIMARY">
             <column name="id_column"/>
         </constraint>
     </table>
 </schema>
--
```

### Steps to reproduce

Create a db_schema.xml and define a table with a column like this 
`<column name="validate_rules" xsi:type="text" length="255" comment="Validate Rules" />`

## Expected result

Either the column is created correctly or the documentation is updated. We need a definition on how to set the size for text columns via db_schema.xml

atwixfirster:
Good catch, @ioweb-gr !

When you use a `TEXT` type for a column you don’t have to specify a storage length like for `CHAR` and `VARCHAR`.

Fixed in https://github.com/magento/devdocs/pull/7045.

Thanks

PR #7045, opened 2020-04-16
magento/devdocs#7029: for text column in db_schema the attribute 'length' is not allowed.

atwixfirster:

This pull request (PR) fixes https://github.com/magento/devdocs/issues/7029.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/extension-dev-guide/declarative-schema/db-schema.html

-  [TEXT](https://www.mysqltutorial.org/mysql-text/) MySQL string data type

Thank you!


devops-devdocs:

rogyar:
An example from the codebase 

https://github.com/magento/magento2/blob/5f3b86ab4bd3e3b94e65429fed33f748d29c1bbe/app/code/Magento/Newsletter/etc/db_schema.xml#L44

dobooth:
running tests

milindsingh:
@atwixfirster I think this can also be updated.

Configure declarative schema   Magento 2 Developer Documentation

atwixfirster:
> @atwixfirster I think this can also be updated.

@milindsingh , **done** in https://github.com/magento/devdocs/pull/7057

Thanks for notice!

lano-vargas:
Also in the docs you may want to update it has an example with type text which contains length  ref: https://developer.adobe.com/commerce/php/development/components/declarative-schema/configuration/#add-an-index

PR #7054, opened 2020-04-17
Add redirect for deprecated Blackfire.io content

meker12:

Add redirects for deprecated Blackfire.io content.  Blackfire is no longer bundled with Magento Cloud. Customers can still use Blackfire by setting up their own account and following the [Blackfire configuration instructions](https://support.blackfire.io/en/articles/771436-step-3-configure-blackfire-to-run-in-all-magento-cloud-environments).

## Affected DevDocs pages

- https://devdocs.magento.com//cloud/project/project-integrate-blackfire.html

### DIFF
--- a/src/cloud/project/project-integrate-blackfire.md
+++ b/src/cloud/project/project-integrate-blackfire.md
@@ -1,0 +1,12 @@
+---
+group: cloud-guide
+title: Blackfire.io for Magento Cloud
+functional_areas:
+  - Cloud
+  - Integration
+  - Setup
+  - Services
+redirect_from:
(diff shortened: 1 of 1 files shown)


src/cloud/project/project-integrate-blackfire.md
```diff
@@ -0,0 +1,12 @@
+---
+group: cloud-guide
```
jeff-matthews, line 2: I don't think you need `group` or `title`. Unless you think keeping `title` would be useful for future refrence.

You definitely don't need `functional_areas`.

meker12:
running tests

PR #7098, opened 2020-04-26
EAV entity types information added

ajithkumar-maragathavel:

This pull request (PR) adds information about the different EAV entity types available in the Magento2. Each and every EAV attribute created in Magento must be associated with any of the EAV entity types.

So I have listed out all the EAV entity types from the **eav_entity_type** table.

## Affected DevDocs pages

-  It's a new page. (/extension-dev-guide/eav-entity-types.html)

### DIFF
--- a/src/_data/toc/php-developer-guide.yml
+++ b/src/_data/toc/php-developer-guide.yml
@@ -153,6 +153,9 @@ pages:
         - label: EAV and extension attributes
           url: /extension-dev-guide/attributes.html
 
+        - label: EAV Entity Types
+          url: /extension-dev-guide/eav-entity-types.html
+
         - label: Plugins (Interceptors)
           url: /extension-dev-guide/plugins.html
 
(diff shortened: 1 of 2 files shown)


devops-devdocs:

aleron75:
Hello @ajithkumar-maragathavel, first of all, thank you for your contribution.

I will close this PR but this gives me the chance to clarify an aspect regarding EAV entities.

The list of entity types from the `eav_entity_type` is a wrong inheritance from the past: only the following entities are EAV ones:

*  customer
*  customer_address
*  catalog_category
*  catalog_product

We can verify which are EAV resource models by searching those extending the `\Magento\Eav\Model\Entity\AbstractEntity`, class.

Non-EAV resource models extend `\Magento\Framework\Model\ResourceModel\Db\AbstractDb` instead.

As a side note, this is also true for Magento 1, at least from version 1.3.x if I remember correctly.

Best.
Alessandro

PR #7099, opened 2020-04-26
Added params with detail for module:status command

chandru-199:
## Purpose of this pull request

This pull request (PR) - Added params with details for module:status command.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/install-gde/install/cli/install-cli-subcommands-enable.html

PR #7101, opened 2020-04-26
Example command not clear

chandru-199:

This pull request (PR) - In this document there is an example to check the module status of the extension. In that example the status checked by the command,

module:status

It will return all enabled and disabled modules. So we can check the particular module's status by passing that module name as argument.

## Affected DevDocs pages

-  guides/v2.3/comp-mgr/install-extensions.html


devops-devdocs:

keharper:
Merge this PR with #7099. These are two small PRs about the `bin/magento module:status` command.

PR #7132, opened 2020-04-29
Add info about Composer dependency for upgrade plugin

jeff-matthews:

This pull request (PR) updates the note about the compatibility of the Composer plugin for upgrading Magento. It is not compatible with Magento 2.3.5.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/comp-mgr/cli/upgrade-with-plugin.html

whatsnew
Added a [note](https://devdocs.magento.com/guides/v2.3/comp-mgr/cli/upgrade-with-plugin.html) about Composer upgrade plugin compatibility with Magento 2.3.5.

### DIFF
--- a/src/guides/v2.3/comp-mgr/cli/upgrade-with-plugin.md
+++ b/src/guides/v2.3/comp-mgr/cli/upgrade-with-plugin.md
@@ -11,7 +11,7 @@ functional_areas:
 -->
 
 {:.bs-callout-warning}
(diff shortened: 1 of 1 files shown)


dshevtsov:
This info would be good-to-know for devdocs users. I think it makes sense to change the Editorial label to `Technical` and add whatsnew.

jeff-matthews:
running tests

PR #7135, opened 2020-04-30
magento/devdocs#: Cloud. Upgrade Magento to 2.3.5

atwixfirster:

This pull request (PR) removes version-specific information from the command instruction to upgrade the Magento version for Cloud Projects via composer. This improvement means we won't have to update the instruction when each time a Magento version is releases.

## Affected DevDocs pages

-  https://devdocs.magento.com/cloud/project/project-upgrade.html

-  [Cloud 2.3.5](https://github.com/magento/magento-cloud/blob/2.3.5/composer.json#L21)

Thank you!


devops-devdocs:

src/cloud/project/project-upgrade.md
```diff
@@ -107,7 +107,7 @@ To upgrade the Magento version:
 1. Change to your Magento root directory and set the upgrade version using the [version constraint syntax].
 
    ```bash
-   composer require "magento/magento-cloud-metapackage":">=2.x.2 <2.x.3" --no-update
+   composer require "magento/magento-cloud-metapackage":">=2.3.5 <2.3.6" --no-update
```
dmytro-ch, line 110: Do we really need to specify the exact version here?
I guess it's rather an example of version constraint syntax, isn't it?

atwixfirster: My idea is: provide a actual example how to upgrade Cloud to the latest Magento version.

Let's ask @meker12 for help here.

@meker12, should we use in the example the latest Magento version or we should leave all as it is here?

Thank you!

rogyar: @atwixfirster your example shows how to upgrade to the exact version, not to the latest one (in this particular case we may consider as latest one but the version constraint syntax is about the exact patch version). 

If we are talking about Magento upgrade, literally 

> When upgrading to 2.3.x from 2.2.x ...

Then I assume that we may use 

```bash
composer require "magento/magento-cloud-metapackage":"2.3.*" --no-update
```

dmytro-ch: The key point is that we should keep using the constraint approach in any case. This is the appropriate way to keep ece-tools up to date along with the patch versions automatically installed on Magento Cloud.

What I wanted to say is that the aforementioned document is not intended to show how to upgrade to the concrete Magento version. I think the initial goal was to highlight the general approaches to be used in order to upgrade Magento and provide some recommendations depending on a situation.

> My idea is: provide a actual example how to upgrade Cloud to the latest Magento version.

@atwixfirster if you wish to apply this change, I guess the following page is the right place :)
https://devdocs.magento.com/cloud/project/ece-tools-upgrade-project.html#metapackage

meker12, line 110: Change the version number in the code sample to use placeholders for current and next version--same technique used in the [Upgrade to ece-tools](https://devdocs.magento.com/cloud/project/ece-tools-upgrade-project.html#metapackage) topic.

```suggestion
   composer require "magento/magento-cloud-metapackage":">=current_version <next_version" --no-update
```



On line 114, there is a note: 

```   You must use the version constraint syntax to successfully update the `{{site.data.var.ct}}` package.```

Update that note to add info about where they can find the most current version constraint: 

```You must use the version constraint syntax to successfully update the `{{site.data.var.ct}}` package. You can always find the latest version constraint in the [`magento-cloud` template](https://github.com/magento/magento-cloud/blob/master/composer.json).```

atwixfirster: **implemented**

meker12:
running tests

PR #7139, opened 2020-04-30
Use of view models section updated

ajithkumar-maragathavel:

This pull request adds information on view models usage. Added a caution for users to use view model class instead of helper class in templates.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/extension-dev-guide/view-models.html

whatsnew
Expanded the example code in the [View Models](https://devdocs.magento.com/guides/v2.3/extension-dev-guide/view-models.html) topic.


devops-devdocs:

src/guides/v2.3/extension-dev-guide/view-models.md
```diff
@@ -14,6 +14,19 @@ Use this approach anytime you need to inject functionality into template files a
  {:.bs-callout-info}
 View models are available in Magento 2.2 onwards. If your code must be compatible with older versions of Magento, consider adding your logic to blocks. For more information about backward compatibility, see [Backward compatibility]({{ page.baseurl }}/contributor-guide/backward-compatible-development/).
 
+The following is an example of helper class directly used in `Magento/Catalog/view/frontend/templates/product/widget/new/content/new_grid.phtml` templates.
+
+```xml
+$postDataHelper = $this->helper(Magento\Framework\Data\Helper\PostHelper::class);
```
rogyar, line 20: Hi @ajithkumar-maragathavel. The described example is good but I'm not sure we should show how to use helpers in view models documentation. If there's a way to use the view model in the similar way, we may put this information. Otherwise, using helper is out of context.

ajithkumar-maragathavel: @rogyar I will provide another example below the current one to show how we can achieve the same functionality using the view model concept. Will that works?
Because my whole idea is to inform developers to use the view models instead of the helper class. Since It is the best practice.

If it is not okay, Please provide me some suggestions.
Thanks for the review.

rogyar: Hi @ajithkumar-maragathavel. I got your idea, thank you for the explanation. My point is, I'm not sure we should show the helper usage in the example. If we show an example of something, the reader get's it as a good practice first of all (even if then we have a mention that it's not a good practice). 
If you have a possibility to replace the helper example with the view model example somehow, it would make this topic much more useful. 

Thank you!

ajithkumar-maragathavel: @rogyar Thanks for your suggestion. I can understand your point now. I will come up with the view model example.
Thanks

ajithkumar-maragathavel: @rogyar I have updated the document with the view model example from Magento_Catalog module. Please review it.
Let me know your thought on this example.
Thanks.

rogyar:
Hi @ajithkumar-maragathavel. Thank you for the update, now the examples are more robust. To make the change more consistent, let's move the examples to the "Examples of View models in Magento" sections since these examples show exactly how the view models implemented in the core.

Thank you!

ajithkumar-maragathavel:
> Hi @ajithkumar-maragathavel. Thank you for the update, now the examples are more robust. To make the change more consistent, let's move the examples to the "Examples of View models in Magento" sections since these examples show exactly how the view models implemented in the core.
> 
> Thank you!

@rogyar Thanks for the suggestion. I have moved the example to corresponding area in the document. Please review it.

dobooth:
running tests

PR #7173, opened 2020-05-07
Updated the customer query intro by adding Wishlist as part of customer response

VenkataRamanaGandham:

This pull request (PR) is to update the customer information by adding wishlist as part of customer query result. So that, users can understand that wishlist query is deprecated and added as part of customer query.

Updated Wishlist attributes table by adding array to the WishlistItem, which is showing singular earlier causing confusion to the user.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/graphql/queries/customer.html

-  ...


devops-devdocs:

keharper:
running tests

PR #7176, opened 2020-05-08
Magento 2.4.0 requires minimum stable MySQL 5.7.x version

adifucan:

Lock manager on the basis of MySQL is now a primary locker instead of lock manager on the basis of the caching system. As a result now minimal stable version of MySQL which is 5.7.9 is required.

## Affected DevDocs pages

-  src/guides/v2.4/install-gde/prereq/mysql.md

Here is PR to Magento 2.4 that makes DB lock manager a primary locker: https://github.com/magento/magento2ce/pull/5651


devops-devdocs:

keharper:
@adifucan would you also update src/guides/v2.4/install-gde/system-requirements-tech.md? The symlink to this file has already been broken.

@meker12 Can you comment whether `cloud/project/project-conf-files_services-mysql.html` should be updated in the 2.4 branch?

meker12:
Not sure if this restriction will apply to projects deployed on Cloud since Cloud uses an automatically configured file lock provider.  See [Configure the lock](https://devdocs.magento.com/guides/v2.3/install-gde/install/cli/install-cli-subcommands-lock.html#instgde-cli-lockconfig) 

@BaDos  @andriyShevtsov  Can you confirm?

andriyShevtsov:
I assume that it should still works as before. But in subject mentioned MySQL 5.7.x and on cloud we use MariaDB. Here is mapping https://mariadb.com/kb/en/mariadb-vs-mysql-compatibility/

src/guides/v2.4/install-gde/system-requirements-tech.md
```diff
@@ -33,7 +33,7 @@ Upgrading the Magento applications and extensions you obtain from Magento Market
 
 ## Database
 
-MySQL 5.6, 5.7
+MySQL 5.7
```
dobooth, line 36: If the minimum is 5.7.9 we should say so here.

dobooth:
Also, note that we cannot make editorial changes ourselves on PRs made from forks, but we can from branches.

src/guides/v2.4/install-gde/system-requirements-tech.md
adifucan: 5.7.9 is a minimal stable version of 5.7 and magento supports only stable versions of a software. Anyway, it's better to clarify! Fixed.

adifucan:
@dobooth shall PR created not from forks but from https://github.com/magento/devdocs/ branches?

dobooth:
@adifucan It's easier for us to process PRs from branches. With forks, we cannot make small edits directly and we cannot update the branch before merging. It is not required that you use branches but it help us keep things moving and we don't have to bug you as often.

jeff-matthews:
I thought we determined that problem was with the settings in the forked repo restricted direct editing.

dobooth:
running tests

PR #7223, opened 2020-05-15
Fixed formatting issue

konarshankar07:

This pull request (PR) will fix the small formatting issue

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/test/unit/writing_testable_code.html


devops-devdocs:

src/guides/v2.3/test/unit/writing_testable_code.md
```diff
@@ -158,7 +158,7 @@ As a rule of thumb, try to keep methods to five or fewer lines of code.
 
 Functions should do only one thing and they should do it very well.
 
-Once you respect the [single responsibility principle][]{:target="_blank"}, you will know exactly what you are testing and your functions will be smaller and clearer.
+Once you respect the [single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle){:target="_blank"}, you will know exactly what you are testing and your functions will be smaller and clearer.
```
rogyar, line 161: Hi @konarshankar07. Thank you for your collaboration. As a rule, we don't need to provide the URL explicitly (inline) in the link if we have the URL in "Links definitions". 
After this change we have the URL in both places: "Link definitions" and inline which does not seem to be the best practice.

src/guides/v2.3/test/unit/writing_testable_code.md
```diff
@@ -27,18 +27,18 @@ We strongly recommend you do *not*:
...
  {:.bs-callout-info}
-This rule applies only to production code. When writing [integration tests]({{ page.baseurl }}/test/integration/integration_test_execution.html), this is not true. In fact, the object manager is recommended for integration tests.
+This rule applies only to production code. When writing [integration tests][integration-tests], this is not true. In fact, the object manager is recommended for integration tests.
 
 ### Collaborator classes
 
 Whenever an external class property, class constant, or a class method is used in a file, this file depends on the class containing the method or constant. Even if the external class is not used as a instantiated object, the current class is still hard-wired to depend on it.
 
-[PHP](https://glossary.magento.com/php) cannot execute the code unless it can load the external class, too. That is why such external classes are referred to as *dependencies*. Try to keep the number dependencies of to a minimum.
+[PHP] cannot execute the code unless it can load the external class, too. That is why such external classes are referred to as *dependencies*. Try to keep the number dependencies of to a minimum.
```
rogyar, line 39:
```suggestion
[PHP][PHP] cannot execute the code unless it can load the external class, too. That is why such external classes are referred to as *dependencies*. Try to keep the number dependencies of to a minimum.
```

konarshankar07: image
If link defination name and text is same then we don't need to specify the link defination name

rogyar: That's right, we are good to go then. 

Thank you

dobooth:
running tests

PR #7229, opened 2020-05-18
Speeding up npm install --prefer-offline --no-audit --no-optional

lfluvisotto:
Speeding up npm install --prefer-offline --no-audit --no-optional

Since we use npm in the project just to install the necessary packages to compile LESS using Grunt.

We must focus on speed and performance in our development environment, staging, uat, production, etc.

We do not need every new run of npm install:

1) Use the network, download the packages and install them again, if we have them in cache. By default, verifies packages against the registry after a minimum cache time to see if the package is still good, so prefering offline mode, no network requests will be done during install. So --prefer-offline should be used.

2) Receive during the installation of package audit, is anyone looking for these results? I think they can be suppressed. So --no-audit should be used.

3) Install package dependencies that we are not going to use. So --no-optional should be used.

Using the additional arguments with npm it will be 15% faster to install the packages. Faster than npm, it would be if we use pnpm, but it would be for another scope.

## Affected DevDocs pages

https://devdocs.magento.com/cloud/project/project-conf-files_magento-app.html

https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/css-topics/css_debug.html

https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/tools/using_grunt.html

https://devdocs.magento.com/page-builder/docs/getting-started/install-pagebuilder.html

https://devdocs.magento.com/guides/v2.3/test/js/jasmine.html

https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/css-topics/gulp-sass.html

### DIFF
--- a/src/cloud/project/project-conf-files_magento-app.md
+++ b/src/cloud/project/project-conf-files_magento-app.md
@@ -260,7 +260,7 @@ dependencies:
 hooks:
     build: |
         cd public/profiles/project_name/themes/custom/theme_name
-        npm install
+        npm install --prefer-offline --no-audit --no-optional
         grunt
         cd
         php ./vendor/bin/ece-tools build
--- a/src/compliance/privacy/__adobe-privacy-js-library.md
+++ b/src/compliance/privacy/__adobe-privacy-js-library.md
@@ -15,7 +15,7 @@ Use this library to retrieve and remove the data stored in the browser by these 
 
 Use one of the following methods to download the library file:
 
--  npm: `npm install @adobe/adobe-privacy`
+-  npm: `npm install --prefer-offline --no-audit --no-optional @adobe/adobe-privacy`
 -  GitHub: <https://github.com/Adobe-Marketing-Cloud/adobe-privacy>
 
 After you have the file, you will need to add it to a custom module or theme installed in your Magento instance.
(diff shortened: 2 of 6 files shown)


devops-devdocs:

jcalcaben:
running tests

meker12:
Checking with Magento security team to determine whether installing npm with the `no-audit` option adversely impacts security.

lfluvisotto:
Hi @meker12 , for those who want to audit the code just need to remove the argument.

In production servers or during the Continuous Integration, Continuous Delivery, Continuous Deployment process doesn't make any sense we keep the audit.

src/cloud/project/project-conf-files_magento-app.md
```diff
@@ -260,7 +260,7 @@ dependencies:
 hooks:
     build: |
         cd public/profiles/project_name/themes/custom/theme_name
-        npm install
+        npm install --prefer-offline --no-audit --no-optional
```
bbatsche, line 263: Build environments on cloud aren't (necessarily) saved across build runs, and looking at some [published benchmarks](http://www.tiernok.com/posts/2019/faster-npm-installs-during-ci/) it seems that using `npm ci` would provide provide much better performance than any of these other flags.

meker12:
@lfluvisotto Instead of providing a prescriptive example, perhaps you could add a note that adding options to the npm install command can improve performance with some suggestions and a link to the npm install documentation.

PR #7233, opened 2020-05-18
Cache clean cmd added in bundling js guide

Mohamed-Asar:

Cache clean command is added in bundling js guide. it is must to clean cache after updating a config

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/themes/js-bundling.html


devops-devdocs:

PR #7255, opened 2020-05-20
MC-34097 Update supported PHP version

dobooth:

This pull request (PR) updates v2.4 to use the latest supported PHP version.

whatsnew
Updated the [supported versions of PHP](https://devdocs.magento.com/guides/v2.4/install-gde/system-requirements-tech.html#php) to 7.3 and 7.4.


keharper:
I see several v2.3 files in this PR. Is that intentional?

src/guides/v2.3/install-gde/prereq/merch_install.md
```diff
@@ -21,9 +21,9 @@ If you don't use GoDaddy, you might need to contact your hosting provider's tech
 
 Magento requires the following software to run:
 
-*  Web server: Apache 2.2 or 2.4
-*  Programming language: [PHP](https://glossary.magento.com/php) 5.6.x or 5.5.x
-*  Database: MySQL 5.6.x
+*  Web server: Apache 2.4
+*  Programming language: [PHP](https://glossary.magento.com/php) 7.3.x or 7.4.x
```
jeff-matthews, line 25: Will Magento 2.3.x be compatible with PHP 7.4 when Magento 2.4.0 releases?

src/guides/v2.3/install-gde/prereq/merch_install.md
```diff
@@ -21,9 +21,9 @@ If you don't use GoDaddy, you might need to contact your hosting provider's tech
...
+*  Database: MySQL 5.7.9 or 8.0
```
jeff-matthews, line 26: Will Magento 2.3.x be compatible with MySQL 8.0 when Magento 2.4.0 releases?

src/guides/v2.3/install-gde/prereq/nginx.md
```diff
@@ -44,20 +44,20 @@ To install and configure `php-fpm`:
 1. Install `php-fpm` and `php-cli`:
 
    ```bash
-   apt-get -y install php7.2-fpm php7.2-cli
+   apt-get -y install php7.3-fpm php7.3-cli
    ```
 
    {:.bs-callout-info}
-   This command installs the latest available version of PHP 7.2.X. See [Magento 2.3.x technology stack requirements]({{ page.baseurl }}/install-gde/system-requirements-tech.html) for supported PHP versions.
+   This command installs the latest available version of PHP 7.3.X. See [Magento 2.4.x technology stack requirements]({{ page.baseurl }}/install-gde/system-requirements-tech.html) for supported PHP versions.
```
jeff-matthews, line 51: I don't think we need to include the version of Magento in this line. Especially since the link to the tech stack page uses the page.baseurl var. It's designed to go to the same page whether you're in 2.3.x or 2.4.x docs.

dobooth: Righto

src/guides/v2.3/install-gde/prereq/merch_install.md
dobooth: Removed/

src/guides/v2.3/install-gde/prereq/merch_install.md
dobooth: Good point.

src/_includes/install/php-versions-template.md
```diff
@@ -2,10 +2,6 @@ Supported PHP versions:
 
 {% for version in supported_php_versions %}
```
jeff-matthews, line 3: Shouldn't we add a new composer.lock file for 2.4 in this PR? I know it won't be the stable release version, but at least it will generate the correct version of PHP for 2.4 at beta.

dobooth:
running tests

hunt0r:
I feel silly posting this here, so please delete if it's inappropriate.
@jeff-matthews your post above references 'the notorious "Line must be part of the diff" known issue'. I have encountered that issue, but unable to find it with Google/Github support searching. My Google-searching led me here :laughing:   Would you mind sharing a link about it?

jeff-matthews:
Hi @hunt0r! I wish I could share a link, but I was actually referring to this being "notorious" for those of us working on this particular project. I think everyone on the `devdocs-admins` team has encountered it at some point.

All we really know is that it has something to do with files that were previously symlinked in our project. For example, after deleting the symlink for `src/guides/v2.4/install-gde/prereq/php-settings.md`, creating a new/"real" file, and adding/pushing, the GitHub interface throws that error if you try to leave inline comments during review.

We haven't been able to track down any official root cause. So if you find something, please come back and let us know!

hunt0r:
That is exactly my case. Unfortunately, it is on a private repo with proprietary code so I can't provide an example.
We had one "file" which was merely a symlink to another.
I deleted the symlink and then copied another file into the same place. Git seemed to handle this fine.
I pushed to github, and then went to write a comment on one line of the file. But got that error, and could not post my comment.

Thanks for the response. Since I can't provide an example, I am hesitant to open a Github support ticket. But if you are willing to open one, please let me know if I can help in some way.

PR #7262, opened 2020-05-21
2fa updates

ccondra:

This pull request (PR) is to update the Two-Factor Authentication topic for the 2.4 beta release.

- 2FA is now enabled by default, and cannot be disabled.
- Removed references and instructions for resetting or disabling 2FA
- "Trust this device" functionality removed.
- Added a diagram of Magento Web API Flow

## Affected DevDocs pages

[Two-Factor Authentication 
](https://devdocs.magento.com/guides/v2.3/security/two-factor-authentication.html)

whatsnew
Removed the option to disable 2FA from the [Two-Factor Authentication 
](https://devdocs.magento.com/guides/v2.4/security/two-factor-authentication.html) topic and added a new section describing the Magento Web API workflow.


ccondra:
running tests

PR #7311, opened 2020-05-29
Fix Composer syntax for installing upgrade plugin

jeff-matthews:

This pull request (PR) fixes the Composer syntax in the CLI upgrade topic for installing the upgrade plugin. 
The composer plugin command needs to be `composer require magento/composer-root-update-plugin=~1.0 --no-update`  (with the equals sign); apparently not using the equals sign causes the command to fail when using zsh instead of bash

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.4/comp-mgr/cli/cli-upgrade.html
- https://devdocs.magento.com/guides/v2.3/comp-mgr/cli/upgrade-with-plugin.html


jeff-matthews:
running tests

PR #7314, opened 2020-05-29
magento/devdocs#: Disabling plugins. Add an example about a particular case.

atwixfirster:

This pull request (PR) adds an example how to correctly disable a plugin which specified with a backslash:

```
<type name="\Magento\Catalog\Controller\Adminhtml\Product\Action\Attribute\Save">
    ....
``` 

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/extension-dev-guide/plugins.html

-  ...


devops-devdocs:

src/guides/v2.3/extension-dev-guide/plugins.md
```diff
@@ -485,6 +485,33 @@ Plugins can be disabled in a `di.xml` file. To disable a plugin, set the `disabl
...
+
+then you should use the next approach
+
+```xml
+<type name="\Magento\Catalog\Controller\Adminhtml\Product\Action\Attribute\Save">
+    <plugin name="bindAttributeIndexerMassUpdate" disabled="true"/>
+</type>
+```
+
+The approach with `Magento\Catalog\Controller\Adminhtml\Product\Action\Attribute\Save` will not work:
```
rogyar, line 507: Please, add a minor clarification here:

```suggestion
The following example with no trailing backslash 
 `Magento\Catalog\Controller\Adminhtml\Product\Action\Attribute\Save` will not work:
```

atwixfirster: **implemented**

Thanks, @rogyar

src/guides/v2.3/extension-dev-guide/plugins.md
```diff
@@ -485,6 +485,33 @@ Plugins can be disabled in a `di.xml` file. To disable a plugin, set the `disabl
...
+<type name="\Magento\Catalog\Controller\Adminhtml\Product\Action\Attribute\Save">
+    <plugin name="bindAttributeIndexerMassUpdate"
+            type="\Magento\GoogleShoppingAds\Plugin\Catalog\Controller\Adminhtml\Product\Action\Attribute\Save\BindAttributeIndexer"/>
+</type>
+```
+
+then you should use the next approach
+
+```xml
+<type name="\Magento\Catalog\Controller\Adminhtml\Product\Action\Attribute\Save">
```
dobooth, line 502: Is there any difference between this example and the existing example above your change? You always have to use the same class name, no?

src/guides/v2.3/extension-dev-guide/plugins.md
```diff
@@ -485,6 +485,33 @@ Plugins can be disabled in a `di.xml` file. To disable a plugin, set the `disabl
...
+    <plugin name="bindAttributeIndexerMassUpdate" disabled="true"/>
+</type>
+```
+
+The following example with no trailing backslash `Magento\Catalog\Controller\Adminhtml\Product\Action\Attribute\Save` will not work:
```
dobooth, line 507: Do you mean the leading backslash?

src/guides/v2.3/extension-dev-guide/plugins.md
atwixfirster: Hi, @dobooth !

`\Magento\Catalog\Controller\Adminhtml\Product\Action\Attribute\Save` is not the same as `Magento\Catalog\Controller\Adminhtml\Product\Action\Attribute\Save`.

It is the same if you compare "Donald" and "Alex".

src/guides/v2.3/extension-dev-guide/plugins.md
atwixfirster: > Do you mean the leading backslash?

yep :)

src/guides/v2.3/extension-dev-guide/plugins.md
dobooth: The example in L480 says to add the `disabled='true'` attribute to disable the plugin. It seems that the example here on L502 is doing the same thing.  If there is no difference in what we are showing, there is no need to duplicate the example. And we can probably just make a note about the importance of the leading slash. A code example of a leading slash is likely overkill.

atwixfirster: @dobooth , let me please clarify.

The idea of my PR is: show developer a case when a plugin can be set on a particular class with a leading `backslash`.

In the example I've used a real code of real module.

For example we have a class `Magento\Catalog\Controller\Adminhtml\Product\Action\Attribute\Save`.

Developer can define a plugin for this class as:

```
<type name="Magento\Catalog\Controller\Adminhtml\Product\Action\Attribute\Save">
  <plugin name="plugin_a" ....>
....
```

or 

```
<type name="\Magento\Catalog\Controller\Adminhtml\Product\Action\Attribute\Save">
  <plugin name="plugin_b" ....>
....
```

So, if you want to disable `plugin_a` you have to call `Magento\Catalog\Controller\Adminhtml\Product\Action\Attribute\Save`. This is like an alias.

So, this is the important notice for developers which want to disable a plugin. They should refer to the original plugin declaration.

Thank you!

dobooth: I see. You can define a plugin two ways: with the leading slash or without. If you want to disable the plugin, you must use the same way to disable as it was defined.

atwixfirster: @dobooth , correct :)

dobooth: Good. Let me know if my changes are clear and correct.

atwixfirster: @dobooth , regarding to a PR purpose your changes are valid.

**Approved**

Thank you ;)

dobooth:
running tests

PR #7330, opened 2020-06-01
Update L2 cache

vzabaznov:
Update L2 cache with a new option that was introduced in 2.4.0

Represent new cache approach called 'stale cache' to community

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.4/config-guide/cache/two-level-cache.html

whatsnew
Added the 'Stale cache options' section to [L2 caching in the Magento application][(https://devdocs.magento.com/guides/v2.4/config-guide/cache/two-level-cache.html).

### DIFF
--- a/src/guides/v2.4/config-guide/cache/two-level-cache.md
+++ b/src/guides/v2.4/config-guide/cache/two-level-cache.md
@@ -7,14 +7,14 @@ functional_areas:
   - Cache
 ---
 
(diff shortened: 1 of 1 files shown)


devops-devdocs:

dobooth:
Hi @vzabaznov Thanks for the excellent contribution! One thing before we review it: The original file is a symlink. It needs to be deleted via `git rm` and then copy the v2.3 file to the 2.4 location or create a new topic, depending on your needs. If we overwrite a symlink like this, the build will fail. Thanks!

vzabaznov:
@dobooth Done

dobooth:
running tests

PR #7382, opened 2020-06-09
Updated docfile.yml to point to separate PB docs repo

bdenham:

This pull request (PR) is to point the devdocs pipeline to a docs-only Page Builder repo. We on the Page Builder team want to separate the docs from the code base for more flexibility and this is our first step.

Successful staging preview: https://devdocs.magedevteam.com/1916/page-builder/docs/

### DIFF
--- a/Docfile.yml
+++ b/Docfile.yml
@@ -16,8 +16,8 @@ content_map:
   filter: true
 -
   directory: src/page-builder
-  repository: magento/magento2-page-builder
-  branch: develop
+  repository: magento-devdocs/page-builder
+  branch: master
   filter: true
 -
   directory: src/page-builder-migration


jeff-matthews:
Looks like the rake init reads from the docfile.yml config, so maybe we don't need rake updates. Will wait for @dshevtsov to confirm.

dshevtsov:
I don't see any strong reason to create a separate repository for the docs, if developers don't want them in their codebase. Following Occam's razor principle, it would make sense rather move the docs back to the devdocs repository instead of multiplying entities without a strong reason.

bdenham:
I want the ability to add and experiment with different tooling within this separate repo. And my team lead wants the same. I think that is strong reason. Using Occam’s razor to justify your point of view on something might seem wise, but maybe it isn’t. I prefer to take a scientific approach: **1. Create a hypothesis** - separating the docs from the code will prove more productive and flexible, which will lead to a better docs experience, **2. Test hypothesis, 3. Revise hypothesis. 4. Rinse and repeat.**  It may be that in the long run we find that it works better to keep the docs with the code repo. But with your approach, we’d never know for sure.

dshevtsov:
> I want the ability to add and experiment with different tooling within this separate repo. And my team lead wants the same. I think that is a strong reason.

This is sufficient reason.

jeff-matthews:
@bdenham, please post an announcement in our internal Slack channel before merging to give writers a heads up **before** they need to reinitialize.

PR #7384, opened 2020-06-09
Removed pre-release date for 2.3.5-p2

jeff-matthews:

This pull request (PR) removes the pre-release date for the 2.3.5-p2 security-only patch from the release schedule. There will be no pre-release.

## Affected DevDocs pages

-  https://devdocs.magento.com/release/


jeff-matthews:
running tests

PR #7398, opened 2020-06-12
Since the sockets php extension becomes required from 2.4.0 onwards, …

hostep:
…we can remove this warning.

This pull request (PR) prepares the documentation for the `sockets` php extension which becomes required in Magento 2.4.0 and up.
It removes this temporary warning and `ext-sockets` should get added to the existing list of extensions. However, this doesn't seem to happen automatically right now, see below under further discussion.

This is a follow up on https://github.com/magento/devdocs/pull/6922

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.4/install-gde/system-requirements-tech.html

See https://github.com/magento/magento2/issues/27200 & https://github.com/magento/magento2/pull/28585#issuecomment-643337713 which is in progress of being merged in 2.4.0

## Further discussion

@dobooth: from what I understand, the following code should generate the required php extensions automatically, right?
```md
<!--{% assign platform-req = site.data.codebase.v2_4.open-source.composer_lock.platform %}-->
{% include install/php-extensions-template.md %}
```

But somehow that's not happening yet, even though the `composer.lock` file in the codebase does show a requirement on the `sockets` extension: https://github.com/magento/devdocs/blob/2fa5ff1b38d306371a041907e201ef6725cfa3bd/src/_data/codebase/v2_4/open-source/composer_lock.json#L3769

Any idea why this isn't being picked up?

----

Also: I'm explicitly not putting this in the 2.3 docs, because in Magento 2.3.5 for example, even though the `composer.lock` file mentions it: https://github.com/magento/magento2/blob/2.3.5/composer.lock#L3758, it's still possible to install Magento without it, because of the version constraint on the `php-amqplib/php-amqplib` module still allows version 2.7.x which doesn't require the `sockets` extension: 
https://github.com/magento/magento2/blob/2.3.5/composer.json#L74
And when setting up a Magento 2.3.5 project without the `sockets` extension, composer will be smart enough to downgrade `php-amqplib/php-amqplib` to 2.7.x

----

@jfrontain: I think we should also mention the requirement of this `sockets` php extension in the release notes for 2.4.0, just FYI


devops-devdocs:

dobooth:
Hi @hostep Thanks for the help. We manually copy the composer.json file to the devdocs repo. When we release the final docs, we grab the final composer.lock file from the main magento repo and copy it over, usually the day before release. So it may not be completely up to date in the beta docs. But in this case, the PHP extensions listed in the sys reqs file coms from the `platform` node: https://github.com/magento/devdocs/blob/2fa5ff1b38d306371a041907e201ef6725cfa3bd/src/_data/codebase/v2_4/open-source/composer_lock.json#L10833
If the extension also needs to be listed here, we will have to work that out with the engineering team.

hostep:
Thanks for the info @dobooth !

Hmm, this means we do have to put it in the [`composer.json` file](https://github.com/magento/magento2/blob/3a65cdce2cac30767677c55ff932904dee9a86b4/composer.json#L15-L29) as well, because since `ext-sockets` is introduced by a dependency, it won't be written to the `platform` section I believe?
But that's not very maintainable since new requirements might pop up by just updating a dependency and it would be nice if the devdocs could detect those automatically somehow?

Here's what *could* work, but it's a bit involved and all in theory, I haven't tested this:
1) Use the `composer.json` & `composer.lock` file of a specific Magento version
2) Run `composer update --no-dev`
3) From the resulting `composer.lock` file, search for all `ext-*` entries in the `require` sections and keep them in memory
4) Run `composer update --prefer-lowest --no-dev`
5) From the resulting `composer.lock` file, search for all `ext-*` entries in the `require` sections and keep them in memory
4) From the list of step 3) & 5), only take the `ext-*` entries which occur in both lists, now you have the bare minimum extensions which are required for a Magento installation.

Just an idea, not sure if this will be super accurate, but it's the best I could come up with right now.

dobooth:
Thanks @hostep for the excellent feedback. We are going to discuss this internally, and figure out if we are missing something here and what the right fix is. Thank for the good info.

dshevtsov:
> We manually copy the composer.json file 
> usually the day before release

We do this automatically at the day of release.

dshevtsov:
We have a task (DOC-80) in our backlog to start using `composer.json` from a composer project (such as `magento/project-community-edition`) instead of `composer.lock` to get information about metapackages and PHP requirements there. Seems like we need to extend this task to start generating data we need instead of raw composer.json/.lock files. We definitely need to involve engineering in this to consider all the details.

Maybe @fascinosum has something to add here.

dshevtsov:
This was last resolution after a meeting with engineering (on Feb 4) from the mentioned above ticket:

1. PHP requirements (versions and extensions): from GitHub's `composer.json`
1. VBEs versions from the `composer.lock` generated after `composer create-project ... ` installation. The versions can be taken from `magento/product-community-edition` dependencies. PHP extensions can be taken from this `composer.lock` too.

dobooth:
running tests

hostep:
Thanks @dobooth!

I'd like to emphasise to gather the php extensions twice, one time after running `composer update` and looking at the resulting `composer.lock` and one time after running `composer update --prefer-lowest`, also looking at the new resulting `composer.lock` file and then only taking the found extensions which appear in both files.

If you would not do it this way, it would mark `ext-sockets` as required for Magento 2.3.5, while it isn't.
If you would do it this way, it will be more accurate.

Just FYI.

PR #7402, opened 2020-06-13
Updated inter server communication section

fruit-ninja:

This is initially confusing because users would assume that they can skip this step if they had their database and apache on the same server. While this did allow the user to complete the installation process, it still prevented curl being used by any of the necessary integrations such as payment and shipping. This update will allow users to make a more accurate decision at this point.

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.3/install-gde/prereq/security.html

### DIFF
--- a/src/guides/v2.3/install-gde/prereq/security.md
+++ b/src/guides/v2.3/install-gde/prereq/security.md
@@ -48,7 +48,7 @@ The preceding commands work only with the Apache web server. Because of the vari
 
 ### Enable inter-server communication
 
-If Apache and the database server are on the same host, you can skip this section and continue with [Opening Ports In Your Firewall](#install-iptables).
+If Apache and the database server are on the same host, you will still need to use the following command if you are planning to use integrations that use curl (ex. Paypal and USPS).
 
 To enable Apache to initiate a connection to another host with SELinux enabled:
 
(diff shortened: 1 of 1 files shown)


devops-devdocs:

src/guides/v2.3/install-gde/prereq/security.md
```diff
@@ -48,7 +48,7 @@ The preceding commands work only with the Apache web server. Because of the vari
 
 ### Enable inter-server communication
 
-If Apache and the database server are on the same host, you can skip this section and continue with [Opening Ports In Your Firewall](#install-iptables).
+If Apache and the database server are on the same host, you will still need to use the following command if you are planning to use integrations that use curl (ex. Paypal and USPS).
```
rogyar, line 51: I would say 

```suggestion
If Apache and the database server are on the same host, you may still need to use the following command if you are planning to use integrations that use curl (ex. Paypal and USPS).
```

The reason is, it's not basically Apache that uses curl. PHP does not require Apache to run curl commands. But with the more or less standard apache installation, the curl command will be executed via the Apache user/group. If your installation has different user/group/security settings you don't need to enable the SELinux command as described. 

Thank you

dobooth:
running tests

PR #7416, opened 2020-06-16
GraphQL: Add id_v2 for gift cards

keharper:

This pull request (PR) adds the id_v2 attribute to the GiftCardAmounts object, per https://github.com/magento/partners-magento2ee/pull/263

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.4/graphql/product/gift-card-product.html

-  ...

### DIFF
--- a/src/guides/v2.4/graphql/product/gift-card-product.md
+++ b/src/guides/v2.4/graphql/product/gift-card-product.md
@@ -1,1 +1,83 @@
-../../../v2.3/graphql/product/gift-card-product.md
+---
+group: graphql
+title: Gift card product data types
+ee_only: True
+redirect_from:
+  - /guides/v2.3/graphql/reference/gift-card-product.html
+---
+
+The `GiftCardProduct` data type defines properties of a gift card, including the minimum and maximum values and an array that contains the current and past values on the specific gift card
+
+It implements the following interfaces:
+
+-  `ProductInterface`
+-  `PhysicalProductInterface`
+-  `CustomizableProductInterface`
+
+## GiftCardProduct object
+
+The `GiftCardProduct` object contains the following attributes:
+
+Attribute | Type | Description
+--- | --- | ---
+`allow_message` | Boolean | Indicates whether the customer can provide a message to accompany the gift card
+`allow_open_amount` | Boolean | Indicates whether customers have the ability to set the value of the gift card
+`giftcard_amounts` | [`GiftCardAmounts`] | An array that contains information about the values and ID of a gift card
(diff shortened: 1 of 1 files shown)


keharper:
I broke symlinks in this PR, making it appear as if all the content is new. My changes are limited to:

- Removing the redirect statement in the frontmatter
- Adding `uid` attributes in the appropriate tables
- Updating the examples to include the `uid` attribute. The output was copy and pasted--it's not fudged.

Note: I didn't previously have an example showing a customizable option, and so I didn't create one to show the `uid` attribute.

rogyar:
Hi @keharper. I believe it's not a big deal since it's just an example. But the `uid` is not the same for different option values (same as `id`). We may change some part of the string randomly to make the value different for all options. 

https://github.com/magento/devdocs/pull/7416/files#diff-445a68c46a7c2e44ad0317182ba7c305R156

keharper:
@rogyar Is there anything I need to change with respect to the `uid` fields?

Also, I forgot to mention that I deliberately did not document the GiftCardCartItem data type in this PR. I need to create a new section in the left navigation of devdocs about interfaces and their implementations. I'll do that all at once in a separate PR.

rogyar:
Hi @keharper. I would suggest randomizing `uid` values for different option values so we have the `uid` values different

```graphql
"options": [
                {
                  "id": 1,
                  "uid": "YnVuZGxlLzEvMS8x",
                 ...
                },
                {
                  "id": 2,
                  "uid": "YnVuZGxlLzEuiBy7",
                  ...
                },
                {
                  "id": 3,
                  "uid": "YnVuZGRlLzbuiBym",
                  ...
                }
              ]
```

Something like that. [Here's the link](https://github.com/magento/devdocs/pull/7416/files#diff-445a68c46a7c2e44ad0317182ba7c305R156) to the mentioned example. Sorry, I cannot use inline suggestions for this PR for some reason.

keharper:
@rogyar The `uid` values are different. The first 10 characters all start with `YnVuZGxlLz` and end with `8x`.

Statis ball uids. Only the character before the 8 differs
"uid": "YnVuZGxlLz**EvMi**8x",
"uid": "YnVuZGxlLz**EvMS**8x",
"uid": "YnVuZGxlLz**EvMy**8x",

Yoga brick uid
"uid": "YnVuZGxlLz**IvNC**8x",

Straps uids. Again, only the character before the 8 differs
"uid": "YnVuZGxlLz**MvNi**8x",
"uid": "YnVuZGxlLz**MvNS**8x",
"uid": "YnVuZGxlLz**MvNy8**x",

Roller uid
"uid": "YnVuZGxlLz**QvOC**8x",

PR #7440, opened 2020-06-23
Update REST table of contents

jeff-matthews:

This pull request (PR) replaces redoc links to the Synchronous REST endpoints in the table of contents with links to Redocly hosted docs.

We've reported the inability to use "anchor" links to the different schemas to Redocly. If/when they fix that, we can update these links accordingly. Or, we can leave these as-is until a fix is in place and rely on redirects in the meantime (already implemented in another PR).

## Affected DevDocs pages

- http://devdocs.magento.com/guides/v2.3/rest/bk-rest.html

### DIFF
--- a/src/_data/toc/rest-api.yml
+++ b/src/_data/toc/rest-api.yml
@@ -12,15 +12,15 @@ pages:
           children:
 
             - label: Admin REST endpoints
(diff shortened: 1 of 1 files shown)


jeff-matthews:
running tests

PR #7469, opened 2020-06-28
Declaring custom UI component updated

aliejaz55:
the line 99 <field name="sendemail_store_id" formElement="my_select"> contains invalid formElement type.

Custom Ui Component contains invalid formElement type.  Which will possible produced the error.  it should be select instead my_select.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/ui_comp_guide/howto/new_component_declaration.html#example-creating-a-custom-secondary-not-basic-component


devops-devdocs:

PR #7483, opened 2020-07-04
Alter the default value of `--consumers-wait-for-messages`

AntonEvers:

`\Magento\Framework\MessageQueue\CallbackInvoker::isWaitingNextMessage` shows a default value of `1` instead of the mentioned `0`. This also makes sense. Otherwise starting from the introduction of this feature anyone who upgrades would be facing queue handlers that have stopped all of a sudden instead of endlessly waiting queue handlers like they were used to.

## Affected DevDocs pages

-  ...

-  \Magento\Framework\MessageQueue\CallbackInvoker::isWaitingNextMessage


devops-devdocs:

rogyar:
Reference from the source code where the default value set to 1 when there’s no configuration value defined

https://github.com/magento/magento2/blob/735579d8aa14f7275d92818848f4a405191e7cb2/lib/internal/Magento/Framework/MessageQueue/CallbackInvoker.php#L91

PR #7510, opened 2020-07-13
Update to Whatsnew since July 6

hguthrie:
Update the whatsnew file for this Monday. Looks like 3 entries.
Just keeping the maintenance low.

Preview:

Screen Shot 2020-07-13 at 09 51 41


hguthrie:
running tests

PR #7515, opened 2020-07-14
Update message-queues.md

nuzil:
Updated and fixed example of redefining topic adapter.

Original example was wrongly written. I will throw fatal error directly on applying and also I added other require params. Without them it will not work at all.

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.4/extension-dev-guide/message-queues/message-queues.html


devops-devdocs:

keharper:
running tests

PR #7536, opened 2020-07-17
Break symlinks before replacing instances of the term `shipping method`

keharper:

This pull request (PR) breaks symlinks in 2.4 files that contain the term "shipping method". This term is to be replaced with "delivery method" in a subsequent PR. I also updated the navigation in the GraphQL TOC.

## Affected DevDocs pages

-  ...

-  ...


keharper:
running tests

PR #7574, opened 2020-07-22
Break symlinks in files that mention the admin token endpoint 

keharper:

This pull request (PR) breaks the symlinks in files that mention the POST /V1/integration/admin/token endpoint. Since 2FA is now required, that endpoint is not the correct one. The technical updates will be in a subsequent PR.

## Affected DevDocs pages

-  ...

-  ...


keharper:
running tests

PR #7591, opened 2020-07-25
Versionless Contribution Guide

dshevtsov:

This pull request (PR) moves the Contribution Guide topics from `guides/v2.x` at the root of the site, since no difference is expected between versions of this guide.

Also, this adds site variables to indicate current (latest) version: `site.gdeurl` and `site.githuburl ` that is useful when linking to versioned pages in non-versioned topics.

## Affected DevDocs pages

-  all topics from _/guides/v2.3/contributor-guide/_ and _/guides/v2.4/contributor-guide/_ were moved to the _/contributor-guide/_


dshevtsov:
running tests

PR #7593, opened 2020-07-26
Update custom js content

mrtuvn:

This pull request (PR) update content for this topic
This content is make misunderstand for use. We should update it to avoid confusing for reader/devs

Recall this issue https://github.com/magento/magento2/issues/24830

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.3/javascript-dev-guide/javascript/custom_js.html

## Related docs
https://devdocs.magento.com/guides/v2.3/javascript-dev-guide/javascript/requirejs.html#requirejs-config-deps

-  ...


devops-devdocs:

rogyar:
I'm not sure what is this section about but the `deps` part of `requirejs-config` is being merged across all modules/themes. So, technically, we cannot override something in that way. Removing it might be a good idea but maybe there's something behind the scenes that the original author wanted to say. Still not sure.

mrtuvn:
@dshevtsov @rafaelstz Can you guys give me some comments

mrtuvn:
@guz-anton What do you think about this update

PR #7611, opened 2020-07-28
DOC-202: Revise MSI modules to be disabled list

shrielenee:

This pull request (PR) updates the list of MSI modules in the [Install Inventory Management topic](https://devdocs.magento.com/extensions/inventory-management/) that must be disabled to use Connector and OMS together.

Preview build # 2100

## Affected DevDocs pages

-  https://devdocs.magento.com/extensions/inventory-management/

-  ...


keharper:
This topic is very out of date, but it is beyond the scope of this PR to fix it. Here's what I'd like to see happen:

**In this PR**

* Truncate the list of modules so that you're only listing a few. Something like this:

```
'Magento_Inventory' => 0,
'Magento_InventoryAdminUi' => 0,
'Magento_InventoryAdvancedCheckout' => 0,
...
```
You could add the last module from the current list, but I'm concerned about the maintenance of the snippet if new modules are adder later.

* The example at the top should show all 1s instead of 0s.

* Update all links to v2.4.

* @JacobBrownAustin verify whether the last sentence in current line 87 is true: "We plan to support this integration in a later release."

**In another PR**

I'd like @sidolov to thoroughly review this topic and submit updates in a new PR. I don't know if it's possible for anyone to upgrade from a version prior to 2.3.

jeff-matthews:
I agree with @keharper regarding the long list of modules. While I initially advised @shrielenee to update the entire list, that's not as practical, so... I owe you a beer @shrielenee 😬 .

keharper:
@sidolov Is there any particular order the modules need to be disabled?

sidolov:
@keharper in case with config.php you don't need to follow the sequence of the modules, they will be disabled in the right way. 
In case with CLI command that provides the ability to disable the module - it will return the suggestion with modules dependencies that should be removed before.

shrielenee:
> This topic is very out of date, but it is beyond the scope of this PR to fix it. Here's what I'd like to see happen:
> 
> **In this PR**
> 
> Let's get out of the business of maintaining a list of an increasing number modules. Instruct the reader to run the `bin/magento module:status --enabled | grep "Magento_Inventory"` command to return a list of Inventory modules. Then run the `bin/magento module:disable <module_list>` command to disable each of these modules. I would have the example command list three modules, but make it clear the reader has to continue the process.
> 
> Confirm the results by showing the first few Inventory modules.
> 
> ```
> 'Magento_Inventory' => 0,
> 'Magento_InventoryAdminUi' => 0,
> 'Magento_InventoryAdvancedCheckout' => 0,
> ...
> ```
> 
> * Update all links to v2.4.
> * @JacobBrownAustin verify whether the last sentence in current line 87 is true: "We plan to support this integration in a later release."
> 
> **In another PR**
> 
> I'd like @sidolov to thoroughly review this topic and submit updates in a new PR. I don't know if it's possible for anyone to upgrade from a version prior to 2.3.

So, part of what you're suggesting is linked in the "To disable Inventory Management, see the instructions for Enable or disable modules." text just above the modules list. https://devdocs.magento.com/guides/v2.3/install-gde/install/cli/install-cli-subcommands-enable.html. It sounds like in this instance, in the topic for this PR, we just need to add the suggested step of:

"Instruct the reader to run the `bin/magento module:status --enabled | grep "Magento_Inventory"` command to return a list of Inventory modules."

I will also potentially bone up the other topic for disabling modules.

shrielenee:
I have revised the instructions to point to existing instructions (so as not to replicate). I also changed 2.3 links to 2.4.

@JacobBrownAustin can you verify whether the last sentence in current line 87 is true: "We plan to support this integration in a later release." ?

JacobBrownAustin:
If we tell them to disable all modules that start with "Magento_Inventory",  but they have connector installed,  this will disabled
Magento_InventoryMessageBus, which is a connector module,  and then connector will no longer be able to process the stock update messages from OMS.
We should add a note about that in this document.  Something like,
"If you have MCOM connector modules installed,  make sure not to disable Magento_InventoryMessageBus because it is a connector module,  not an MSI module".

JacobBrownAustin:
About  "We plan to support this integration in a later release."  , 

That is still one of our long term goals.  We haven't been working on it recently,  but now that we have finished with our 4.0.0 release,  we are going to start investigating it again.
Also,  now that we've dropped support for Magento 2.2,  and are supporting Magento 2.4,  it makes it even more important than before.  (Magento 2.2 didn't have MSI modules)

shrielenee:
Got it @JacobBrownAustin --- Added this 
Screen Shot 2020-07-30 at 9 14 12 AM

shrielenee:
running tests

meker12:
@shrielenee  Noticed this as missing whatsnew in PR description.

shrielenee:
Thanks @meker12 --- Added.

PR #7619, opened 2020-07-29
Note that security releases are not always -p1

navarr:

We add a single word here to modify the semantics of the statement.  It might otherwise imply that security releases are always and only `-p1`, which is incorrect as we have had two -p2 releases so far.

## Affected DevDocs pages

- https://devdocs.magento.com/release/index.html


devops-devdocs:

rogyar:
Hi @navarr. What do you think about extending this statement and making it more explicit? I see it as something like the following 

> These releases are typically appended with `-pN`, where N is the patch incremental number (example: 2.3.5-p1, 2.3.5-p2).

hostep:
Can we also instead of calling these `Security-only releases`, call them `Hotfix-only releases` ?
We've seen instances in the past where -pX releases did not contain any security fix at all, but just a high priority bugfix (2.3.2-p2, 2.3.4-p2 and 2.3.5-p1 come to mind)

coreydulimba-ad:
Security releases do contain hot fixes since Magento deems them critical to run Magento. suggest blurb in **bold** is added

Security-only releases provide fixes for vulnerabilities that have been identified in previous quarterly patch releases. You can install time-sensitive security fixes without applying the hundreds of functional fixes and enhancements that a full quarterly patch release contains. **These releases also include quality hot fixes since Magento has determined they are critical to run Magento effectively.**

jeff-matthews:
@navarr, will you please add that sentence?

navarr:
I can tell you I was _not_ expecting this PR to blow up like this.

I've added a line in bold that I think reads slightly better than the initial suggestion.

src/release/index.md
```diff
@@ -24,7 +24,7 @@ _<sup>*</sup>There will not be a 2.4.5 release in 2021 Q4._
 
 ## About security-only releases
 
-Security-only releases provide fixes for vulnerabilities that have been identified in previous quarterly patch releases. You can install time-sensitive security fixes without applying the hundreds of functional fixes and enhancements that a full quarterly patch release contains. These releases are appended with `-p1`.
+Security-only releases provide fixes for vulnerabilities that have been identified in previous quarterly patch releases. You can install time-sensitive security fixes without applying the hundreds of functional fixes and enhancements that a full quarterly patch release contains. These releases are appended with `-pN`, where N is the incremental patch version beginning with 1 (example: 2.3.5-p1, 2.3.5-p2). **These releases may also include hotfixes determined to be crticial for the operation of a Magento application.**
```
meker12, line 27:
```suggestion
Security-only releases provide fixes for vulnerabilities that have been identified in previous quarterly patch releases. You can install time-sensitive security fixes without applying the hundreds of functional fixes and enhancements that a full quarterly patch release contains. These releases are appended with `-pN`, where N is the incremental patch version beginning with 1 (example: 2.3.5-p1, 2.3.5-p2). **These releases can also include hotfixes required to address critical issues that affect the Magento application.**
```

meker12:
running tests

PR #7628, opened 2020-07-29
Remove blank line from metadata

meker12:

Removed blank line from metadata that caused page to render incorrectly.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.4/security/security-txt.html


meker12:
running tests

meker12:
running tests

meker12:
running tests

meker12:
It builds and works locally. In my browser, it still says that the page was updated 36 hours ago, even though I refreshed. Am looking again.

jeff-matthews:
Ah, yes. I think you're right about the deployment option. You should select the `all` parameter on the devdocs-production pipeline.

PR #7631, opened 2020-07-30
Assign page.guide_version via configuration scopes

dshevtsov:

This pull request (PR) delegates generation of the `page.guide_version` parameter to native Jekyll means using configuration scopes, instead of expensive plugin with regex. This allows to reduce build time by ~1 min.

### DIFF
--- a/_plugins/page-params/page-baseurl-generator.rb
+++ b/_plugins/page-params/page-baseurl-generator.rb
@@ -15,20 +15,11 @@
 # if the path doesn't contain "guides/v2.x", then the version is unset and returns the nil object (same as null)
 Jekyll::Hooks.register :pages, :post_init do |page|
   # Glossary. Create and assign variables to be used in the script.
-  pattern = %r{\Aguides/v(?<version_from_path>2\.\d)}
   baseurl = page.site.baseurl
-  path = page.path
   data = page.data
   version = data['guide_version']
 
-  # Process 'page.guide_version' pararmeter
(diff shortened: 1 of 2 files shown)


_plugins/page-params/page-baseurl-generator.rb
```diff
@@ -6,29 +6,19 @@
 # This custom plugin dynamically sets and injects the page.baseurl variable
 # based on the page's destination.
 #
-# The hook introduces the page.baseurl and
-# page.guide_version data parameters for each page.
+# The hook introduces the page.baseurl parameter for each page.
 # For pages at guides/v2.x the page.baseurl parameter is set
 # as "{site.baseurl}/guides/v#{version}".
 # The {version} is taken from the 'guide_version' front matter parameter on the page;
```
jeff-matthews, line 12: Is it still possible to use `guide_version` in the metadata? If not, we should remove this comment.

dshevtsov: Yes, it's still possible. Front matter parameters always have privilege over default parameters assigned via scopes. (This is how we use layout at the moment.) For more information, refer to [Precedence in Jekyll](https://jekyllrb.com/docs/configuration/front-matter-defaults/#precedence).

dshevtsov:
This change breaks `include/exclude versions` feature in TOC.
Converting to draft until resolved.

dshevtsov:
running tests

PR #7615, opened 2020-07-29
Update security-txt.md

kalpmehta:
## Purpose of this pull request

Added contributor acknowledgement for contributing content to the  new `security.txt` file documentation.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.4/security/security-txt.html

PR #7634, opened 2020-07-30
Fix encoding issue on the Security.txt page

jeff-matthews:

This pull request (PR) fixes an encoding issue introduced in #7615 that is breaking the page.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.4/security/security-txt.html


meker12:
running tests

PR #7658, opened 2020-08-03
Consolidate info about security releases on the policy page

jeff-matthews:

This pull request (PR) consolidates info about security releases on the policy page.

## Affected DevDocs pages

- https://devdocs.magento.com/release/
- https://devdocs.magento.com/release/policy/


coreydulimba-ad:
thanks, Jeff! this looks good to me. not sure if the hot fix part needs to be bolded. I did that in the ticket just to clarify the addition but, i'll leave that up to you to keep / remove

coreydulimba-ad:
@jeff-matthews  oh and one other consideration should we add a link to the policy page? Something like "To read more about our release policy see here" if you think that is not needed then ignore

jeff-matthews:
Thanks @coreydulimba-ad. If we want to draw attention to that sentence, there's probably a better way to do it (e.g., note format). If not, then no bolding is necessary.

coreydulimba-ad:
@jeff-matthews i don't think we need to draw special attention to it, thanks!

smiverma:
@jeff-matthews and @coreydulimba-ad Do we want to want to change the line to "These releases can also include hotfixes (depending on when these issues are fixed) required to address critical issues that affect the Magento application", to consider the fact that only hotfixes done by code freeze will make it to the patch? Might be confusing but wanted to bring it up.

dobooth:
Is this one ready?

jeff-matthews:
Yes, I think so @dobooth. If @coreydulimba-ad and @smiverma want to add additional language later, they can open a PR.

PR #7678, opened 2020-08-05
Remove obsolete topics and refresh config

dshevtsov:

This pull request (PR) removes an old file (and a symlink) with no references in navigation:` src/guides/v2.3/extension-dev-guide/Contribute_edg.md`; and a file that duplicates information about contributing guidelines,`src/contributor-guide/contributing_docs_suggested.md`.

This also updates the site configuration by removing unused parameters, sorting some lists, and updating in-code documantation.

## Affected DevDocs pages

-  https://devdocs.magento.com/contributor-guide/contributing_docs_suggested.html


_config.yml
```diff
@@ -157,19 +158,16 @@ optional_front_matter:
 # Exclude all .md directories and files except page-builder/ to prevent painfully long build times.
 
 relative_links:
+  enabled:     true
+  collections: false
   exclude:
-    - _includes
```
jeff-matthews, line 161: I don't think we need to crawl the `_includes` directory for relative links. I think removing it from the exclude list will increase build time.

dshevtsov: > I don't think we need to crawl the _includes directory for relative links.

I don't think it's been crawled, considering the Jekyll architecture. The files in __includes_ do not contain publishable source files, only chunks of content that's been included in future topic at one of processing layers. I assume that it's been crawled at that layer, when the content was included.

> will increase build time.

I checked this, it didn't change the build time.

dshevtsov:
running tests

dshevtsov:
running tests

PR #7691, opened 2020-08-07
Use updated way on how echo is done in phtml

arnoudhgz:

Use updated way on how echo is done in phtml

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.4/frontend-dev-guide/translations/translate_theory.html


devops-devdocs:

rogyar:
Basically, both ways are correct. But the short echo parameter (introduced in the current changeset) is preferable in most cases.

dobooth:
That's a nice little feature!

PR #7697, opened 2020-08-07
Typing mistake

aliejaz55:
Corrected Typing mistake.

Typing mistake corrected

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.4/frontend-dev-guide/layouts/xml-manage.html


devops-devdocs:

rogyar:
Hi @aliejaz55. Thank you for your idea. 
"elements' order" means literally "orders of elements" (plural), and "element's order" means "order of element" (singular). But in this particular context, we are talking about sort order of one block. So, I find your edit totally correct.

PR #7731, opened 2020-08-12
GraphQL: Add statement about cached/uncached queries

### LINKED ISSUE #7718: Add note about mixing cacheable and noncacheable queries in a single request
# General issue

## Description:
It would be great if we could list that we have a bit of a caveat with GraphQL caching.

Unlike typical REST APIs, GraphQL will let you include as many distinct queries as you'd like in a single request.

But, with Magento's Full Page GraphQL cache, a single query that is not cacheable will cause _all queries in the same request to bypass cache_.

**Example that hits cache**
```graphql
{
   urlResolver(url: "foo.html") {
      type
      relative_url
   }
}
```

**Example that completely misses the cache**
```graphql
{
   urlResolver(url: "foo.html") {
      type
      relative_url
   }
   # storeConfig is a non-cacheable query, so it invalidates
   # all caching on this request
   storeConfig {
      copyright
   }
}
```

## Possible solutions:
Developers should become familiar with the [list of cacheable queries](https://devdocs.magento.com/guides/v2.4/graphql/caching.html#cached-queries), and separate requests for cacheable data from requests for non-cacheable data

keharper:

This pull request (PR) adds a statement about mixing cached and uncached queries in a single call.

Fixes #7718 

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/graphql/caching.html
- https://devdocs.magento.com/guides/v2.4/graphql/caching.html

-  ...

### DIFF
--- a/src/guides/v2.3/graphql/caching.md
+++ b/src/guides/v2.3/graphql/caching.md
@@ -5,9 +5,12 @@ title: GraphQL caching
 
 Magento can cache pages rendered from the results of certain GraphQL queries with [full-page caching]({{page.baseurl}}/extension-dev-guide/cache/page-caching.html). Full-page caching improves response time and reduces the load on the server. Without caching, each page might need to run blocks of code and retrieve large amounts of information from the database. Only queries submitted with an HTTP GET operation can be cached. POST queries cannot be cached.
 
-## Cached queries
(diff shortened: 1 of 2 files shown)


keharper:
running tests

PR #7782, opened 2020-08-27
Fix list of allowed attributes in escaper

dannynimmo:

Removed `target` as this attribute is not allowed by the escaper. List of allowed attributes can be found here: https://github.com/magento/magento2/blob/2.4-develop/lib/internal/Magento/Framework/Escaper.php#L45

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.4/extension-dev-guide/xss-protection.html

- https://github.com/magento/magento2/blob/2.4-develop/lib/internal/Magento/Framework/Escaper.php

### DIFF
--- a/src/guides/v2.3/extension-dev-guide/xss-protection.md
+++ b/src/guides/v2.3/extension-dev-guide/xss-protection.md
@@ -93,7 +93,7 @@ let settings = <?= $myJson ?>
 
 You can pass in an optional array of allowed tags that will not be escaped.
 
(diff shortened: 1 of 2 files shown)


devops-devdocs:

nuzil (changes requested):
Hi @dannynimmo can you please update also version for 2.3 I see there are same change

dannynimmo:
@nuzil sure, done

nuzil (approved):

dobooth:
running tests

PR #2186, opened 2018-06-08
Typo fixes and edits for Cloud docs

meker12:
## This PR is a:
- [X] Content fix or rewrite

## Summary
 
Fixes typos and language issues in Cloud docs.

PR #7796, opened 2020-09-01
2FA updated diagram

ccondra:
This pull request (PR) corrects the instruction in the diagram for Google Authenticate. Should be:

POST /tfa/provider/google/authenticate

{"username":"admin,"password":"mypassword","otp":"123456"}

## Affected DevDocs pages

[Two-Factor Authentication](https://devdocs.magento.com/guides/v2.4/security/two-factor-authentication.html)

See staging: #2186


ccondra:
running tests

PR #7809, opened 2020-09-02
Removing MFTF references

dobooth:

This pull request (PR) removes:
- https://devdocs.magento.com/guides/v2.4/reference/mftf/action-groups.html
- https://devdocs.magento.com/guides/v2.4/reference/mftf/functional-tests.html

by request of the MFTF team.


jeff-matthews (approved):

dshevtsov:
This reference documentation belongs to the magento2 codebases not to the MFTF codebase. Why would we reduce amount of content? Wouldn't it be better to move it to another location if these docs do not belong to the MFTF guide?

jeff-matthews:
@dobooth has thoroughly vetted the request from the MFTF team to remove these docs and I support the request. Especially since they receive almost no user traffic according to our analytics.

dshevtsov:
I do not support this.

jeff-matthews:
@dobooth, please provide the reasons that the MFTF team used to justify their request.

dshevtsov:
These topics are the only sources of information about out-of-the box Magento functional tests and action groups (building blocks of functional tests). Low traffic doesn't make it less valuable.

dshevtsov:
From what I figured from the discussion, the team wanted to remove https://devdocs.magento.com/mftf/docs/commands/codeception.html and https://devdocs.magento.com/mftf/docs/commands/mftf.html, because the corresponding commands were removed in latest MFTF version. And I have no objection to this.

jeff-matthews:
> Low traffic doesn't make it less valuable.

I disagree. And just because we _have the ability_ to publish certain types of reference info, that doesn't necessarily mean that we _should_. Especially when other experts offer opinions against it.

jeff-matthews:
I've said all I care to say about the subject. I trust the PR author has fulfilled the request per the team's discussion.

dobooth:
* It’s not useful in the current form as a mega long list.
* It’s not helpful to the test writing experience. We don’t think people would consult this list to make a decision which action group to use.
* It’s always unclear to us which branch/tag this is pointing at. Is it 2.4-develop? 2.4.0? 2.4.1? Will it be changed to point to 2.4.2 when it comes out? Etc. This is important because action groups and test change a lot. New ones are added, some are removed, etc.
* In general, I feel strongly it has caused more confusion and discussions than good.

dobooth:
running tests

dshevtsov:
> It’s not useful in the current form as a mega long list.

It's an issue of formatting.

> would consult this list to make a decision 

Its main goal is to provide a list of what is available.

> It’s always unclear to us which branch/tag this is pointing at.

This is a matter of how the topic is presented.

Main issue of the topics is that they need to be reformatted according to this feedback. I don't like the current format either. But it shouldn't be the reason to remove it, instead of making it more reader-friendly.

tomreece:
The pages in their current form are a more source of stress than benefit. We've had lots of discussion about these pages throughout their lifetime. Also you've had to implement, maintain, and run code on your end to generate them.

We concluded that we do not find these pages beneficial enough to warrant all the discussion and maintenance of the pages. I believe it's truly easier if they just go away until such time that we can present the information in a better way.

dshevtsov:
>in their current form

So this is an issue of formatting.

dobooth:
running tests

Issue #1786, opened 2018-03-30
Cloud: some documentation to explain we can modify php.ini file would be helpful


hostep:
Hi guys

This was discovered by talking to the Magento Cloud support, but apparently there is a file [`php.ini`](https://github.com/magento/magento-cloud/blob/master/php.ini) included in the Magento Cloud files, and this file allows us to change certain php values on the server (also on staging & production).
I can't seem to find anything about this in the documentation for the Cloud, so this would make a nice addition I think for other people looking for that info?
I think it makes sense to add it underneath the ['Configure environments'](http://devdocs.magento.com/guides/v2.2/cloud/env/environments.html) section?

Thanks!

Pieter

jeff-matthews:
Thanks @hostep! We'll add it to our queue.

hguthrie:
There is a section in the Cloud guide that mentions customizing `php.ini` settings. Ideally, we could extract that section to a new topic and expand on it. The section is in:
[Configure environments > Application > Configure PHP options](https://devdocs.magento.com/guides/v2.2/cloud/project/project-conf-files_magento-app.html#customize-phpini-settings)

hguthrie:
Adding the corresponding internal tracking: MCLOUD-1620

BarnyShergold:
@hguthrie @jeff-matthews I've just taken another look at this page : https://devdocs.magento.com/cloud/project/project-conf-files_magento-app.html 

It is much too long and contains information about a wide variety of topics. I would recommend this page be broken up in small pages and made sub-pages of Configure Environment > Application. To join "Set Cache For Static Files"

I would be happy to do this task if we can agree it should be done- this would make PHP.INI it's own page and menu point

hguthrie:
@BarnyShergold I completely agree with you. All the configuration information in that topic is applicable to the application configuration and breaking out the subjects makes sense. The internal tracking seems to have become technical debt, but the original intent was to do as you describe. Properties and Variables is enough for the one topic, but customizing PHP settings and understanding Workers should be separate topics under Application.

If you would like to take that on, it would help out immensely. 😄

BarnyShergold:
@hguthrie - I'll take this one

PR #7813, opened 2020-09-03
Refactor Cloud Application Page to smaller pages with Menu Entries

BarnyShergold:

This pull request refactors the Application Configuration page for Cloud as it is too big and it's not clear what is on the page

Deals with issue #1786 

## Affected DevDocs pages

https://devdocs.magento.com/cloud/project/project-conf-files_magento-app.html
-  ...


devops-devdocs:

BarnyShergold:
One of the files has been flagged as conflicting but I don't understand why - what's the best way to resolve on GitHub?

BarnyShergold:
@rogyar Actually this split was deliberate as it is so that PHP.INI can appear in the menu - which was the reason why the issues was raised. The complaint was that that the ability to modify PHP.INI was buried! So this was the purpose of splitting the two. Otherwise I would have put them on a single page. What do you think? If you have another suggestion so PHP.INI can stand out?

dmytro-ch (commented):
Hello @rogyar,
I agree with @BarnyShergold.

"Customize php.ini settings" is not part of Application configurations in terms of managing the `.magento.app.yaml`. So I'm not sure it should be even stored under the [Application](https://devdocs.magento.com/cloud/project/project-conf-files_magento-app.html) section at all.
Taking into account that each of these four sections (`Application`, `Environment variables`, `Routes`, and `Services`) represents a separate config type with its own config file, so it might be even worth to store "Customize php.ini settings" in somewhere else. E.g. directly under [Configure environments](https://devdocs.magento.com/cloud/env/environments.html) similarly as we currently have the [Restore an environment](https://devdocs.magento.com/cloud/env/restore-environment.html) there.

BarnyShergold:
@rogyar - On reflection I can see @dmytro-ch's suggestion is correct - we need to have each section relating to configuration file and whilst it's optional, php.ini is separate.

I propose leaving the PHP Application page as it is - it relates to .magento.app.yaml and add a new menu point between Services & Restore : Custom PHP.INI

I have a further suggestion to put to the discussion - add the configuration file name to the menu so people can see it. I might know about .magento app.yaml but where in the "Configure Environments" is it?

How about -
Application (.magento.app.yaml)
Build & Deploy (.magento.env.yaml)
Routes (routes.yaml)
Services (services.yaml)

Content remains the same.

Let's agree before I make further changes.

hguthrie:
The intent of the Configure environment section is to guide users on how to manage the YAML configuration files.
For me, it seems that Restore an environment really belongs in Managing your project.

Regarding the Custom PHP.INI, if it needs to stand alone from the `.magento.app.yaml` file, then that move makes sense. Let me take a little time to look at it closer thins morning and I will get back to you all. 🙂

hguthrie:
> Actually, I agree with that @hguthrie - looks like we've opened up a can of worms!
> 
> Can I propose that we even consider moving "Docker Development" to come under "Local Development" (and remove 'setup')? Someone new might think the Docker section applies to the Production/Staging/Integration environments.

Docker development used to be like that, but there is another project underway for Docker. Let's leave that alone for now.

BarnyShergold:
Yeah - another PR :) Otherwise it never ends

But let me know about the above and I'll do the rejigging

src/cloud/project/project-conf-files_magento-app.md
```diff
@@ -5,507 +5,16 @@ functional_areas:
...
-
--  Tasks like queue workers or updating indexes.
--  Tasks to run periodic reporting that are too long for a cron job.
--  Tasks should happen "now", but not block a web request.
--  Tasks are large enough that they risk blocking a deploy, even if they are subdivided.
--  The task in question is a continually running process rather than a stream of discrete units of work.
+### [Properties]({{ site.baseurl }}/cloud/project/project-conf-files_magento-app_properties.html)
 
-A basic, common worker configuration could look like this:
+### [Variables]({{ site.baseurl }}/cloud/project/project-conf-files_magento-app_variables.html)
```
hguthrie, line 14: With the above suggestion, you can remove all of these headings.

hguthrie (changes requested):
This is my first round of suggestions.

First: Filenames—The filename for the application topic (`project-conf-files_magento-app.md`) is from an older time and we no longer use this convention. Could you consider removing the `project-conf-files_` from each of the **NEW** filenames and avoid underscores? For example: `magento-app-workers.md`

Second: I made a healthy-sized suggestion for the Application topic, since it looks like it's now a landing page. The headings that are links could cause some weirdness in navigation. I suggested creating two clean sections. File structure section could be about the makeup or contents of the YAML file and the Customizations section links to the various customizations that one can do...maybe list could grow.

My descriptions are just filler and may not be entirely accurate. You can improve on those, if needed.

src/cloud/project/project-conf-files_magento-app.md
BarnyShergold: Okay I will look into those. It will take time to rename all the files as there are links to them everywhere!

hguthrie: Yes, sorry about that. Just don't want to continue that wonky file naming. I honestly have no idea what `project-conf-files_` meant. It's from some year ending in B.C. I think.

BarnyShergold: Before Cloud?

BarnyShergold: What I'll do it get the content right first (just in case we need new files) - then when you give a 'verbal' oaky - I'll refactor all the files name and then you can do a formal review - that okay?

hguthrie (commented):
I understand why you want to add the filenames in the left navigation, but there isn't much room to work with. We used to have the filenames there and removed them for that reason.

Lists need two spaces after the dash. The suggestions I made should fix the output.

src/cloud/project/project-conf-files_magento-app.md
```diff
@@ -11,8 +11,8 @@ The `.magento.app.yaml` has many default values, see [a sample `.magento.app.yam
...
-- [PHP Application]({{ site.baseurl }}/cloud/project/project-conf-files_magento-app_php-application.html)—Configure runtime PHP options.
-- [Workers]({{ site.baseurl }}/cloud/project/project-conf-files_magento-app_workers.html)—Create a worker instance and understand background tasks.
-- [Set Cache For Static Files]({{ site.baseurl }}/cloud/project/set-cache.html)—Set cache time-to-live (TTL) for your media and static files.
+-[Properties]({{ site.baseurl }}/cloud/project/project-conf-files_magento-app_properties.html)—Define property values for application instance.
+-[Variables]({{ site.baseurl }}/cloud/project/project-conf-files_magento-app_variables.html)—Review environment variables required for the application instance.
+-[PHP Application]({{ site.baseurl }}/cloud/project/project-conf-files_magento-app_php-application.html)—Configure runtime PHP options.
+-[Workers]({{ site.baseurl }}/cloud/project/project-conf-files_magento-app_workers.html)—Create a worker instance and understand background tasks.
+-[Set Cache For Static Files]({{ site.baseurl }}/cloud/project/set-cache.html)—Set cache time-to-live (TTL) for your media and static files.
```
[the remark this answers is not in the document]
BarnyShergold: I can take the file names out the list - just wanted us to see what it looked like
The reason I took the spaces out was that lint tests were complaining about them!

I'll make the updates and we can see.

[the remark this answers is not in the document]
BarnyShergold: @hguthrie - So apart from the proposed filename changes - how we looking for content?

BTW when I change the filenames, should I put this at the top of each and add the old file name :
redirect_from:
  - /cloud/deploy/configure-deploy.html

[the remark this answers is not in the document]
hguthrie: I don't think redirects are necessary in this case. I think that redirect was created when the config file topics split out...but that was long ago.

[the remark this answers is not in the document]
hguthrie: > I can take the file names out the list - just wanted us to see what it looked like
> The reason I took the spaces out was that lint tests were complaining about them!
> 
> I'll make the updates and we can see.

I think the order looks good. This will help to focus those topics for Workers and php.ini customizations.

hguthrie (commented):
Thank you! 🏆

PR #7788, opened 2020-08-31
Updated supported and recommended elasticsearch version

meker12:
## Purpose of this pull request

Updated the Elastiicsearch version compatibility and software version compatibility matrices to show Elasticsearch 7.7.x as the supported and recommended version for Magento Commerce on the Cloud platform.

## Affected DevDocs pages

- https://devdocs.magento.com/cloud/project/project-conf-files_services-elastic.html
- https://devdocs.magento.com/cloud/project/project-conf-files_services.html

whatsnew
Updated the Elastiicsearch version compatibility and software version compatibility matrices to show Elasticsearch 7.7.x as the supported and recommended version for Magento Commerce on the Cloud platform.

PR #7817, opened 2020-09-03
Fix Cloud services list alignment to correct broken list

hguthrie:
There is a spacing problem that affected a numbered list in the [Services topic](https://devdocs.magento.com/cloud/project/project-conf-files_services.html#downgrade-version).

### DIFF
--- a/src/cloud/project/project-conf-files_services.md
+++ b/src/cloud/project/project-conf-files_services.md
@@ -166,7 +166,7 @@ The following table lists the services used in {{site.data.var.ece}} and their v
 {:.error-table}
 Service   |  Magento 2.4  |Magento 2.3  | Magento 2.2
 --------- | ------------- |-------------| ------------
-`elasticsearch` | 7.7.x | **Magento version 2.3.5 and later**— 5.2, 6.5, 6.8, 7.5, 7.6, 7.7<br>**Magento version 2.3.1 to 2.3.4**— 5.2, 6.5<br>**Magento version 2.3.0**— 5.2  | **Magento version 2.2.8 and later**— 5.2, 6.5 <br>**Magento version 2.2.0 to 2.2.7**— 5.2
+`elasticsearch` | 7.7 | **Magento version 2.3.5 and later**— 5.2, 6.5, 6.8, 7.5, 7.6, 7.7<br>**Magento version 2.3.1 to 2.3.4**— 5.2, 6.5<br>**Magento version 2.3.0**— 5.2  | **Magento version 2.2.8 and later**— 5.2, 6.5 <br>**Magento version 2.2.0 to 2.2.7**— 5.2
 `mariadb` | 10.2, 10.3, 10.4 | **Magento version 2.3.0 to 2.3.5**–10.1 to 10.2<br> | 10.1 to 10.2
 `nginx`   | | 1.9           | 1.9
 `node`    | | 6, 8, 10, 11  | 6, 8, 10, 11
--- a/src/cloud/project/project-conf-files_services.md
+++ b/src/cloud/project/project-conf-files_services.md
@@ -282,12 +282,12 @@ To downgrade a service version by renaming an existing service:
        database: "mysql:mysql"
    ```
 
-  > Updated `.magento.app.yaml` configuration
+   > Updated `.magento.app.yaml` configuration
 
-  ```yaml
-  relationships:
-      database: "mysql2:mysql"
-  ```
+   ```yaml
+   relationships:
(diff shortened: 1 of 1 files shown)


meker12 (approved):

src/cloud/project/project-conf-files_services.md
```diff
@@ -166,7 +166,7 @@ The following table lists the services used in {{site.data.var.ece}} and their v
 {:.error-table}
 Service   |  Magento 2.4  |Magento 2.3  | Magento 2.2
 --------- | ------------- |-------------| ------------
-`elasticsearch` | 7.7.x | **Magento version 2.3.5 and later**— 5.2, 6.5, 6.8, 7.5, 7.6, 7.7<br>**Magento version 2.3.1 to 2.3.4**— 5.2, 6.5<br>**Magento version 2.3.0**— 5.2  | **Magento version 2.2.8 and later**— 5.2, 6.5 <br>**Magento version 2.2.0 to 2.2.7**— 5.2
+`elasticsearch` | 7.7 | **Magento version 2.3.5 and later**— 5.2, 6.5, 6.8, 7.5, 7.6, 7.7<br>**Magento version 2.3.1 to 2.3.4**— 5.2, 6.5<br>**Magento version 2.3.0**— 5.2  | **Magento version 2.2.8 and later**— 5.2, 6.5 <br>**Magento version 2.2.0 to 2.2.7**— 5.2
```
BarnyShergold, line 169: Going to have to assume that all the 2.3 and 2.2 numbers are correct

BUT

These documents -
https://devdocs.magento.com/guides/v2.4/install-gde/prereq/elasticsearch.html
https://devdocs.magento.com/guides/v2.4/release-notes/backward-incompatible-changes/index.html

State that the version of Elasticsearch for 2.4.0 is 7.6.x - which contradicts the 7.7 on this document

BarnyShergold (changes requested):

src/cloud/project/project-conf-files_services.md
hguthrie: @meker12  Don't get me in trouble. I only meant to fix some numbering. 😆  Maybe this is a good time to fix the ES versions? I don't want to go that far if you already have something staged?

BarnyShergold: I realise this is for Cloud but if the versions are different, should not mention be made of this to avoid possible confusion?

Eg the pages above say "You can use other versions at your discretion, but we recommend using the tested version of Elasticsearch." - so someone might question 7.7 if they have seen it?

BarnyShergold: > @meker12 Don't get me in trouble. I only meant to fix some numbering. 😆 Maybe this is a good time to fix the ES versions? I don't want to go that far if you already have something staged?

@hguthrie - I think some versions of services on Cloud are higher than lowest supported, just that as the pages say "You can use other....." - someone might be confused as to why the Cloud versions are higher. Perhaps a generic statement saying "Cloud uses higher versions of services than mentioned in the Standard Requirements" ?

BarnyShergold: I may be making a mountain out of a molehill but in my experience if there is a chance that someone can misinterpet documentation then they will. 

As Terry Pratchett said "Million-to-one chances crop up nine times out of ten"

hguthrie: I don't disagree with the logic, but the Cloud environment will only support the versions listed here. The table represents the compatibility for Cloud environment and sample Cloud instance. And Cloud usually uses "latest".

I can certainly review the consistency of the Elasticsearch version information for Cloud support of 2.4.x.

hguthrie:
There is no rush to push this through. Will review Elasticsearch and update accordingly. 🙂

src/cloud/project/project-conf-files_services.md
meker12: so specifically on Cloud, customers must use Elasticsearch 7.7 -- that's the version that is available on the Cloud infrastructure.  I just updated this based on guidance from Cloud engineering in PR --https://github.com/magento/devdocs/pull/7788

I agree that it would be helpful to add a statement that says the versions supported for Magento Commerce on Cloud platform might be different than Magento Commerce Enterprise or Open source.  How to say that gracefully?

BarnyShergold: Okay - agreed

BarnyShergold (approved):

hguthrie:
running tests

Issue #6332, opened 2020-01-03
Replace the section that instructs users to manually create a webhook with guidance about autogenerated webhook


jeff-matthews:
# General issue

## Description:

https://devdocs.magento.com/cloud/integrations/bitbucket-integration.html

Upon creating the integration with bitbucket as per the guide, it now creates the webhook in bitbucket for you. However the next task tells you to create the webhook. I had 2 webhooks created and sending updates on every push to the repository. I believe this caused a conflict. Once I removed the manually created webhook, the integration began working.

## Possible solutions:

Replace the section that instructs users to manually create a webhook with guidance about the autogenerated webhook.

PR #7830, opened 2020-09-07
Guidance about auto-generated webhook on Bitbucket

BarnyShergold:

Satisfy requirement on issue #6332 

## Affected DevDocs pages

https://devdocs.magento.com/cloud/integrations/bitbucket-integration.html
-  ...


devops-devdocs:

src/cloud/integrations/bitbucket-integration.md
```diff
@@ -164,14 +164,18 @@ To create an OAuth consumer in Bitbucket:
...
+method of setting up a Bitbucket intergration detailed on this page, when followed correctly, should automatically create a webhook. But it is advisable to
+check this and if not added automatically, create it manually.
 
 1. Log in to your [Bitbucket](https://bitbucket.org/account/signin/) account.
 
 1. Click **Repositories** and select your project.
 
 1. Click **Settings** > **Workflow** > **Webhooks**.
 
+1. Check to see if a webhook has been added for Cloud to the repository. If this has been added **STOP**.
```
rogyar, line 177: I like the **STOP** idea here, but it might be a bit scary for readers. I would suggest something like the following:
```suggestion
1. Check to see if a webhook has been added for Cloud to the repository. If it has been added, you may skip the following steps.
```

BarnyShergold: Things have changed, years ago STOP was used often in process flow docs 😄 

However, I'm okay to soften the language but I think the word "should" is better then "may" as "may" implies it's just as okay to continue. ('this has' is correct English)

hguthrie: Actually, neither are required. In a step, you can be succinct. For example:

"Verify the webhook integration. If the webhook is active, skip the remaining steps."
BUT, they need some example with response. Not everyone is as comfortable with Bitbucket.

On another note, although I am in favor of verification steps, what has transpired that requires this step?

BarnyShergold: @hguthrie - Seems that the issue raiser found that if the previous steps are followed, the required hook is created for you. So if you dive in and just follow the Hook creation steps, you end up with 2 hooks and it causes problems.

Happy to even more succinct and go with your suggestion, but what do you mean "example with response"?

TBH if someone is not comfortable using BitBucket, should they be doing this? 😄

hguthrie: We know how curious persons are. Or even those that are on auto-pilot before their cup of coffee.

If I want a user to verify something, I typically describe an easy method to verify the their action is complete, even if it is super-simple. And show or describe to them a proof-of-life response that I would expect to see. For example, to verify an integration, you could use CLI to check integrations `mgc integrations`, but that may not sufficient enough to confirm a hook.

In this case, you could provide a link to the section that describes how to test the integration: https://devdocs.magento.com/cloud/integrations/bitbucket-integration.html#test-the-integration

BarnyShergold: @hguthrie - Okay I had added a way for them to identify if they have a Magento Cloud hook (you never know they could have others, we do) and rather than link to the separate page it says go to the testing section on the same page.

I think I'm happy with this now 😄

hguthrie (changes requested):
Great! A couple of minor changes...

hguthrie (approved):

hguthrie:
running tests

src/cloud/integrations/bitbucket-integration.md
```diff
@@ -164,13 +164,17 @@ To create an OAuth consumer in Bitbucket:
 
 ### Add a webhook in BitBucket
 
-In order to communicate events—such as a push—with your Cloud Git server, you need to create a webhook for your BitBucket repository.
+In order to communicate events—such as a push—with your Cloud Git server, is it necessary to have a webhook for your BitBucket repository. The method of setting up a Bitbucket integration detailed on this page, when followed correctly, automatically creates a webhook. It is important to verify the webhook to avoid creating multiple integrations.
```
meker12, line 167:
```suggestion
In order to communicate events—such as a push—with your Cloud Git server, must have a webhook for your BitBucket repository. The method of setting up a Bitbucket integration detailed on this page, when followed correctly, automatically creates a webhook. It is important to verify the webhook to avoid creating multiple integrations.
```

PR #7841, opened 2020-09-08
Facilitate rake update tasks

dshevtsov:

This pull request (PR) updates rake update tasks:
- Omits pulling updates for submodules at subrepos
- Reads directories with subrepos from Docfile instead of hardcoded
- Instantiates content map only once and reuses it in tasks

### How to test

Run `rake update:subrepos` (or `rake update:all`)

- The result should report about all the directories listed in the Docfile
- No failures caused by submodules at subrepos

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -17,6 +17,9 @@ require 'colorator'
 # Require helper methods from the 'lib' directory
 Dir.glob('lib/**/*.rb') { |file| require_relative(file) }
 
+# Instantiate Docfile data for usage in tasks
+@content_map = DocConfig.new.content_map
+
 desc "Same as 'rake', 'rake preview'"
 task default: %w[preview]
 
--- a/rakelib/multirepo.rake
+++ b/rakelib/multirepo.rake
@@ -7,8 +7,7 @@ namespace :multirepo do
   desc 'Create a file tree for devdocs website and get all required content'
   task :init do
     protocol = ENV['token'] ? "https://#{ENV['token']}@github.com/" : 'git@github.com:'
-    content_map = DocConfig.new.content_map
-    content_map.each do |subrepo|
+    @content_map.each do |subrepo|
       repo_url = protocol + subrepo['repository'] + '.git'
       add_subrepo(subrepo['directory'], repo_url , subrepo['branch'], subrepo['filter'])
     end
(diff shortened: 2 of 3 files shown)


jeff-matthews (commented):
This works as described, however, I think we should discuss the use of submodules in general. 

I'm glad that the current rake task failed and notified us that we're using a submodule in the page builder subrepo because we should not be doing that in production. 

With these changes, it won't be as obvious if a submodule is introduced into a subrepo and I worry that it could potentially break something or create an unexpected user experience.

bdenham (approved):
Works well. 

A bit of explanation for this scenario. In my case, having the parliament submodule in the `master` branch is/was a test for operating parliament alongside the devdocs pipeline. And until today, that had been working fine for a couple of weeks. I setup the `magento2-page-builder-docs` to trigger GitHub actions that publish the parliament output to the repo's `gh-pages` branch. This allows me to check both outputs (devdocs vs parliament) while I work on achieving parity on the parliament side.

Meanwhile, the devdocs pipeline is/was still doing its own thing, processing and publishing the same markdown files with Jekyll. This ensured I could always compare to parliament's output.

I used the `docs/` directory on the `master` branch is my single source of truth.

In such a scenario, the two pipelines don't intersect. The `docs/` directory stays pristine with markdown files only.

Hope that helps in explaining this particular scenario.

dshevtsov:
In the page-builder case, the link to a submodule existed only at the `.git` directory. It was fetched, but wasn't checked out, because of sparse checkout filtering.

dshevtsov:
The issue with `Could not access submodule 'parliament'` appeared last week for me. I just forgot to report about it.

dshevtsov:
I think it's okay to allow a code repository to contain submodules, because developer teams may need to use them. Since we do not pull or checkout any content for the submodules, I don't see major risks in explicitly avoiding them. For example, if a submodule was public, then nothing would've break, and the addition of a submodule would be less noticeable.

jeff-matthews (approved):
Ok, if there are no major concerns from anyone else about using submodules, then I'm good. 

@bdenham, that means you can keep master as the source of truth for both scenarios you describe.

dshevtsov:
running tests

bdenham:
Great. I'll keep it there since it will be a bit easier to keep doing the comparisons that way. 

@dshevtsov, I'm glad to hear you got an error on it last week. That was a head scratcher. Mystery solved.

Issue #4495, opened 2019-05-12
Provide detailed steps of how you can execute tests on your local machine before creating a PR


hostep:
# Community engineering issue

Type of Doc:
- [x] New topic
- [ ] Topic Update, Devdoc link: 
- [ ] Other

Magento 2 version:
- [ ] 2.1.X
- [x] 2.2.X
- [x] 2.3.X
- [ ] back port
- [ ] forward port

Notifications of Changes:
- [ ] Changes to Admin interface?
- [ ] New software requirements?
- [ ] New REST endpoints or updated GraphQL query syntax?
- [ ] New command line option?

## Content

Hi!

When creating a PR on the magento2 repo, there are a bunch of [automated tests](https://github.com/magento/magento2/wiki/Magento-Automated-Testing) which get executed. But there currently doesn't seem to be a clear explanation of how to execute these tests yourself.
It would be nice if we have some documentation which says how to run those tests, so we can already prepare the PR ourselves to get accepted faster and not having to go back and forth and wait until the automated tests have run after changing something.
There is already some good documentation around [how to run unit tests](https://devdocs.magento.com/guides/v2.3/test/unit/unit_test_execution_cli.html), but the other tests (static tests for example) are not documented properly I think. There is [this page](https://devdocs.magento.com/guides/v2.3/config-guide/cli/config-cli-subcommands-test.html), which says to call `bin/magento dev:tests:run static`, but that doesn't seem to give the same results as the static tests ran through automatisation.

## Additional information

For example, I created a PR https://github.com/magento/magento2/pull/22733 which apparently failed on some static tests. And then a Magento developer fixed those. Which is great btw, but I would like to be able to see those failures myself and fix them myself if possible before creating the PR.
When [asking](https://github.com/magento/magento2/pull/22733#issuecomment-491242429) how to execute these static test, I didn't get a very clear answer, so I went looking in the documentation where to find information regarding that and I couldn't find anything around it (unless I missed it somehow).

Thanks!

lorikrell:
@hostep, great request! I'm working with @dmanners on a set of [Beginner Guides](https://github.com/magento/magento2/wiki/Getting-Started) and the [wiki](https://github.com/magento/magento2/wiki) for Magento2 repo to document some of this information. I've assigned myself to the ticket, and will work towards that content. 

We've covered the Issue process. Next is PR! More info soon for this one.

DanielRuf:
Exactly the issue that I face at the moment.
[...]

PR #7951, opened 2020-09-29
Added Static and SemVer tests docs

gabrieldagama:

This pull request (PR) is to add documentation on how to run static and SVC tests.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.4/test/static/static_test_execution.html
-  https://devdocs.magento.com/guides/v2.4/test/svc/semver_test_execution.html

whatsnew
Added the [Static Testing](https://devdocs.magento.com/guides/v2.4/test/static/static_test_execution.html) and [Semantic Version Checker](https://devdocs.magento.com/guides/v2.4/test/svc/semver_test_execution.html) topics to the Test Guide.


src/guides/v2.4/test/static/static_test_execution.md
```diff
@@ -0,0 +1,56 @@
+---
```
BarnyShergold, line 1: This file is identical to the 2.3 version - therefore it should be created as a symlink, not a separate file.

src/guides/v2.3/test/svc/semver_test_execution.md
```diff
@@ -0,0 +1,38 @@
...
+**Step 3.** In order to run the Semantic Version Checker we need to have one folder with the feature changes and another folder with mainline code (without changes),
+so you may need to clone the Magento repository to a separate folder to perform the comparison.
+
+**Step 4.** Navigate the `magento-semver` folder and run the Semantic Version Checker compare command:
+
+```bash
+.bin/svc compare ../magento2-mainline ../magento2
+```
+
+The first parameter is the mainline code (2.3-develop branch) without any changes, and the second parameter is the path to the Magento with your feature changes.
```
BarnyShergold, line 36: Branches are not mentioned as this constantly changes. Please rework to not mention a branch name.

src/guides/v2.4/test/svc/semver_test_execution.md
```diff
@@ -0,0 +1,38 @@
+---
```
BarnyShergold, line 1: This file is identical (it will be if the reference to the development branch is removed) to the 2.3 version - therefore it should be created as a symlink, not a separate file.

BarnyShergold (changes requested):
Please add links to the 2.3 versions of the files in the description
Please fix linting errors

src/guides/v2.3/test/static/static_test_execution.md
```diff
@@ -0,0 +1,56 @@
...
+
+```text
+# Format: <componentType=module|library|theme|language|*> <componentName> <globPattern> or simply <globPattern>
+app/code/Magento/CatalogSearch/Model/Search
+```
+
+**Step 5.** Create a copy of `dev/tests/static/phpunit.xml.dist` file and call it `phpunit.xml`:
+
+**Step 6.** Add a new testsuite to the `dev/tests/static/phpunit.xml` file inside the `<testsuites>` node:
+:
```
gabrieldagama, line 39:
```suggestion

```

gabrieldagama:
Hi @BarnyShergold, thanks for reviewing it. I've applied the suggestions, can you review it again?

Thanks.

BarnyShergold:
@gabrieldagama - Not sure what's happened to your commits but - 
If I read this right, you have deleted the 2.3 version and now 2.3 is a symlink of 2.4 - this is the wrong way around I'm afraid. 2.3 should be the real file and 2.4 the symlink. This is how it's done through the whole of devdocs.

The 2.4 files don't seem to have all the suggested fixes applied - eg "straight forward" is one word and line 39 of static_text_execution.md needed removing.

Can you please check and correct?

gabrieldagama:
@BarnyShergold yeah, I probably messed up with the commits there, my bad. I've re-applied the suggestions.

Thanks.

BarnyShergold (approved):
Excellent work

src/guides/v2.3/test/svc/semver_test_execution.md
```diff
@@ -0,0 +1,37 @@
...
+so you may need to clone the Magento repository to a separate folder to perform the comparison.
+
+1. Navigate the `magento-semver` folder and run the Semantic Version Checker compare command:
+
+   ```bash
+   bin/svc compare ../magento2-mainline ../magento2
+   ```
+
+The first parameter is the mainline code without any changes, and the second parameter is the path to the folder with your changes.
+This command you output the results of the Semantic Version Checker tool.
```
dobooth, line 37: Where do the results output to? Is it obvious?

src/guides/v2.3/test/static/static_test_execution.md
```diff
@@ -0,0 +1,55 @@
...
+---
+
+Executing Magento 2 static tests is straightforward. They can be executed in several ways.
+
+### Running static tests on all files
+
+To run static tests on all files, navigate to the Magento base directory and execute the following command:
+
+```bash
+bin/magento dev:test:run static
```
hostep, line 16: @gabrieldagama: out of curiosity, are these the exact same tests that are being ran when you submit a PR on github?
Unless something has changed recently that's not the case. It seems like there are a bunch of extra static tests which are being executed on github.
[I'm looking for easy to comprehend information](https://github.com/magento/devdocs/issues/4495) about how to execute all these static tests on your local machine, it will save so much waiting time when working on a PR if we have this information.

src/guides/v2.3/test/svc/semver_test_execution.md
gabrieldagama: The output is the result of the CLI command in this case, I guess there is a typo there, it should be:
```This command will output the results of the Semantic Version Checker tool.```

Maybe we can make it more explicit by saying:
```This command will output on the console the results of the Semantic Version Checker tool.```

What do you think @dobooth ?

File: src/guides/v2.3/test/static/static_test_execution.md (hunk 6 above)
Comment (gabrieldagama) [reply src/guides/v2.3/test/static/static_test_execution.md:16]: Hi @hostep, this is a valid point, thanks for bringing it up! Yes, they are the same tests, but they run a little bit differently, on our automated tests, the Static Tests are split in their separate test suite respectively (Less, HTML, PHP, Integrity...) and when you run locally, it runs all the suites at once. We should get the same results locally.

But there are somethings that I've noticed:
- Some of the tests are run based on changed files (code style, for example), so if you commit your changes before running the tests you may not get the errors.
- We do have some differences in EE and B2B versions, so if you don't have those installed you may get different results.

Maybe we can add this information to the docs as well, what do you think?

Hope this clarifies a little bit your question. If you have some PR that we can look at to see the differences it would be great!

Comment (hostep) [reply src/guides/v2.3/test/static/static_test_execution.md:16]: Ok thanks for the info!

Ah yes, I seem to remember that it was almost always some extra tests in B2B that failed. Unfortunately we are only using Open Source so I have no access to those extra tests.

What is the reason for the static tests to be different on EE & B2B projects btw? These are static tests, in my opinion they should be 100% the same between the different type of Magento projects. I understand there is a difference for integration & functional tests, but I don't understand why there is a difference for static tests?
Would be great if the differences in static tests could get aligned so they are the same over all types of Magento projects, no?

Comment (gabrieldagama) [reply src/guides/v2.3/test/static/static_test_execution.md:16]: I agree with you, from what I can see the PHP check are exactly the same, it just changes some files that are ignored or not. But, there are some differences in other static tests, like Integrity and Legacy tests. I'm fully aware of exactly the reasons and what are the differences, but I may be able to check internally.

But yeah, from my point of view, ideally we would have a defined test strategy that is publicly available and applied to all versions.

Comment (hostep) [reply src/guides/v2.3/test/static/static_test_execution.md:16]: I also seem to remember vaguely that php mess detector static test ran on B2B and not on OS, but I might be mistaken

File: src/guides/v2.3/test/svc/semver_test_execution.md (hunk 5 above)
Comment (dobooth) [reply src/guides/v2.3/test/svc/semver_test_execution.md:37]: The results of the Semantic Version Checker are outputted to the console.

Comment (gabrieldagama) [inline src/guides/v2.3/test/svc/semver_test_execution.md:37]:
```suggestion
The results of the Semantic Version Checker are outputted to the console.
```

dobooth (approved):

dobooth:
running tests

PR #7953, opened 2020-09-29
Update php extensions template to include ext-sockets

meker12:

Update the PHP extensions template to include `ext-sockets` requirements in v2.3 and 2.4.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.4/install-gde/prereq/php-settings.html#verify-installed-extensions
- https://devdocs.magento.com/guides/v2.3/install-gde/prereq/php-settings.html

### DIFF
--- a/src/_includes/install/php-extensions-template.md
+++ b/src/_includes/install/php-extensions-template.md
@@ -7,6 +7,7 @@
   {% endif %}
 {% endfor %}
 {% unless page.guide_version == '2.3' %}
+
 *  ext-sockets
 
 {% else %}
--- a/src/_includes/install/php-extensions-template.md
+++ b/src/_includes/install/php-extensions-template.md
@@ -13,5 +13,5 @@
 {% else %}
 
 {:.bs-callout-warning}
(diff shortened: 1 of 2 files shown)


dshevtsov (approved):

src/_includes/install/php-extensions-template.md
```diff
@@ -6,3 +6,12 @@
...
   {% endif %}
 {% endfor %}
+{% unless page.guide_version == '2.3' %}
+
+*  ext-sockets
+
+{% else %}
+
+{:.bs-callout-warning}
+If you install Magento by cloning from the [GitHub](https://github.com/magento/magento2) repository, you must have the [ext-sockets](https://github.com/php-amqplib/php-amqplib/blob/master/CHANGELOG.md#281---2018-11-13) installed on your instance.
```
hostep, line 16: Like mentioned previously, I would remove this, it only applies for Magento 2.3.4 and 2.3.5. And who is installing Magento by cloning the git repo anyways? It's not really recommended for production deployments and now that those versions are outdated I don't see the point of mentioning this notice.

Just my 2 cents.

hostep (approved):
Looks good!

(I have no insight if Magento 2.3.6 won't also require the sockets extension, so maybe this needs to be revised after 2.3.6 gets released)

meker12:
running tests

Issue #7580, opened 2020-07-23
Add details to Define the GraphQL schema for a module 


RimantasAdeoWeb:
# General issue

The lack of examples in some sections make this page uninformative.

Path to file: /guides/v2.4/graphql/develop/create-graphqls-file.html

## Description:

There are no examples under "Declare input attributes", "Specify output attributes", "Define the output interface", "Define enumerations"
It makes these sections nearly useless to people not already familiar with Magento's integration of GraphQl - it just says what is possible to do but doesn't show how to do it. 

## Possible solutions:

Link to a schema file in some module which has a good example of an implementation of whatever the section is about.
Add an example in the page itself, such as in the "Define queries" section

keharper:
Thanks for taking this PR, @chandru-199

chandru-199:
If any GraphQL expertise take this PR, it will be better I think. Thank you!.

gelanivishal:
@magento I am working on this

sathiyaaa:
@magento I am working on this

PR #7996, opened 2020-10-05
Added more detail for GraphQl schema for a module

gelanivishal:

This pull request (PR) for https://github.com/magento/devdocs/issues/7580 to add more detail about GraphQL schema for a module

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.4/graphql/develop/create-graphqls-file.html
- https://devdocs.magento.com/guides/v2.3/graphql/develop/create-graphqls-file.html

### DIFF
--- a/src/guides/v2.3/graphql/develop/create-graphqls-file.md
+++ b/src/guides/v2.3/graphql/develop/create-graphqls-file.md
@@ -50,17 +50,63 @@ If all your module's attributes are extension attributes for existing modules, t
 
 You must explicitly define each attribute that can be used as input in a GraphQL query. In the simplest cases, you can create a single `type` definition that includes all the input, output, and sorting attributes for an object. This might not be possible if your module performs calculations, or otherwise has attributes that aren't available at the time of the query.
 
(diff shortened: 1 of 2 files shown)


keharper:
@gelanivishal I deleted this content (more or less, I'm not sure if it's exactly the same) somewhere around 2.3.2. I had hoped to have a solid mock resolver that I could point to, but that didn't happen, so I removed all references to the sample Volumizer module.

At that time, we didn't have very much GraphQL coverage outside of Catalog and Customer. I don't know if using Catalog as the basis for a sample module is the most intuitive. (Maybe it is--I just want some thought put into it.) However, I know that the syntax of the `products` query changed in 2.3.4 or 2.3.5. The `filter` argument now requires an `ProductAttributeFilterInput` object, so some updates are required.

This PR needs the following:

* Evaluate whether to continue using Volumizer as an example
* Add an intro that describes what you're doing. In the earlier version of this topic, this info was at the beginning of the topic
* Make sure all real code samples are up to date

keharper:
Closing this PR due to inactivity. @gelanivishal, feel free to reopen.

PR #8007, opened 2020-10-07
MCLOUD-6939: Errors occur running magento commands in composer cloud docker installation

oshmyheliuk:

Added documentation for new Cloud build variable `SKIP_COMPOSER_DUMP_AUTOLOAD` to skip composer dump-autoload processing when deploying a Cloud project to the Cloud Docker environment. 

## Affected DevDocs pages

-  https://devdocs.magento.com/cloud/env/variables-build.html

## Magento source code

https://github.com/magento/ece-tools/pull/784

### DIFF
--- a/src/cloud/env/variables-build.md
+++ b/src/cloud/env/variables-build.md
@@ -203,7 +203,7 @@ stage:
 
 ### `SKIP_COMPOSER_DUMP_AUTOLOAD`
 
-Skip executing of `composer dump-autoload` command.
+Do not run the composer dump-autoload command.
 
 -  **Default**— _Not set_
 -  **Version**—Magento 2.1.4 and later
--- a/src/cloud/env/variables-build.md
+++ b/src/cloud/env/variables-build.md
@@ -201,6 +201,19 @@ stage:
     SKIP_SCD: false
 ```
 
+### `SKIP_COMPOSER_DUMP_AUTOLOAD`
+
+Skip executing of `composer dump-autoload` command.
+
+-  **Default**— _Not set_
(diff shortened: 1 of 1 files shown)


src/cloud/env/variables-build.md
```diff
@@ -201,6 +201,19 @@ stage:
     SKIP_SCD: false
 ```
 
+### `SKIP_COMPOSER_DUMP_AUTOLOAD`
+
+Skip executing of `composer dump-autoload` command.
```
BarnyShergold, line 206:
```suggestion
Skip execution of `composer dump-autoload` command.
```

BarnyShergold (changes requested):

src/cloud/env/variables-build.md
meker12: Suggest: Avoid using terms like execute and terminate:

 Do not run the `composer dump-autoload` command.

Why do you need to set this variable?  What does it accomplish?

meker12 (changes requested):

src/cloud/env/variables-build.md
oshmyheliuk: @meker12 Executing `composer dump-autoload` command causes an error when running some Magento commands such as `bin/magento deploy:mode:set production` or `bin/magento -n setup:upgrade`. These commands can be run on writable systems, for example in the magento-cloud-docker `test` container. So we added a possibility to skip it.
This is a core issue https://github.com/magento/magento2/issues/23251, but I don't know when it will be fixed.

PR #8059, opened 2020-10-15
Update elasticsearch.md

jonashrem:
Clarified Elasticsearch documentation

As Elasticsearch has no relevant breaking changes between minor versions (see https://www.elastic.co/guide/en/beats/libbeat/current/breaking-changes.html ), Elasticsearch 7.x should generally be remanded instead of 7.6.

I also renamed the html anchor to match Elasticsearch version.

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.4/install-gde/prereq/elasticsearch.html

### DIFF
--- a/src/guides/v2.4/install-gde/prereq/elasticsearch.md
+++ b/src/guides/v2.4/install-gde/prereq/elasticsearch.md
@@ -25,7 +25,7 @@ As of Magento 2.4, all installations must be configured to use [Elasticsearch][]
 
 You must install and configure Elasticsearch 7.6.x before upgrading to Magento 2.4.x.
 
-Be aware, that Magento 2.4.x is only tested with Elasticsearch 7.6.x at the moment. You can use other (7.x) versions at your discretion, but we recommend using the tested version of Elasticsearch.
+Magento 2.4.x is tested with Elasticsearch 7.6.x only. You can use other 7.x versions at your discretion, but we recommend using the tested version of Elasticsearch.
 
 {:.bs-callout-info}
 Magento does not support Elasticsearch 2.x, 5.x, and 6.x.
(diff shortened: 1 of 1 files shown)


jeff-matthews:
@jonashrem, in this context, "supported" means that ES 7.6 is the only version of the ES 7.x release line that we have officially tested with Magento. We cannot claim support for all ES 7.x versions.

jeff-matthews:
However, I suppose that the anchor updates are still valid, so I'll reopen this.

jonashrem:
I see. I have reverted that part.

What is your opinion about a hint that only Elasticsearch 7.6 has been tested but other 7.x Versions "should work in general"?

jeff-matthews:
That's probably a good idea @jonashrem. We have a hint like that on the [main system requirements page](https://devdocs.magento.com/guides/v2.4/install-gde/system-requirements-tech.html#elasticsearch), so it makes sense to include it here as well.

jonashrem:
yes, there is also a formulations like this for database ( https://devdocs.magento.com/guides/v2.4/install-gde/system-requirements-tech.html#database )

jonashrem:
I just added the information.

src/guides/v2.4/install-gde/prereq/elasticsearch.md
```diff
@@ -23,7 +23,9 @@ As of Magento 2.4, all installations must be configured to use [Elasticsearch][]
 
 ## Supported versions {#es-spt-versions}
 
-You must install and configure Elasticsearch 7.6.x before upgrading to Magento 2.4.0.
+You must install and configure Elasticsearch 7.6.x before upgrading to Magento 2.4.x.
+
+Be aware, that Magento 2.4.x is only tested with Elasticsearch 7.6.x at the moment. You can use other (7.x) versions at your discretion, but we recommend using the tested version of Elasticsearch.
```
jeff-matthews, line 28:
```suggestion
Magento 2.4.x is tested with Elasticsearch 7.6.x only. You can use other 7.x versions at your discretion, but we recommend using the tested version of Elasticsearch.
```

jeff-matthews (changes requested):
Some minor editorial suggestions

jeff-matthews (approved):
Thanks @jonashrem!

jeff-matthews:
running tests

PR #8105, opened 2020-10-22
Update install-cli-subcommands-admin.md

chickenland:
updated example admin user creation to include quotes wrapped around it as per the requirement in the definition for it above


BarnyShergold (approved):
Nice little tweak

hguthrie (dismissed):
Interesting. I bet there are few other places in the docs where this could be added for consistency.

BarnyShergold:
> Interesting. I bet there are few other places in the docs where this could be added for consistency.

I found at least 20 instances....

keharper:
The command works for me as listed. Why do we state to enclose the password in single quotes?

keharper:
Let me ask a more directed question. Are there special characters that can interfere with the execution of the command if the entire password is not enclosed in single quotes? I can't think of any other possible reason that we'd state to enclose the password value.

hguthrie:
I agree with you. The only reason that instantly comes to my mind is to assure literal interpretation.

keharper:
I'm requesting the following modifications to this PR.

1. Revert the command example.

2. Modify the description of the `--admin-password` argument (line 47) to be similar to the following

If the password string contains special characters that require literal interpretation (such as backslashes or spaces), enclose the password in single quotations.

Then delete the "For example" sentence.

keharper:
I forgot to mention @chickenland @BarnyShergold

dobooth:
Hi @chickenland Any updates here?

dobooth:
Closing as abandoned.

keharper (approved):

keharper:
running tests

PR #8179, opened 2020-11-04
Added Purchase Order to comany resource table

dobooth:

This pull request (PR) adds Purchase Order items to the Company Resource table.

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.4/b2b/roles.html

whatsnew
Added Purchase Order resources to the [Manage Company Roles](https://devdocs.magento.com/guides/v2.4/b2b/roles.html) topic.


BarnyShergold (changes requested):
RE 2.4 file -
These 2 sections are not arranged as tables (as on 2.3 version of page) and are really hard to read. Also this page has the first list but this is not on the 2.3 page.

If the first list has is supposed to mimic the Roles and Permissions screen then please indicate this. 

The second list is better arrange as the table on 2.3 - it is difficult to read properly.

Had to add this as a general comment as GitHub kept saying "line must be part of the diff" when trying to add a comment

BarnyShergold:
Table  should be -
--- | ---

(Can't add suggestions or comments)

The table is being badly messed up due to the extra column and not used

BarnyShergold (changes requested):
Still can't add comments!

keharper:
@BarnyShergold I asked for these changes because we need to better show the permissions hierarchy. In 2.4.2, GraphQL adds several mutations/queries that assign and return B2B roles. I didn't ask for 2.3 to be updated because B2B purchase orders are not available in 2.3. 

I'm not at a point at the moment where I can do a preview build of the latest changes. But from scanning that table, it looks like it will show the roles in a nested format, similar to what the B2B user will see.

The "line must be part of the diff"  error sometimes occurs when a symlink is broken. We reported the problem to GH well over a year ago, and it remains a problem.

BarnyShergold (approved):
Looks awesome now! 😄

keharper (approved):

dobooth:
running tests

PR #8184, opened 2020-11-05
pass "entity_type_code" instead of "entity_type_id"

dfelton:

Frontend code should not have to rely on knowing the `eav_entity_type.entity_type_id` value of the model it is querying attribute information on. Local tests show that this field also supports `eav_entity_type.entity_type_code` values as well.

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.3/graphql/queries/custom-attribute-metadata.html
- https://devdocs.magento.com/guides/v2.4/graphql/queries/custom-attribute-metadata.html


rogyar:
Can confirm that the entity type code should be passed instead of ID. An example of the test from the codebase is below

https://github.com/magento/magento2/blob/2.3/dev/tests/api-functional/testsuite/Magento/GraphQl/Catalog/ExceptionFormatterDefaultModeTest.php#L22

rogyar (approved):

dobooth (approved):

PR #8196, opened 2020-11-06
Installation Guide reorganization

jeff-matthews:

This pull request (PR) is the first attempt at reorganizing the table of contents (TOC) for the _Installation Guide_. The goal is to remove unnecessary content and move the recommended content so that it is easier to find.

## Affected DevDocs pages

- Several pages in the Installation Guide

## Additional info

I didn't change topic titles to match revised TOC labels because I expect some churn as we deliberate internally. After we decide on titles, I can make updates. For example:

Screen Shot 2020-11-06 at 10 20 08 AM

See internal wiki for link to internal staging build.

## To do

- [ ] Rename topic titles to match TOC label changes (where applicable)
- [ ] Delete files for entries that were removed from the TOC; set up redirects if necessary

### DIFF
--- a/src/_data/main-nav.yml
+++ b/src/_data/main-nav.yml
@@ -59,7 +59,7 @@
   children:
 
     - label: Installation Guide
-      url: /install-gde/bk-install-guide.html
+      url: /install-gde/install-flow-diagram.html
 
     - label: Magento Extensions Guide
       url: /extensions/
--- a/src/_data/toc/installation-guide.yml
+++ b/src/_data/toc/installation-guide.yml
@@ -3,13 +3,7 @@ pages:
 
     - label: Installation flow diagram
       url: /install-gde/install-flow-diagram.html
-
-    - label: Quick start (recommended installation)
-      url: /install-gde/composer.html
-
-    - label: Install options
-      url: /install-gde/bk-install-guide.html
-
+    
     - label: Magento system requirements
       url: /install-gde/system-requirements.html
 
@@ -73,6 +67,12 @@ pages:
 
             - label: Update installation dependencies
               url: /install-gde/install/prepare-install.html
+
+    - label: Quick start (recommended installation)
+      url: /install-gde/composer.html
+
+    - label: Install options
+      url: /install-gde/bk-install-guide.html
 
     - label: Setup Wizard Install
       include_versions: ["2.3"]
--- a/src/_data/toc/installation-guide.yml
+++ b/src/_data/toc/installation-guide.yml
@@ -17,11 +17,11 @@ pages:
         - label: nginx
           url: /install-gde/prereq/nginx.html
 
-        - label: PHP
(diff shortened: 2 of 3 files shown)


dobooth:
I know it's way early but I really prefer alphabetized TOCs, unless there is a better order (tutorial steps, etc).

jeff-matthews:
I do too, when possible. I think we can probably do that within subsections of the TOC, but we may not have that luxury with the root order.

Something else to consider is updating the link to the Installation Guide in the top nav. Currently, it goes to https://devdocs.magento.com/guides/v2.4/install-gde/bk-install-guide.html, but that topic is no longer in the same place and if that's the entry point for the guide, it dumps you in an expanded TOC instead of the top node.

shrielenee:
* Should we have an intro overview page at the top of the Installation guide section---something to say "We have a flow diagram, if you're a contributor do this and if you're not do this, and here is the general flow you should follow" type thing?
* Should System requirements go above any install-related topics in the TOC to follow a kind of "gather knowledge, then proceed to install" approach? Same question about Prerequisites?
* Should Enable or disable modules be in Post-install, as it is, or in Tutorials? Same with Uninstall modules or Enable or disable maintenance mode (and similar topics)? I am not sure where to draw the line between stuff you should/could do after install, and stuff that is totally optional depending on what your goals are.---Update: I see you made a comment about this on our planning page :)

shrielenee:
> I do too, when possible. I think we can probably do that within subsections of the TOC, but we may not have that luxury with the root order.
> 
> Something else to consider is updating the link to the Installation Guide in the top nav. Currently, it goes to https://devdocs.magento.com/guides/v2.4/install-gde/bk-install-guide.html, but that topic is no longer in the same place and if that's the entry point for the guide, it dumps you in an expanded TOC instead of the top node.

Yes, and I'm wondering if we need an install intro topic page at the top to link to. I mentioned this in my previous comment here.

jeff-matthews:
Agreed. The original intro page is the "How to get the Magento software" page in [this comment](https://github.com/magento/devdocs/pull/8196#issuecomment-723171779). That's also the page we use to link from the top nav (Setup > Installation Guide). We need to update it to whatever the top level node in the TOC will be.

dshevtsov:
It is unclear what does _Round 1_ means from a PR title. Since the PR is public, it should be clear to any contributor.

jeff-matthews:
It was taken from my commit message and I forgot to update it when creating the PR.

dobooth (approved):

shrielenee:
I will approve this with consideration for these comments in future iterations: https://github.com/magento/devdocs/pull/8196#issuecomment-723173860

shrielenee (approved):

jeff-matthews:
Thanks @dobooth & @shrielenee. 

I'll push some changes to address your comments Shrie before merging.

jeff-matthews:
Will complete To do items in a separate PR.

PR #8215, opened 2020-11-11
Add Page Builder modules to the Module Reference Guide

dshevtsov:

This pull request (PR) adds Page Builder modules to the Module Reference Guide in both versions 2.3 and 2.4.
Also,
- Removing blank lines and extra spaces at the beginning and at the end of a topic.
- Changing labels in tables of contents to clarify relevance to Magento products.

## Affected DevDocs pages

-  most of module topics in the MRG [2.3](https://devdocs.magento.com/guides/v2.3/mrg/intro.html) and [2.4](https://devdocs.magento.com/guides/v2.4/mrg/intro.html)

whatsnew
Added Page Builder modules to the Module Reference Guide for [2.4](https://devdocs.magento.com/guides/v2.4/mrg/intro.html) and [2.3](https://devdocs.magento.com/guides/v2.4/mrg/intro.html).

### DIFF
--- a/src/_data/codebase/v2_3/mrg/page-builder/CatalogPageBuilderAnalytics.yml
+++ b/src/_data/codebase/v2_3/mrg/page-builder/CatalogPageBuilderAnalytics.yml
@@ -1,0 +1,9 @@
+---
+title: Magento_CatalogPageBuilderAnalytics
+source_repo: magento2-page-builder
+release: v1.3.3
+github_path: app/code/Magento/CatalogPageBuilderAnalytics/README.md
+last_modified_at: '2019-03-05 18:34:09 +0100'
+content: The Magento_CatalogPageBuilderAnalytics module configures data definitions
+  for a data collection related to the PageBuilder module entities to be used in [Advanced
+  Reporting](http://devdocs.magento.com/guides/v2.2/advanced-reporting/modules.html).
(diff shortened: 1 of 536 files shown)


src/_data/codebase/v2_3/mrg/page-builder/CatalogPageBuilderAnalytics.yml
```diff
@@ -0,0 +1,9 @@
+---
+title: Magento_CatalogPageBuilderAnalytics
+source_repo: magento2-page-builder
+release: v1.3.3
```
jeff-matthews, line 4: Where is the "v" in the release property coming from (`release: v1.3.3`)? I don't see any other release properties for any other modules prepended with a "v". Does it matter?

dshevtsov: This property corresponds to a release tag as-is. That's how PageBuilder releases are tagged, differently from other repos (ce/ee/b2b/inventory).

dshevtsov: Since there is no single convention for tagging across different repositories. I used to add `v` during generation, but in case with PageBuilder it turned to `vv`. I decided to use it as-is to avoid any such issues in the future.

jeff-matthews: Good to know, thanks.

jeff-matthews (approved):

dshevtsov:
running tests

PR #8246, opened 2020-11-19
7786-Changed the code for afterSave function in adding-attributes.md file

### LINKED ISSUE #7786: `beforeSave` functionality is incorrectly described
# General issue

https://devdocs.magento.com/guides/v2.4/extension-dev-guide/extension_attributes/adding-attributes.html

Directly following the general patterns described on [this page](https://devdocs.magento.com/guides/v2.4/extension-dev-guide/extension_attributes/adding-attributes.html) - for an extension attribute on a product, with a plugin connected to the product repository - the `afterSave` function _will not_ update the extension attribute. For example, making an API call with updated extension attribute will leave it unchanged.

## Description:

Essentially, the data in the product instance returned as the result of a `save` call (passed as the 2nd parameter to an 'after' plugin function) is not what was sent via the API. 

After a little digging, the data that reaches the `beforeSave` function **is** correct. However, the final step in the product repository save process is a get call, presumably to refresh data, which correspondingly triggers the `afterGet` function to (re-)populate the extension attributes on the product returned as the result.

The code example on the page is using the `ProductInterface` that is the result returned from the function, due to the attribute sequence, hence why it can't / won't update the extension attribute.

However, it should also be noted that is specifically a peculiarity of the product repository and does not necessarily hold true for other repositories.

## Possible solutions:

We can't simply return the original parameter object, because that would invalidate any processing carried out before / during save. Similarly, we can't just return the result object by itself, since that has the incorrect extension attributes and potentially might be used by something else.  And we can't change this to a `beforeSave` function, because that causes potential issues with product creation (hard to link an extension attribute to a product that doesn't exist in the database yet.)

To resolve this correctly, you actually need both the result object and the original parameter object:
```
public function afterSave
(
     \Magento\Catalog\Api\ProductRepositoryInterface $subject,
     \Magento\Catalog\Api\Data\ProductInterface $result, /** result from the save call **/
     \Magento\Catalog\Api\Data\ProductInterface $entity  /** original parameter to the call **/
     /** other parameter not required **/
) {
    $extensionAttributes = $entity->getExtensionAttributes(); /** get original extension attributes from entity **/
    $ourCustomData = $extensionAttributes->getOurCustomData();
    $this->customDataRepository->save($ourCustomData);

    $resultAttributes = $result->getExtentionAttributes(); /** get extension attributes as they exist after save **/
    $resultAttributes->setOurCustomData($ourCustomData); /** update the extension attributes with correct data **/
    $result->setExtensionAttributes($resultAttributes);

    return $result;
}
...
```

chiranjeevi-cj:

This pull request (PR) 
https://devdocs.magento.com/guides/v2.4/extension-dev-guide/extension_attributes/adding-attributes.html

Directly following the general patterns described on this page - for an extension attribute on a product, with a plugin connected to the product repository - the afterSave function will not update the extension attribute. For example, making an API call with updated extension attribute will leave it unchanged.

## Description

Essentially, the data in the product instance returned as the result of a save call (passed as the 2nd parameter to an 'after' plugin function) is not what was sent via the API.

After a little digging, the data that reaches the beforeSave function is correct. However, the final step in the product repository save process is a get call, presumably to refresh data, which correspondingly triggers the afterGet function to (re-)populate the extension attributes on the product returned as the result.

The code example on the page is using the ProductInterface that is the result returned from the function, due to the attribute sequence, hence why it can't / won't update the extension attribute.

However, it should also be noted that is specifically a peculiarity of the product repository and does not necessarily hold true for other repositories.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.4/extension-dev-guide/extension_attributes/adding-attributes.html

Fixes #7786


src/guides/v2.3/extension-dev-guide/extension_attributes/adding-attributes.md
```diff
@@ -106,14 +106,20 @@ Likewise, the `afterSave` plugin should manipulate the entity data before return
 ```php
 public function afterSave
 (
-    \Magento\Catalog\Api\ProductRepositoryInterface $subject,
-    \Magento\Catalog\Api\Data\ProductInterface $entity
+     \Magento\Catalog\Api\ProductRepositoryInterface $subject,
```
BarnyShergold, line 109: Indenting on this line (and next 2) should have one less space

src/guides/v2.3/extension-dev-guide/extension_attributes/adding-attributes.md
```diff
@@ -106,14 +106,20 @@ Likewise, the `afterSave` plugin should manipulate the entity data before return
...
+     \Magento\Catalog\Api\Data\ProductInterface $result, /** result from the save call **/
```
BarnyShergold, line 110: Normally in PHP single line comments are done with // e.g.` // result from the save call`
Is there a reason the /** .... **/ format is used?

BarnyShergold (changes requested):
Thanks for your contribution - just some formatting queries

src/guides/v2.3/extension-dev-guide/extension_attributes/adding-attributes.md
chiranjeevi-cj: no, I will update it thanks.

chiranjeevi-cj:
@BarnyShergold As in the site(https://devdocs.magento.com/guides/v2.4/extension-dev-guide/extension_attributes/adding-attributes.html), every comment was done with /** .... **/ that's why I followed it.
Should I need to change?

chiranjeevi-cj:
@BarnyShergold Is there anything that needs to be changed?

src/guides/v2.3/extension-dev-guide/extension_attributes/adding-attributes.md
BarnyShergold: No - if it's being used elsewhere then you should stick to it for convention. It's not a must.

BarnyShergold (approved):
Thanks for your contribution!

dobooth (approved):

dobooth:
running tests

PR #8257, opened 2020-11-20
Fix downgrade example

meker12:
Fix versions in example to downgrade MySQL

Fix example in instructions to downgrade Cloud service version, to use MYSQL versions that can be downgraded. Current example downgrades from 10.3 to 10.2 which is not supported.

## Affected DevDocs pages

<!-- REQUIRED List the affected pages on devdocs.magento.com (URLs). 

-  https://devdocs.magento.com/cloud/project/services.html#downgrade-version

### DIFF
--- a/src/cloud/project/services.md
+++ b/src/cloud/project/services.md
@@ -259,7 +259,7 @@ To downgrade a service version by renaming an existing service:
 
      ```yaml
      mysql:
-         type: mysql:10.3
+         type: mysql:10.4
          disk: 2048
      ```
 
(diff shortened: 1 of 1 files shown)


keharper (approved):

meker12:
running tests

meker12:
@jeff-matthews or @dshevtsov  Can one of you run the production build job with update all.  This change to the mySQL versions in the downgrade procedure is not reflected in the published documentation, but it is showing in the [master branch source](https://github.com/magento/devdocs/blob/master/src/cloud/project/services.md).

The current example shows an unsupported downgrade scenario (MySQL 10.3 to 10.2).

jeff-matthews:
@meker12 , the job is running now and should be finished in ~7 minutes.

meker12:
Fixed. Thank you!

PR #8265, opened 2020-11-24
Replace the swagger-ui library with Redoc for 2.2.x

jeff-matthews:

This pull request (PR) replaces the swagger-ui library with Redoc for generating REST API reference docs from the Magento 2.2.x schema.

Since we already have Redoc in the project for Magento 2.3.x asynchronous endpoints, it makes sense to use it for Magento 2.2.x as well instead of having two libraries that do the same thing.

It takes longer for the 2.2.x schema to load with Redoc (~15 seconds), but I think it's worth the security tradeoff.

See internal staging build `#85`.

### Questions about Redoc

- Do we still need to publish reference for Asynchronous Endpoints?
  - This [page](https://devdocs.magento.com/redoc/2.3/index.html) links to both Redocly hosted docs (Synchronous Endpoints) and "local" docs rendered with the Redoc library (Asynchronous Endpoints).
  - Can we remove the Asynchronous Endpoints like we've done in 2.4.x? If so, we can remove the `src/redoc/` directory.
  - If not, that's fine, I just wanted to make sure. See the following screenshot for a glance at the analytics for those docs:

Screen Shot 2020-11-24 at 8 57 40 AM

## Affected DevDocs pages

- https://devdocs.magento.com/swagger/index.html


dobooth (approved):

dshevtsov (commented):
`swagger/index.html` path is misleading for the REST API 2.2 reference in Redoc.
I wonder if it's possible to move it to `guides/v2.2` where the rest of 2.2 documentation is stored.

jeff-matthews:
> `swagger/index.html` path is misleading for the REST API 2.2 reference in Redoc.
> I wonder if it's possible to move it to `guides/v2.2` where the rest of 2.2 documentation is stored.

I considered that too. Technically, "swagger" refers to the schema specification (now known as OpenAPI) and "swagger-ui" refers to the lib that generates HTML from the "swagger/OpenAPI" schema, but I take your point.

I think we can easily move the 2.2 index file to the `src/redoc/` directory and add a `redirect_from:` the `swagger/` directory.

keharper:
If we're going through the trouble of re-skinning the REST docs, then we need to process the schema so that it's optimized for ReDoc. 

The asynchronous REST calls were a big deal in 2.3, but they were never going to be adopted as widely as the standard synchronous calls. I think they should remain, but they don't have to be carried forward.

jeff-matthews:
Thanks @keharper. How do we optimize the schema for Redoc?

keharper (approved):

jeff-matthews:
Confirmed that internal staging build `#87` redirects as expected. This PR is ready to merge.

jeff-matthews:
running tests

PR #8295, opened 2020-12-01
GraphQL: Add list of supported cards to Payflow Pro

keharper:

This pull request (PR) adds the list of credit cards that Magento supports to the Payflow Pro documentation. This is the only Magento-developed GraphQL payment method that handles card details directly.

This work is per MC-39519 and internal developer request.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.4/graphql/payment-methods/payflow-pro.html

-  https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Payment/view/base/web/js/model/credit-card-validation/credit-card-number-validator/credit-card-type.js#L13

### DIFF
--- a/src/_includes/graphql/payment-methods/payflow-pro-attributes.md
+++ b/src/_includes/graphql/payment-methods/payflow-pro-attributes.md
@@ -8,7 +8,7 @@ Attribute |  Data Type | Description
 
 ### CreditCardDetailsInput object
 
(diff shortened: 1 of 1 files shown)


erikmarr (approved):
I would move the `cc_type` values as well as the intro sentence to them ("Magento supports the following values for `cc_type`. The merchant's payment processor might support only a subset of these values.") below the table of attributes. I was confused about what was required until I realized the values in the attribute table were required not the `cc_type` values.

erikmarr (approved):

prabhuram93 (approved):

keharper:
running tests

PR #8296, opened 2020-12-01
Fix wrong cli command

dani97:
As per steps, the testsuite is defined in phpunit.xml and not phpunit.xml.dist. To run the tests we have to use phpunit.xml file

This pull request (PR) corrects the wrong example cli command to run static tests
The testsuite as per the previous steps are defined in phpunit.xml and not in phpunit.xml.dist
So we have to supply phpunit.xml as parameter

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/test/static/static_test_execution.html


BarnyShergold (approved):
Nice little fix, thanks for your contribution.

keharper (approved):

PR #8353, opened 2020-12-08
Customer query improvements

RakeshJesadiya:

This pull request (PR) improvement the Customer Query Page with Mapping of Objects link assignment. Improve the User experience while referring the page.

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.3/graphql/queries/customer.html
https://devdocs.magento.com/guides/v2.4/graphql/queries/customer.html


keharper (approved):
I'm accepting this PR only because of the additions you made to the src/_includes/graphql/customer-output-24.md

Do not update any links in which you're just moving the array braces to be inside the link. They might be ugly, but they're not incorrect either.

PR #8373, opened 2020-12-10
Use a layout processor rather than plugin in the "Customize the view of a checkout step" docs

Serializator:

Using a separate layout processor is preferred over an after or around plugin bound to the `Magento\Checkout\Block\Checkout\LayoutProcessor` class.

I rewrote the documentation in regards to "Customize the view of a checkout step" so that it uses the layout processor approach rather than a plugin.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.4/howdoi/checkout/checkout_customize.html


BarnyShergold (approved):
Brilliant! Nice work

dobooth (approved):

dobooth:
running tests

PR #8405, opened 2020-12-17
Update the linting workflow on GitHub

dshevtsov:

This pull request (PR:
1. Removes the Markdown lint test as a duplicate of the Lint Code Base test.
1. Configures linters to run on any pull request, not only when _master_ or _develop_ are base branches.


.github/workflows/linter.yml
```diff
@@ -11,14 +11,11 @@ name: Lint Code Base
...
-#############################
+###################################
+# Start the job on a pull request #
+###################################
 on:
-  pull_request:
-    branches:
-      - master
-      - 2.*-develop
+  pull_request
```
jeff-matthews, line 18: So we run the check on all branches, but we only require a passing check on protected branches, right? We configure that with the repo settings instead of the GitHub Action workflow config.

dshevtsov: Correct.

jeff-matthews (approved):

dshevtsov:
running tests

Issue #6317, opened 2020-01-02
Setup multiple domains info not correct in Set up multiple websites or stores with nginx topic


hungpvdev:
# General issue
Setup multiple domains using multiple virtual hosts is not correct
## Description:
-- In multiple virtual hosts, using Nginx "map" directive will not work correctly due to "map" directive only run once for the same variable. 

> Since variables are evaluated only when they are used, the mere declaration even of a large number of “map” variables does not add any extra costs to request processing. 

http://nginx.org/en/docs/http/ngx_http_map_module.html#map

## Possible solutions:
Instead of using map directive, we will set variable directly, map directive does not necessary due to we create Nginx file for each domain. 
set $MAGE_RUN_TYPE store;
set $MAGE_RUN_CODE fr;

## Additional information:
http://nginx.org/en/docs/http/ngx_http_map_module.html#map

dineshvb:
@magento I am working on this

ajithkumar-maragathavel:
@magento I am working on this

PR #8433, opened 2020-12-22
Issue-6317- Added a disclaimer about the map directive

dineshvb:

This pull request (PR) covers #6317 

## Affected DevDocs pages

-  /guides/v2.3/config-guide/multi-site/ms_nginx.html
-  /guides/v2.4/config-guide/multi-site/ms_nginx.html

### DIFF
--- a/src/guides/v2.3/config-guide/multi-site/ms_nginx.md
+++ b/src/guides/v2.3/config-guide/multi-site/ms_nginx.md
@@ -127,7 +127,7 @@ To create multiple virtual hosts:
    ```
 
    {:.bs-callout-info}
-   The `map` directive might not work correctly since it runs only once for the same variable. And since the variables are evaluated only when they are used, the mere declaration even of a large number of `map` variables does not add any extra costs to request processing. So, instead of using `map` directive, we will set variable directly like, ```set $MAGE_RUN_TYPE store; set $MAGE_RUN_CODE fr;```
+   The `map` directive might not work correctly, because it runs only once for the same variable. The variables are evaluated only when they are used, and as a result, the mere declaration even of a large number of `map` variables does not add any extra costs to request processing. Instead of using a `map` directive, the following example sets a variable directly, like `set $MAGE_RUN_TYPE store; set $MAGE_RUN_CODE fr;`
 
 1. Create another file named `german.mysite.mg` in the same directory with the following contents:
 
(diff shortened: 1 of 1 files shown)


src/guides/v2.3/config-guide/multi-site/ms_nginx.md
```diff
@@ -126,6 +126,9 @@ To create multiple virtual hosts:
    }
    ```
 
+   {:.bs-callout-info}
+   The `map` directive might not work correctly since it runs only once for the same variable. And since the variables are evaluated only when they are used, the mere declaration even of a large number of `map` variables does not add any extra costs to request processing. So, instead of using `map` directive, we will set variable directly like, ```set $MAGE_RUN_TYPE store; set $MAGE_RUN_CODE fr;```
```
keharper, line 130:
```suggestion
   The `map` directive might not work correctly, because it runs only once for the same variable. The variables are evaluated only when they are used, and as a result, the mere declaration even of a large number of `map` variables does not add any extra costs to request processing. Instead of using a `map` directive, the following example sets a variable directly, like ```set $MAGE_RUN_TYPE store; set $MAGE_RUN_CODE fr;```
```

Those triple backticks are not likely to render as desired. You probably only need one set.

keharper (changes requested):
@dineshvb  This PR will require technical review. Adobe/Magento will be shut down from Dec 24 to Jan 4, so it'll be a while until the PR can be processed further.

src/guides/v2.3/config-guide/multi-site/ms_nginx.md
```diff
@@ -126,6 +126,9 @@ To create multiple virtual hosts:
    }
    ```
 
+   {:.bs-callout-info}
+   The `map` directive might not work correctly, because it runs only once for the same variable. The variables are evaluated only when they are used, and as a result, the mere declaration even of a large number of `map` variables does not add any extra costs to request processing. Instead of using a `map` directive, the following example sets a variable directly, like `set $MAGE_RUN_TYPE store; set $MAGE_RUN_CODE fr;`
```
meker12, line 130: Not understanding the relationship between the directive running only once for the same variable, and declaring large number of `map` variables are evaluated only when used ...and does not add any extra processing costs.
```suggestion
   The `map` directive might not work correctly because it runs only once for the same variable. The variables are evaluated only when they are used, and as a result, even the declaration of a large number of `map` variables does not add any extra costs to request processing. Instead of using a `map` directive, the following example sets a variable directly, like `set $MAGE_RUN_TYPE store; set $MAGE_RUN_CODE fr;`
```

meker12 (changes requested):

dobooth:
This is copied directly from http://nginx.org/en/docs/http/ngx_http_map_module.html. We will not tolerate plagiarized content in our documentation. If the point still needs to be made, you must write it in your own words.

Issue #8493, opened 2021-01-09
Feedback on page: /guides/v2.4/extension-dev-guide/view-models.html


audio-engineer:
# General issue

## Description:

Line 3 in the "How to write view models" section is wrong. If `$viewModel = $block->getViewModel();` is used in the template to reference to the view model class, it won't work when using `<argument name="viewModel">`.

On the other hand, if `$viewModel = $block->getData("viewModel");` is used, `<argument name="viewModel" xsi:type="object">OrangeCompany\Catalog\ViewModel\MyNewViewModel</argument>` is valid.

## Possible solutions:

Line 3 in the example should be replaced with `<argument name="view_model" xsi:type="object">OrangeCompany\Catalog\ViewModel\MyNewViewModel</argument>`, OR it should be mentioned that `$viewModel = $block->getViewModel();` only works in conjunction with `<argument name="view_model" xsi:type="object">OrangeCompany\Catalog\ViewModel\MyNewViewModel</argument>` and that `$viewModel = $block->getData("<argument name>");` should be used if another argument name is used.

## Additional information:

Tested on Magento 2.4.1

pawan:
> # General issue
> ## Description:
> Line 3 in the "How to write view models" section is wrong. If `$viewModel = $block->getViewModel();` is used in the template to reference to the view model class, it won't work when using `<argument name="viewModel">`.
> 
> On the other hand, if `$viewModel = $block->getData("viewModel");` is used, `<argument name="viewModel" xsi:type="object">OrangeCompany\Catalog\ViewModel\MyNewViewModel</argument>` is valid.
> 
> ## Possible solutions:
> Line 3 in the example should be replaced with `<argument name="view_model" xsi:type="object">OrangeCompany\Catalog\ViewModel\MyNewViewModel</argument>`, OR it should be mentioned that `$viewModel = $block->getViewModel();` only works in conjunction with `<argument name="view_model" xsi:type="object">OrangeCompany\Catalog\ViewModel\MyNewViewModel</argument>` and that `$viewModel = $block->getData("<argument name>");` should be used if another argument name is used.
> 
> ## Additional information:
> Tested on Magento 2.4.1

Issue confirmed!

pawan:
@magento I am working on this

BarnyShergold:
@dobooth - Now this PR : #8459 is merged, can we now close this issue as being completed?

audio-engineer:
[...]

PR #8494, opened 2021-01-10
Fix set and get view model method code

pawan:

This pull request (PR) is fix for issue #8493

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.4/extension-dev-guide/view-models.html

### DIFF
--- a/src/guides/v2.3/extension-dev-guide/view-models.md
+++ b/src/guides/v2.3/extension-dev-guide/view-models.md
@@ -60,7 +60,7 @@ You can access the public methods for the view model class in the template:
 
 /** @var $viewModel \OrangeCompany\Catalog\ViewModel\MyNewViewModel */
 
-$viewModel = $block->getViewModel();
+$viewModel = $block->getData('viewModel');
 
 ?>
 <h1><?= $block->escapeHtml($viewModel->getTitle()); ?></h1>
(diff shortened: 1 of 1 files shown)


pawan:
Adobe CLA sign

hostep:
@pawan, thanks, but this duplicates https://github.com/magento/devdocs/pull/8459 I believe, maybe you can collaborate on that ticket to get everything explained better?

src/guides/v2.3/extension-dev-guide/view-models.md
```diff
@@ -139,3 +139,5 @@ $postArray = $viewModel->getPostData(
     ['product' => $_item->getEntityId()]
 );
 ```
+{:.bs-callout-info}
+If you use `argument name="view_model"` you can access it using `$viewModel = $block->getViewModel();` Or You can use `argument name="viewModel"` and can access `$viewModel = $block->getData('viewModel');`
```
BarnyShergold, line 143: IF this is true then the to changes above are mutually exclusive - only one of them should have been changed.

BarnyShergold (changes requested):

src/guides/v2.3/extension-dev-guide/view-models.md
pawan: yes, It depends on implementation! either we can use Getter/Setter OR we can get Data of argument!

pawan: @BarnyShergold  First change I made as per example and second change is from actual changes from file [https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Catalog/view/frontend/layout/catalog_product_view.xml#L178](url)

dobooth:
Thanks for this, @pawan. Since we have 3 PRs around this fix, feels proper to take the first one. Thanks for contributing, though. Nice to see so many eyes on the docs. We will take #8459 for the fix.

pawan:
OK, why it is showing 2 closed pull request at [https://opensource.magento.com/profile/pawan/contribution_statistic](url)

PR #8505, opened 2021-01-12
Fixed view model name on Layout XML file

DeoJr:

This pull request (PR) correction the model view name, the name must be separated by an underscore

## Affected DevDocs pages

- https://magento.stackexchange.com/questions/227479/block-getviewmodel-returning-null/318026#318026

-  ...

https://magento.stackexchange.com/questions/227479/block-getviewmodel-returning-null/318026#318026


hostep:
Thanks @DeoJr, but it looks like there are already at least 2 competing PR's around this topic:
- https://github.com/magento/devdocs/pull/8459
- https://github.com/magento/devdocs/pull/8494

aleron75 (approved):

dobooth:
Thanks for pointing that out, @hostep. I think it's proper to take the first PR that fixed it, so we will go with #8459 and close these other two. But I still would like a technical review check on #8459 from someone. And thanks to @pawan  and @DeoJr for submitting PRs around this.

PR #8709, opened 2021-02-19
Explain mixins

### LINKED ISSUE #2578: Add info how to declare custom less mixins
There is info in M2 docs how to declare custom variables or how to override existing less variables.
For example: custom less variables can be define in theme in `web/css/source/_variables.less`
But where custom less mixing must be declared?

Santoshziffity:

This pull request (PR) to explaining mixins

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/coding-standards/code-standard-less.html#mixins

Fixes #2578


src/guides/v2.3/coding-standards/code-standard-less.md
```diff
@@ -740,6 +740,86 @@ For mixins grouping use the double underscore "__" prefix.
...
+   -moz-transition: 300ms ease-in-out;
+   -webkit-transition: 300ms ease-in-out;
+   -o-transition: 300ms ease-in-out;
+   width: 100%;
+}
+```
+
+### Mixins with parameters
+
+There are also mixins with parameters. Calling these mixins, we pass them parameter values. Creating such mixin, we recommend setting parameter values by default (since there might occur problems when calling the mixin without specifying the parameter value). Let’s make a mixin with parameters from the example above and see how to call it:
```
BarnyShergold, line 787:
```suggestion
There are also mixins with parameters. When calling these mixins, we pass them parameter values. When creating this type of mixin, we recommend setting parameter values by default (since problems might occur when calling the mixin without specifying the parameter value). The example below shows how to create a mixin with parameters based on the example above and shows how to call it:
```

src/guides/v2.3/coding-standards/code-standard-less.md
```diff
@@ -740,6 +740,86 @@ For mixins grouping use the double underscore "__" prefix.
...
+    -o-transition: @animation-speed @animation-type;
+}
+.example-1 {
+    .animation-1(
+        @animation-speed: 1500ms
+    );
+}
+```
+
+Mixin parameters are set in parentheses (there can be one or several parameters) with default values (default values are set after a colon).
```
BarnyShergold, line 806:
```suggestion
Mixin parameters are set in parentheses (there can be one or more parameters) with default values (default values are set after a colon).
```

src/guides/v2.3/coding-standards/code-standard-less.md
```diff
@@ -740,6 +740,86 @@ For mixins grouping use the double underscore "__" prefix.
...
+
+Calling a mixin in parentheses, we indicate the value of those mixin parameters that will differ from the default values. If we didn’t specify any parameters adding a mixin as – .animation-1, the mixin would be added with standard parameter values.
```
BarnyShergold, line 808:
```suggestion
Calling a mixin in parentheses, it is necessary to indicate the value of those mixin parameters that will differ from the default values. If parameters are not specified when adding a mixin such as `.animation-1`, then the mixin would be added with standard parameter values.
```

src/guides/v2.3/coding-standards/code-standard-less.md
```diff
@@ -740,6 +740,86 @@ For mixins grouping use the double underscore "__" prefix.
...
+
+As a result, after compiling the .less file into a .css file, the .example-1 element will have the following:
```
BarnyShergold, line 810:
```suggestion
As a result, after compiling the `.less` file into a `.css` file, the `.example-1` element will have the following:
```

src/guides/v2.3/coding-standards/code-standard-less.md
```diff
@@ -740,6 +740,86 @@ For mixins grouping use the double underscore "__" prefix.
...
+```css
+.example-1 {
+   transition: 1500ms ease-in-out;
+   -moz-transition: 1500ms ease-in-out;
+   -webkit-transition: 1500ms ease-in-out;
+   -o-transition: 1500ms ease-in-out;
+}
+```
+
+Magento has an agreement to name mixin parameters with @_ instead of @. So, the @animation-speed parameter from the example above would be named as @animation-speed.
```
BarnyShergold, line 821: If this is so, then why do the above example NOT have parameters named @_ ? Also should not the name be @_animation-speed?

BarnyShergold (changes requested):
Good addition but there corrections and a question at the end to address.

dobooth:
Is this covering #2578?

Santoshziffity:
> Is this covering #2578?

Yes

dobooth (approved):

dobooth:
running tests

PR #8720, opened 2021-02-22
Fix Markdown

dshevtsov:

This pull request (PR) fixes [MD007 - Unordered list indentation](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md007---unordered-list-indentation). It has not been detected by the GitHub check linting because the linter it uses cannot process an unordered sublist if its parent list is ordered. A note from the official documentation:

> Note: This rule applies to a sublist only if its parent lists are all also unordered (otherwise, extra indentation of ordered lists interferes with the rule).

The violation was detected locally by the `rake test` implementation of the Markdown check.


jeff-matthews (approved):

jeff-matthews:
Thanks @dshevtsov

dobooth:
Thanks. I will run this one.

PR #8741, opened 2021-02-25
TOC changes to match one more logical update to PB org changes

bdenham:

This pull request (PR) pulls the **styles** and **viewports** directories out of the content types directory and into their own directories on the same level as content types. This makes much more sense because they are technically independent of content types (even though they are often used in content types).

## Affected DevDocs pages

- styles and viewport topics in Page Builder docs


jeff-matthews (approved):

bdenham:
running tests

PR #8844, opened 2021-03-30
add missing email required field and correct billingAddress field name

gixid192:
Fix the error according to this question: https://magento.stackexchange.com/questions/334702

According to this redoc link: https://magento.redoc.ly/2.4.2-guest/tag/guest-cartscartIdpayment-information
The email is required and also, `billingAddress` is mistyped

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.4/rest/tutorials/orders/order-create-order.html

### DIFF
--- a/src/guides/v2.3/rest/tutorials/orders/order-create-order.md
+++ b/src/guides/v2.3/rest/tutorials/orders/order-create-order.md
@@ -25,6 +25,8 @@ When you submit payment information, Magento creates an order and sends an order
 {:.bs-callout-info}
 Use the `V1/guest-carts/<cartId>/payment-information` endpoint to set the payment information on behalf of a guest. Do not include an authorization token.
 
+This guest-carts endpoint also requires `email` in the payload (same level as `paymentMethod` and `billing_address`)
+
 **Endpoint:**
 
 `POST <host>/rest/<store_code>/V1/carts/mine/payment-information`
@@ -44,7 +46,7 @@ Use the `V1/guest-carts/<cartId>/payment-information` endpoint to set the paymen
   "paymentMethod": {
     "method": "banktransfer"
   },
-  "billingAddress": {
+  "billing_address": {
     "email": "jdoe@example.com",
     "region": "New York",
     "region_id": 43,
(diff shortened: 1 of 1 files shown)


BarnyShergold (approved):
Nice!

keharper:
Hi @gixid192. Thanks for finding this issue on StackExchange and creating a PR to update the docs so quickly. Unfortunately, the fix isn't quite correct.

The tutorial is done from the perspective of a logged-in customer. The example uses the `POST /V1/carts/mine/set-payment-information` endpoint. Here's what ReDoc says about that endpoint.

customer-payment-info

The person on StackExchange was trying to use `POST V1/guest-carts/'.$quote_id.'/payment-information`. This is its syntax.

guest-payment-info

So, you see that the `email` attribute does not exist for `POST /V1/carts/mine/set-payment-information`. Adding it to the attribute to the example breaks the code sample for the main use case in the tutorial.

What we need to do instead is add a sentence to line 26 stating that the guest version of the endpoint also requires an email address at the top level of the payload.

And finally, it's not necessary to change `billing_address` to `billingAddress`. Although ReDoc/Swagger lists objects in camel case, Magento accepts either form. Although it's not wrong to change the form to `billingAddress`, it creates inconsistencies with other steps in the tutorial. I don't think it's worth the effort to comb through the examples and change the casing while the examples still work,

keharper:
Also, please rebase this PR to `master`. The changes apply to released versions of Magento. Our automation will subsequently update the `2.4.3-develop` branch.

gixid192:
Thanks so much for your clarification @keharper. I will add a note as you suggested into the article. 

One question, could I add the email as a comment like the image?
image

keharper:
@gixid192 In general, JSON isn't supposed to contain comments. I tested adding a comment to the payload of an endpoint, and that resulted with an error. We can't add a comment to the code sample because it's meant to be copy-and-paste. I would have accepted that suggestion if it didn't cause an error.

So, revert these changes and change line 26 to something like

Use the `V1/guest-carts/<cartId>/payment-information` endpoint to set the payment information on behalf of a guest. Do not include an authorization token. The payload must include the `email` attribute at the same level as `paymentMethod` and `billing_address`.

gixid192:
Hi @keharper,
I'm sorry that I had to create a new repo for this, I made a mess and cannot revert back.

Please take a look in the new PR: https://github.com/magento/devdocs/pull/8850

keharper:
No worries, @gixid192. I've had that problem myself. Thanks for your contribution.

PR #8926, opened 2021-05-07
Add upgrade CLI command

Santoshziffity:

This pull request (PR)  for Add upgrade CLI command 

## Affected DevDocs pages
-  https://devdocs.magento.com/guides/v2.3/config-guide/deployment/pipeline/example/shared-configuration.html
-  https://devdocs.magento.com/guides/v2.3/config-guide/deployment/pipeline/example/environment-variables.html
-  https://devdocs.magento.com/guides/v2.3/config-guide/deployment/pipeline/example/cli.html

### DIFF
--- a/src/_includes/config/split-deploy/example_build-sync.md
+++ b/src/_includes/config/split-deploy/example_build-sync.md
@@ -13,7 +13,7 @@ To update your build system:
 1. Upgrade Magento:
 
    ```bash
-   php bin/magento setup:upgrade
+   bin/magento setup:upgrade
    ```
 
 1. Compile code:
--- a/src/_includes/config/split-deploy/example_build-sync.md
+++ b/src/_includes/config/split-deploy/example_build-sync.md
@@ -10,6 +10,12 @@ To update your build system:
    git pull mconfig m2.2_deploy
    ```
 
+1. Run the upgrade command:
+
+   ```bash
+   php bin/magento setup:upgrade
+   ```
+
 1. Compile code:
 
    ```bash
--- a/src/_includes/config/split-deploy/example_build-sync.md
+++ b/src/_includes/config/split-deploy/example_build-sync.md
@@ -10,7 +10,7 @@ To update your build system:
    git pull mconfig m2.2_deploy
    ```
 
-1. Run the upgrade command:
+1. Upgrade Magento:
 
    ```bash
    php bin/magento setup:upgrade


src/_includes/config/split-deploy/example_build-sync.md
```diff
@@ -10,6 +10,12 @@ To update your build system:
    git pull mconfig m2.2_deploy
    ```
 
+1. Run the upgrade command:
+
+   ```bash
+   php bin/magento setup:upgrade
```
dobooth, line 16: We don't do `php` at the start of commands anymore, correct?

dobooth (changes requested):

src/_includes/config/split-deploy/example_build-sync.md
dobooth: Ok....so you can remove the `php`.

dobooth (changes requested):

hguthrie (approved):

dobooth:
Hey @Santoshziffity Just want to make sure: This is a big command to run...and it's in an include file. Are we sure this is appropriate for all instances where the include is used? Does any backup need to happen before running it?

ajithkumar-maragathavel:
> Hey @Santoshziffity Just want to make sure: This is a big command to run...and it's in an include file. Are we sure this is appropriate for all instances where the include is used? Does any backup need to happen before running it?

@dobooth  Santoshziffity will not be available for quite some time. we might need to check this with someone else.

src/_includes/config/split-deploy/example_build-sync.md
```diff
@@ -10,6 +10,12 @@ To update your build system:
    git pull mconfig m2.2_deploy
    ```
 
+1. Run the upgrade command:
```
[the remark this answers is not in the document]
hguthrie: If possible, please clearly state what is going on here. I used Upgrade Magento as a suggestion, but that's maybe not the most apt description. I have seen `Update the database schema and data.` used before. Please confirm what this does in this particular step.

BarnyShergold (changes requested):
Running this command at this stage requires a lot more information and explanation to be added as firstly it REQUIRES not only Magento to be installed on the system that the command is run on but also access to a Magento database. 

Even though the intention is to allow code files to be committed to the repository.

It also depends on the CI/CD workflow used.

EG if this were being run on a local machine, then Magento would need to be fully working - what happens if the database is not available?

I feel more work needs to be done to cover all eventualities.

dobooth:
I think I would more comfortable closing this PR for now. Once @Santoshziffity is back in action, we can discuss the optimal solution for this issue. Thanks all for the conversation. This is what makes open source a good idea.

PR #8964, opened 2021-05-25
Update integration_test_execution.md

danieluniform:
User should be notified of startup errors if the amqp- settings are not disabled.

This pull request (PR) ...

## Affected DevDocs pages

-  ...

-  ...


BarnyShergold (changes requested):
This is a great correction however you need to also do this on the 2.4 version of the file.

As you have another PR on these files, can you please -
1. Apply the changes on #8963 & #8964 to the 2.3 & 2.4 versions (so #8963 has 2 files on the PR and 2 changes to each file)
2. Close this PR.

Going forward, please always check if a change should be added to both 2.3 & 2.4 and place on the same PR - many 2.4 files are actually symlinks of the 2.3 version so only one change is needed.

PR #9091, opened 2021-08-11
Remove setup_version

dfelton:

This pull request helps with clarity by reducing unnecessary code.

`setup_version` is used for very specific types of `Setup/*.php` scripts. Not all modules make use of such scripts. Including reference to it here in this guide serves only as a distraction as there is no reference to it explaining what it does, nor should there as that is not the topic of this guide.

Some people including myself might even argue use of `setup_version` should be discouraged in favor of other practices available to us when performing Magento development.

Similar PR: https://github.com/magento/devdocs/pull/8980

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.4/javascript-dev-guide/javascript/js_mixins.html
- https://devdocs.magento.com/guides/v2.3/javascript-dev-guide/javascript/js_mixins.html

Not Applicable

### DIFF
--- a/src/guides/v2.3/javascript-dev-guide/javascript/js_mixins.md
+++ b/src/guides/v2.3/javascript-dev-guide/javascript/js_mixins.md
@@ -210,7 +210,7 @@ Be sure to add the origin module as the over-written module dependency (use the 
 ```xml
 <?xml version="1.0"?>
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Module/etc/module.xsd">
-    <module name="ExampleCorp_CartFix" setup_version="0.0.1">
+    <module name="ExampleCorp_CartFix">
         <sequence>
             <module name="ExampleCorp_Sample" />
         </sequence>


BarnyShergold (changes requested):
Now that I've seen a second PR on the same fix (just different files) it would seem prudent that if setup_version needs to be removed from DEVDOCS that rather than fixing one file at a time when someone spots it to extend this PR to ALL other docs.

I have checked and there are 32 instances across 24 files (.md), so it should not take too long to remove them.

Please can each of these be reviewed and removed as required? Then we will have properly fixed this issue.

dfelton:
@BarnyShergold sure thing, I think that is a great suggestion. I'll clone the devdocs repository locally so that I can search for these occurrences, thus far this and my last PR which you are alluding to have just been done in the web browser with GitHub's direct editing tools.

Will push an update as soon as I am able to.

BarnyShergold:
@dfelton - Do you have an update for us on this PR? Been 2 weeks since we heard from you 😄

BarnyShergold:
@dfelton - If we've not heard from you by the end of 3rd Sept then this PR will be closed (you can always reopen at a later date)

BarnyShergold:
Closing due to inactivity and no response from creator.

PR #9108, opened 2021-08-25
Documented an alternative Integration Tests data fixture declaration …

bgorski:
…using the @magentoDataFixture annotation

## Important:
The current (pre-PR) version of the modified file is just a symlink to the same file from the 2.3 docs. You'd have to use it if you want to display the diff yourself.
This is a screenshot of what's changed:
https://tinyurl.com/yhycp2fq

This pull request documents an alternative fixture declaration which is available in Magento for some time already, but not documented in the devdocs at all.
The documented way's example:
```php
/**
 * @magentoDataFixture Magento/Cms/_files/pages.php
 */
```
The undocumented way's example:
```php
/**
 * @magentoDataFixture VendorName_ModuleName::Test/Integration/_files/fixture_name.php
 */
```

## Affected DevDocs pages

- https://devdocs.magento.com/guides/v2.4/test/integration/annotations/magento-data-fixture.html

Although this kind of data fixture declaration is not documented in the devdocs, it's documented in the Integration Test Framework in the method that's responsible for resolving the path to the fixture.
[`dev/tests/integration/framework/Magento/TestFramework/Workaround/Override/Fixture/Resolver.php`](https://github.com/magento/magento2/blob/2.4-develop/dev/tests/integration/framework/Magento/TestFramework/Workaround/Override/Fixture/Resolver.php#L251) contains the following method:
```php
    /**
     * Check is the Annotation like Magento_InventoryApi::Test/_files/products.php
     *
     * @param string $fixture
     * @return bool
     */
    private function isModuleAnnotation(string $fixture): bool
    {
        return (strpos($fixture, '::') !== false);
    }
```
The method is called in the same file by the `processFixturePath` method:
```php
    /**
     * Converts fixture path.
     *
     * @param TestCase $test
     * @param string $fixture
     * @return string|array
     * @throws LocalizedException
     */
    private function processFixturePath(TestCase $test, string $fixture)
    {
        if (strpos($fixture, '\\') !== false) {
            // usage of a single directory separator symbol streamlines search across the source code
            throw new LocalizedException(__('Directory separator "\\" is prohibited in fixture declaration.'));
        }

        $fixtureMethod = [get_class($test), $fixture];
        if (is_callable($fixtureMethod)) {
            $result = $fixtureMethod;
        } elseif ($this->isModuleAnnotation($fixture)) {
            $result = $this->getModulePath($fixture);
        } else {
            $result = INTEGRATION_TESTS_DIR . '/testsuite/' . $fixture;
        }

        return $result;
    }
```
In case of module annotations, it goes into the `elseif` there and uses `$this->getModulePath($fixture);` to resolve the path. That method, in turn, uses the ComponentRegistrar to determine the module path (usually either `app/code/...` or `vendor/...`). It uses ComponentRegistrar::MODULE there, which means it won't work for libraries, lang packs, themes or setup, only for modules (which makes sense).

### DIFF
--- a/src/guides/v2.4/test/integration/annotations/magento-data-fixture.md
+++ b/src/guides/v2.4/test/integration/annotations/magento-data-fixture.md
@@ -42,29 +42,29 @@ As mentioned above, there are two ways to declare fixtures:
 ### Fixture as a separate file
 
 Define the fixture in a separate file when you want to reuse it in different test cases.
-To declare the fixture, use one of the following conventions 
+To declare the fixture, use one of the following conventions:
 
-- Fixture declaration as a path relative to the testsuite directory
-
-  - Relative to `dev/tests/integration/<test suite directory>`
-  - With forward slashes `/`
-  - No leading slash
+-  Fixture declaration as a path relative to the testsuite directory
+   -  Relative to `dev/tests/integration/<test suite directory>`
+   -  With forward slashes `/`
+   -  No leading slash
 
   Example:
-  ```
+
+  ```php
   /**
    * @magentoDataFixture Magento/Cms/_files/pages.php
    */
   ```
 
-- Fixture declaration as a path relative to a module
-
-  - Relative to the directory of a module available in the project
-  - With forward slashes `/`
-  - No leading slash in the path part of the declaration
+-  Fixture declaration as a path relative to a module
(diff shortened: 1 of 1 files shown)


BarnyShergold:
For some reason I cannot make suggestion reviews on this PR - can you please check access rights?

PR #9166, opened 2021-10-01
update install command and added pub in base url as per latest version

ravi-chandra3197:

This pull request (PR) ...

## Affected DevDocs pages

update install command and added pub in base url as per latest version
-  ...

-  ...

### DIFF
--- a/src/guides/v2.4/install-gde/install/cli/install-cli-install.md
+++ b/src/guides/v2.4/install-gde/install/cli/install-cli-install.md
@@ -59,7 +59,7 @@ If an error displays when you run these commands, verify that you updated instal
 The install command uses the following format:
 
 ```bash
-magento setup:install --<option>=<value> ... --<option>=<value>
+bin/magento setup:install --<option>=<value> ... --<option>=<value>
 ```
 
 The following tables describe the installation option names and values, such as installation commands. See [Sample localhost installations](#install-cli-example).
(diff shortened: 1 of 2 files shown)


BarnyShergold (changes requested):
Changes are great, but please fill out PR description as required in the WIKI.

ravi-chandra3197:
Hello @BarnyShergold 
I have Added PR description could you please review it?

hostep:
I don't think this PR is valid. The base url should never contain `pub`.
The document root defined in your webserver should point to the `pub` directory, but the base url in the database of Magento really shouldn't contain `pub`.

ravi-chandra3197:
> I don't think this PR is valid. The base url should never contain `pub`. The document root defined in your webserver should point to the `pub` directory, but the base url in the database of Magento really shouldn't contain `pub`.
> 
> _Update_: And someone should probably also double check if the initial instructions were correct, I don't think that `--base-url=http://127.0.0.1/magento2/` is correct and it probably should be `--base-url=http://127.0.0.1/` ?

Hello @hostep while we set up Magento without a virtual host we need to use a URL `http://127.0.0.1/magento2/pub/`  to point pub dir. 
I am using multiple magento2 setup in my local with URL pub url e.g  `http://127.0.0.1/magento2/pub/`, `http://127.0.0.1/magento242/pub/`, `http://127.0.0.1/magento243/pub/`, etc.

hostep:
Mkay, I have no experience with running Magento not inside a virtual host, so you might be correct.
It feels strange that Magento's doc would have recommendations to use it like that, because it's far from how shops are hosted in a production environment. And what you'll see locally (url-wise) will be so different from production. So I wonder why people even want to use this way of working?

dobooth:
@ravi-chandra3197 Why would you not instead change the Apache DocumentRoot to include `pub`?

PR #9189, opened 2021-10-13
Update 2-3-7-p2.md

denis-zyk:
Phrasing
> The Content Security Policy directive `frame-ancestors` now supports the source expression `unsafe-inline`

is totally incorrect, because in MDN Web Docs it literally says:
> Note: The `frame-ancestors` directive's syntax is similar to a source list of other directives (e.g. default-src), but **doesn't allow `'unsafe-eval'` or `'unsafe-inline'`** for example. It will also not fall back to a `default-src` setting.

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors

This pull request (PR) fixes misleading statement in release notes for 2.3.7-p2.

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.3/release-notes/2-3-7-p2.html

### DIFF
--- a/src/guides/v2.3/release-notes/2-3-7-p2.md
+++ b/src/guides/v2.3/release-notes/2-3-7-p2.md
@@ -37,7 +37,7 @@ This release includes fixes for the following known issues, which were first ide
 
 **Recent penetration test vulnerabilities** have been fixed in this release. <!--- MC-42431-->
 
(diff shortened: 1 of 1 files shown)


jfrontain:
@denis-zyk, thanks so much catching this. I've already fixed it in context of another PR. Closing.

Issue #9390, opened 2022-03-23
system-requirements : error in composer version in Magento >= 2.4.2 


tales-desde-mileto:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which topic?

Composer versions in system-requirements.html

### What's wrong with the content?

Incorrect composer version in version >= 2.4.2
https://devdocs.magento.com/guides/v2.4/install-gde/system-requirements.html
Table: src/_includes/install/system-requirements-table.md

2.4.5 | 2.4.4 | 2.4.3-p2 | 2.4.3     | 2.4.2 works with composer 2
But in last doc we can read: composer 1
Or, perhaps, both version.
Now, if we read this, and we want to install 2.4.3p2 and have composer 2, according to this manual, I would say that it is not compatible. But it is compatible.

Is compatible from 2.4.2:
In doc:
https://devdocs.magento.com/guides/v2.4/release-notes/commerce-2-4-2.html
we can read:
------------
Magento 2.4.2 is now compatible with Composer 2.x. We recommend that merchants migrate to Composer 2.x. Although you can install this release using Composer 1.x, Composer 1.x will soon reach end-of-life. For an overview of Composer 2.x features, see [Composer 2.0 is now available!](https://blog.packagist.com/composer-2-0-is-now-available/)
------------
Then, if 2.4.2 works with both version: composer 1 and 2. Not only 1.

In previous versions of this doc, they show this value: composer 2. This is the commit.
diff --git a/src/_includes/install/system-requirements-table.md b/src/_includes/install/system-requirements-table.md
index 423a41db83..9764877222 100644
--- a/src/_includes/install/system-requirements-table.md
+++ b/src/_includes/install/system-requirements-table.md

### What changes do you propose?

Put all versions, like php row, there are values like: "7.2,7.3"
Then, in composer row, if works with version 1 and 2, show both.

jeff-matthews:
Hi @tales-desde-mileto,

According to our engineering team, Adobe has not officially tested Composer 2 on versions earlier than 2.4.4. So although Composer 2 may be "compatible" with those versions, it does not necessarily mean that Adobe "supports" it. We try to make that clear in the introductory paragraph before the table:

> This table shows versions of third-party software dependencies that Adobe has tested with specific Adobe Commerce and Magento Open Source releases. Adobe only supports the combination of system requirements described in the following table.
>
[...]

PR #9425, opened 2022-04-25
Fixed claimed composer version compatibility with Magento 2.3.7-p3

hostep:

This pull request fixes the composer version that Adobe recommends for Magento 2.3.7-p3

Claiming support for composer version 1 on Magento 2.3.7-p3 makes no sense, because composer version 2 was supported with the release of 2.3.7 (and continued in 2.3.7-p1 and 2.3.7-p2).
And in my experience on about 10 different Magento projects running on 2.3.7-p3, it works perfectly fine.

In my opinion version 2 should be added as well to Magento 2.4.2 and higher, but there is still some ongoing [discussion around that](https://github.com/magento/devdocs/issues/9390).

## Affected DevDocs pages

-  https://devdocs.magento.com/guides/v2.4/install-gde/system-requirements.html

N/A

### DIFF
--- a/src/_data/codebase/v2_3/system-requirements.yml
+++ b/src/_data/codebase/v2_3/system-requirements.yml
@@ -12,7 +12,7 @@
   Apache: '2.4'
   nginx: '1.8'
 2.3.7:
-  Composer: '2'
+  Composer: '1'
   Elasticsearch: '7.9'
   MariaDB: '10.3'
   MySQL: '5.7'


jeff-matthews:
@hostep, as I mentioned in #9390 ([comment](https://github.com/magento/devdocs/issues/9390#issuecomment-1076406554)), the system requirements table on devdocs represents the combination of software that Adobe used to do regression testing for a release. Those versions are therefore "supported" and we only add supported versions to the table.

hostep:
Why did they downgrade composer from 2 to 1 between 2.3.7 and 2.3.7-p3 for regression testing? That makes not much sense?

Anyway, that table gives your users a very confusing experience. If I would be a newcomer in this world and would end up on that page without much knowledge about Magento, I would already start questioning the validity of the entire page and question if the page is being properly maintained and if I can really trust it or not.
Which is really bad if you want to welcome new users.

jeff-matthews:
According to engineering, all images that were used in automated testing for versions earlier than 2.4.4 included Composer 1, so that's clearly a mistake in the table that we can fix.

I think this conversation and others like it make it clear that we do our best to keep this table up to date.

jeff-matthews:
I have another PR that will fix this. Will close this after merging.

src/_includes/install/system-requirements-table.md
```diff
@@ -2,7 +2,7 @@
 {:.comparison-table}
 |                    | 2.4.5 | 2.4.4 | 2.4.3-p2 | 2.4.3     | 2.4.2 | 2.4.1 | 2.4.0            | 2.3.7-p3 | 2.3.7 | 2.3.6 | 2.3.5      | 2.3.4         | 2.3.3         | 2.3.2      | 2.3.1      | 2.3.0      |
 |--------------------|-------|-------|----------|-----------|-------|-------|------------------|----------|-------|-------|------------|---------------|---------------|------------|------------|------------|
-| Composer           | 2.x   | 2.x   | 1.x      | 1.x       | 1.x   | 1.x   | 1.x              | 1.x      | 2.x   | 1.x   | 1.x        | 1.x           | 1.x           | 1.x        | 1.x        | 1.x        |
+| Composer           | 2.x   | 2.x   | 1.x      | 1.x       | 1.x   | 1.x   | 1.x              | 2.x      | 2.x   | 1.x   | 1.x        | 1.x           | 1.x           | 1.x        | 1.x        | 1.x        |
```
jeff-matthews, line 5:
```suggestion
| Composer           | 2.x   | 2.x   | 1.x      | 1.x       | 1.x   | 1.x   | 1.x              | 1.x      | 1.x   | 1.x   | 1.x        | 1.x           | 1.x           | 1.x        | 1.x        | 1.x        |
```

jeff-matthews (changes requested):
I thought this was an issue instead of a PR. My mistake. If you accept the code suggestions, I can merge this.

hostep:
@jeff-matthews: okay changed according to your recommendations.

I do get that Adobe only wants to claim compatibility with this exact set of versions, but it's definitely not being used like this out in the real world (maybe only in a Magento Cloud environment, but even that I strongly doubt).

Would you be open to an alternative community-maintained versioning table which will help your users a lot better? By broadening the version support to what is actually proven to work in the real world. And makes a lot more sense than what you have currently.
For example something like this: https://github.com/navarr/magento2-supported-versions

jeff-matthews:
Thanks, @hostep. I'm not open to maintaining two system requirements tables on devdocs because I think that would create more confusion than clarity. Aside from the issue you raised about Composer 2, are there any other issues you have with the system requirements table?

hostep:
Yeah, here are some issues:

- the versions are too strict/narrow (but that's because the table does not represent "system requirements" but actually "here's the software that is guaranteed to work because these are the only versions we tested on". This missing information could for example lead to people to postpone Magento upgrades because they think they will have to upgrade a bunch of other software, and that's not always needed. Some examples:
    - Elasticsearch, I would specify 7.x instead of a specific version number, for example we have Magento 2.3.5 shops that run just fine on ES 7.17
    - Magento 2.3.7 works perfectly fine on both PHP 7.3 and 7.4 & Magento 2.4.1 and 2.4.2 run perfectly fine on PHP 7.3
    - Magento 2.4.x works perfectly fine on MySQL 5.7
    - Magento 2.3.7 works perfectly fine with Redis 6.2 or Redis 5.0 or anything in between
    - Magento > 2.3.5 work perfectly fine with Varnish 7.0
    - ...
- not all "requirements" are actually required, so maybe indicate that some of them are optional (rabbitmq, redis, varnish & elasticsearch/opensearch but that last pair only for Magento 2.3.x, for 2.4.x it is actually required)
- back to composer: composer v2 works perfectly fine on 2.3.7, 2.4.2 and 2.4.3 and brings with it an very high productivity gain for the developer, so it would be a shame that this table tells the developer that composer v1 is supposedly "unsupported" which might make them think that they can't use composer v2

hostep:
@jeff-matthews: reworked my PR after [your recent changes](https://github.com/magento/devdocs/commit/49bfe4e1e51c71048d5250bb35f71cdbc846e7f8). I like that you guys splitted up the table in a part for Magento 2.3 and one for 2.4, it makes it easier to digest now 👍

jeff-matthews:
Thanks for that @hostep!

jeff-matthews (approved):

jeff-matthews:

m2-github-services:
@jeff-matthews the branch with the code is successfully imported into the `magento-devdocs/devdocs` repository. Branch name: `imported-hostep-devdocs-9425`.

Issue #9447, opened 2022-05-27
Page Not Found for Change message queue from MySQL to AMQP link


engcom-Hotel:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which topic?

https://devdocs.magento.com/guides/v2.4/config-guide/mq/manage-message-queues.html

### What's wrong with the content?

The link mentioned in the text `Change message queue from MySQL to AMQP` is not working and redirecting to a 404 page.

### What changes do you propose?

Please correct the link or remove the text if it is not required.

erikmarr:
@engcom-Hotel thanks for your feedback! Looks like the link is working now.

PR #9450, opened 2022-05-31
Corrected broken AMQP link

BarnyShergold:

This pull request (PR) solves #9447 

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.4/config-guide/mq/manage-message-queues.html


hguthrie (changes requested):
Looks like something going on with `whats-new.md`. Can you investigate and fix that?

BarnyShergold:
> Looks like something going on with `whats-new.md`. Can you investigate and fix that?
I had to update some file permissions on my local machine and whilst the `whats-new.md` has not actually changed, it got added to the PR when I added to the commit. So it's an empty commit on that file. 

I wouldn't expect this would cause a problem, but I'm not sure how to reverse out if you require it.

Can you let me know?

hguthrie:
okay, I'll see what I can do.

hguthrie:

hguthrie:
Actually, I did some testing and this is not a case of removing version, I think. The topic was migrated to the Dev site, and I don't think the redirect is working. 
@jeff-matthews I will update the config guide with the following link, but adding you for visibility since it looks like migration problem:
https://developer.adobe.com/commerce/php/development/components/message-queues/#change-message-queue-from-mysql-to-amqp
We can close this PR, since it won't fix the problem, but thanks for alerting us on the broken link @BarnyShergold !

hguthrie:
@jeff-matthews Wait! yay! I did manage to see the redirect work, but I am still skeptical about this PR. There is something weird with `100644 → 100755` but don't worry. Changes will make it to config guide.

hguthrie:
Yes, but there is something else weird about this PR. It's pointing to something blank and I can't get it to behave. I have made the corrections outside of this PR.

jeff-matthews:
Agreed. I haven't seen that empty commit business before.

PR #9469, opened 2022-06-17
From ES to OS

rafaelstz:

Since ElasticSearch won't be supported anymore by the Cloud version and this doc is under the Cloud section, in this PR I'm switching the installation instructions to OpenSearch.

## Affected DevDocs pages

-  https://devdocs.magento.com/cloud/before/before-setup-env-install.html

Ref:
- https://devdocs.magento.com/cloud/project/services-elastic.html
- https://devdocs.magento.com/cloud/project/services-opensearch.html

### DIFF
--- a/src/cloud/before/before-setup-env-install.md
+++ b/src/cloud/before/before-setup-env-install.md
@@ -138,8 +138,8 @@ To install {{site.data.var.ee}} using the command line:
      --timezone=America/Chicago \
      --language=en_US \
      --use-rewrites=1 \
-     --search-engine=elasticsearch7 \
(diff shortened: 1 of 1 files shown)


hguthrie:
We don't have an OpenSearch engine.

rafaelstz:
Hi @hguthrie, but what about this reference:
Screen Shot 2022-08-04 at 1 18 25 PM

hguthrie:
We use OpenSearch, but we don't have an engine to reference in the configuration. We still use `elasticsearch7` in the `search-engine` notation. The highlight in your screenshot is for the type, which OpenSearch is valid.

PR #9494, opened 2022-08-08
Don't install dev dependencies by default

ssx:

This pull request corrects a glaring oversight of composer installing development packages.

## Affected DevDocs pages
https://devdocs.magento.com/cloud/project/magento-app-properties.html

### DIFF
--- a/src/cloud/project/magento-app-properties.md
+++ b/src/cloud/project/magento-app-properties.md
@@ -64,7 +64,7 @@ dependencies:
 hooks:
     build: |
         set -e
-        composer --no-ansi --no-interaction install --no-progress --prefer-dist --optimize-autoloader
+        composer --no-dev --no-ansi --no-interaction install --no-progress --prefer-dist --optimize-autoloader
 ```
 
 ## `access`


michaelcasey316 (approved):
Thanks for submitting this pull request. We're planning to review this internally.

hguthrie:
Thank you for taking the time to offer your suggestion. After some internal investigation, we must decline this proposed edit. The same `.magento.app.yaml` file persists in all environments in most cases; therefore, we have to maintain a very generic example.

You indicated this was a "glaring oversight"; are you experiencing any concerning issues as a result of using the provided command?

ssx:
> Thank you for taking the time to offer your suggestion. After some internal investigation, we must decline this proposed edit. The same `.magento.app.yaml` file persists in all environments in most cases; therefore, we have to maintain a very generic example.
> 
> You indicated this was a "glaring oversight"; are you experiencing any concerning issues as a result of using the provided command?

By default it'll install packages like YireoCheckout tester that can easily leak data. Development packages shouldn't be installed on production.

ssx:
> Thank you for taking the time to offer your suggestion. After some internal investigation, we must decline this proposed edit. The same `.magento.app.yaml` file persists in all environments in most cases; therefore, we have to maintain a very generic example.
> 
> You indicated this was a "glaring oversight"; are you experiencing any concerning issues as a result of using the provided command?

Sure you'd go for the most secure option by default and allow the users who need to change this to change it. If they're installing dev dependencies on a live site you'd hope they knew what they were doing whereas at the moment they probably aren't even aware.

hguthrie:
I appreciate this additional information. I will inquire further. Thank you!

Schrank:
That reminds me on https://thephp.cc/artikel/phpunit-ein-sicherheitsrisiko (unfortuantely german) but TL;DR: development tools are for development and in such PHPUnit 5 has a Remote Code Execution "vulnerability"... but it isn't - phpunit doesn't belong on production in the first place.

(I don't know wether phpunit is part of the packages, but dev tools don't belong on production env)

hguthrie:
I am sorry for the long wait. Unfortunately, I am not able to receive technical approval because this has the potential to interfere with the Cloud platform framework. This is a default example that works for all environments. Users have the freedom to customize their code and module usage as they see fit to accommodate their security requirements.

Issue #9498, opened 2022-08-12
System requirements Composer version is incorrect for v2.4.2


damienwebdev:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which topic?

https://github.com/magento/devdocs/blob/master/src/_data/codebase/v2_4/system-requirements.yml

### What's wrong with the content?

The v2.4.2 `composer` requirement is incorrect. The requirement is composer 2, not 1.

You could instead make it `1,2`

https://devdocs.magento.com/guides/v2.4/release-notes/open-source-2-4-2.html#platform-enhancements

### What changes do you propose?

Change it to 2.

PR #9499, opened 2022-08-12
Update system-requirements.yml

damienwebdev:

This pull request (PR) fixes the incorrectly listed composer version requirement.

## Affected DevDocs pages

https://devdocs.magento.com/guides/v2.4/install-gde/system-requirements.html

Fixes : #9498

### DIFF
--- a/src/_data/codebase/v2_4/system-requirements.yml
+++ b/src/_data/codebase/v2_4/system-requirements.yml
@@ -68,7 +68,7 @@
   AWS ElastiCache: Redis 6
   AWS Elasticsearch: '7.9'
 2.4.2:
-  Composer: '2'
+  Composer: '1,2'
   Elasticsearch: '7.9'
   MariaDB: '10.4'
   MySQL: '8.0'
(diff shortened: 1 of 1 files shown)


BarnyShergold (changes requested):
If this is to be corrected then the 2.4.3 & 2.4.3-p2 also need to be updated to the same

hostep:
And the same change is needed for Magento 2.3.7:
- https://devdocs.magento.com/guides/v2.3/release-notes/open-source-2-3-7.html#platform-upgrades

But we already had that discussion [many](https://github.com/magento/devdocs/issues/9390) [times](https://github.com/magento/devdocs/pull/9425) before. I wonder if Adobe will continue blocking this or not because it's kind of ridiculous that they have conflicting documentation about composer support in various versions of Magento.

dobooth:
Composer 2 is supported by 2.4.2-p1 and -p2. Currently we don't have sys req columns for those. Either way, the table is now on Experience League and the changes should be made there.

PR #9528, opened 2022-10-13
add changelog to Open Source Edition

jonashrem:

This pull request copies the changelog from commerce 2.4.5-p1 and 2.4.4-p2 to open source
## Affected DevDocs pages

new pages

/


jeff-matthews:
Thanks for the contribution @jonashrem. I'll have to double check, but I thought that Composer packages for security patches are available for Adobe Commerce only.

jonashrem:
my understanding is, that the updates are available for open source as well.

See  https://github.com/magento/magento2/tree/2.4.5-p1 for example

dobooth:
Thanks for this @jonashrem. We are going to decline this PR as we think the existing files are sufficient to cover the changes.

